PR #2, opened 2017-01-25
Default to only building on JDK 8

jglick:
We only really recommend building on JDK 8, so do not waste resources doing extra builds with JDK 7.

All plugins using a 2.x parent POM will anyway be specifying an explicit baseline Java Platform version—7 by default. So all that is skipped here is running tests on Java 7. Since the majority of users are on 8, and we will soon be requiring 8 for new versions of Jenkins, this will just mean that we will not be catching functional problems which only manifest themselves on an EOL platform. I think that is a reasonably price to pay for cheaper builds.

(And if you are on an obsolete parent POM, well what are you doing using a shiny new `Jenkinsfile`?)

@reviewbybees

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -20,7 +20,7 @@ buildPlugin
 
 ==== Optional arguments
 
-* `jdkVersions` (default: `[7, 8]`) - JDK version numbers, must match a version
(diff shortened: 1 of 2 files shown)


rtyler:
This is reasonable to me, FYI @batmat

PR #3, opened 2017-01-31
Allow Jenkins versions to be specified as a new axis

jglick:
Experimental, plan to test in context.


rtyler:
> plan to test in context.

Does that mean this hasn't been tested at all?

jglick:
Correct.

jglick:
BTW you [messed up](http://stackoverflow.com/a/18149299/12916) here. There is no way to specify a value for an option in the middle without redundantly specifying all the earlier ones. Probably deserves a `buildPlugin2` that takes a single `Map` argument.

jglick:
> this hasn't been tested at all

And how could it be? Without Replay permission the only way I can try it is to have a job start running it. Which is fine by me; it may well need to be tuned somehow.

jglick:
In part this is a workaround for [JENKINS-41631](https://issues.jenkins-ci.org/browse/JENKINS-41631), though there are other reasons to want to retest against new LTS lines.

jglick:
Works now. @reviewbybees

vars/buildPlugin2.groovy
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env groovy
+
+/**
+ * Simple wrapper step for building a plugin
+ */
+def call(Map params = [:]) {
+    def platforms = params.containsKey('platforms') ? params.platforms : ['linux', 'windows']
+    def jdkVersions = params.containsKey('jdkVersions') ? params.jdkVersions : [8]
+    def jenkinsVersions = params.containsKey('jenkinsVersions') ? params.jenkinsVersions : [null]
```
jglick: added

vars/buildPlugin2.groovy
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env groovy
+
+/**
+ * Simple wrapper step for building a plugin
+ */
+def call(Map params = [:]) {
```
jglick: switching call format so we can actually do parameter defaults

vars/buildPlugin2.groovy
```diff
@@ -0,0 +1,93 @@
...
+    Map tasks = [failFast: failFast]
+    for (int i = 0; i < platforms.size(); ++i) {
+        for (int j = 0; j < jdkVersions.size(); ++j) {
+            for (int k = 0; k < jenkinsVersions.size(); ++k) {
+                String label = platforms[i]
+                String jdk = jdkVersions[j]
+                String jenkinsVersion = jenkinsVersions[k]
+                String stageIdentifier = "${label}-${jdk}${jenkinsVersion ? '-' + jenkinsVersion : ''}"
+
+                tasks[stageIdentifier] = {
```
jglick: using same computed value

vars/buildPlugin2.groovy
```diff
@@ -0,0 +1,93 @@
...
+
+                        stage("Build (${stageIdentifier})") {
+                            List<String> mavenOptions = [
+                                '--batch-mode',
+                                '--errors',
+                                '--update-snapshots',
+                                '-Dmaven.test.failure.ignore=true',
+                                "-DskipAfterFailureCount=${failFast}",
+                            ]
+                            if (jenkinsVersion) {
```
jglick: added

vars/buildPlugin2.groovy
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Simple wrapper step for building a plugin
+ */
+def call(Map params = [:]) {
+    def platforms = params.containsKey('platforms') ? params.platforms : ['linux', 'windows']
+    def jdkVersions = params.containsKey('jdkVersions') ? params.jdkVersions : [8]
+    def jenkinsVersions = params.containsKey('jenkinsVersions') ? params.jenkinsVersions : [null]
+    def repo = params.containsKey('repo') ? params.repo : null
+    def failFast = params.containsKey('failFast') ? params.failFast : true
+    Map tasks = [failFast: failFast]
```
jglick: making `failFast` also be a `parallel` option

vars/buildPlugin2.groovy
```diff
@@ -0,0 +1,93 @@
...
+                                        bat mavenCommand
+                                    }
+                                }
+                            }
+                        }
+
+                        stage("Archive (${stageIdentifier})") {
+                            timestamps {
+                                junit '**/target/surefire-reports/**/*.xml'
+                                if (failFast && currentBuild.result == 'UNSTABLE') {
```
jglick: added

rtyler:
I don't actually think anybody is taking advantage of the poorly-implemented optional parameters of `buildPlugin`, I would rather just drop that in place instead of keeping `buildPlugin` in its current form around.

The use is [limited](https://ci.jenkins.io/job/Plugins/) already, and I'm willing to manage any fallout if we see some.

jglick:
> I don't actually think anybody is taking advantage of the poorly-implemented optional parameters of `buildPlugin`

`hubot-steps-plugin`, so [its build](https://ci.jenkins.io/job/Plugins/job/hubot-steps-plugin/job/master/) would be broken unless I also filed a PR to switch to the new syntax. Since there is no real cost to introducing variants (we will surely want to make some other incompatible change in the future), my preference was to just be compatible.

Let me know what you think. In any event it was impossible for me to test my changes under the original name, because you made the library `implicit` (and also, apparently, defined it twice? with names `pipeline-library` as well as `infra/pipeline-library` IIRC?) and so it prevented me from prototyping changes (lacking **Replay** permission) since the `master` version would get loaded instead.

vars/buildPlugin2.groovy
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env groovy
+
+/**
+ * Simple wrapper step for building a plugin
+ */
+def call(Map params = [:]) {
+    def platforms = params.containsKey('platforms') ? params.platforms : ['linux', 'windows']
```
orrc: Is there an advantage to this syntax over `def platforms = params.platforms ?: ['linux', 'windows']`?

(just wondering)

jglick:
> Is there an advantage to this syntax

Yes—it permits, say,

```groovy
buildPlugin(failFast: false)
```

which would not work correctly with `?:` syntax.

PR #4, opened 2017-02-10
add support for Gradle projects

daspilker:


daspilker:
a Gradle project can be tested here: https://ci.jenkins.io/job/Plugins/job/job-dsl-plugin/view/Pull%20Requests/job/PR-994/

vars/buildPlugin.groovy
```diff
@@ -34,47 +36,73 @@ def call(Map params = [:]) {
...
+                            } else {
+                                List<String> gradleOptions = [
+                                        '--no-daemon',
+                                        'cleanTest',
+                                        'build',
+                                ]
+                                command = "gradlew ${gradleOptions.join(' ')}"
+                                if (isUnix()) {
+                                    command = "./" + command
+                                }
```
orrc: What about Windows (`gradlew.bat`)?

daspilker: It's not necessary to add the extension on Windows.

orrc: Good point! But in any case, sorry — I misread.  This is just adding the `./` prefix for Unix, not actually *running* the command…

rtyler:
@daspilker thanks for taking the time to do this :fireworks:

PR #6, opened 2017-02-28
[WiP] - Change towards hosting modules on ci.jenkins.io

oleg-nenashev:
I work on the new release of the [Windows Agent Installer Module](https://github.com/jenkinsci/windows-slave-installer-module), which currently has no active CI. So I would like to have a build flow defined by this lib. Downstream PR with Jenkinsfile: https://github.com/jenkinsci/windows-slave-installer-module/pull/7

Changes:

- [x] - Add the new `buildModule()` command. Now it is just a wrapper on the top of `buildPlugin()`, but it may change in the future (e.g. integration tests with the latest Jenkins core, etc.).
- [x] - Add the `allowEmptyTestResults` option to both `buildPlugin()` and `buildModule()`

@reviewbybees, esp. @rtyler and @jglick


jglick:
> integration tests with the latest Jenkins core

That is already supported with `buildPlugin`.

> Add the `allowEmptyTestResults` option

This seems like a bad idea. You should always have `InjectedTest` at least.

README.adoc
```diff
@@ -29,6 +29,8 @@ buildPlugin()
   execute the steps against in parallel
 * `jenkinsVersions`: (default: `[null]`) - a matrix of Jenkins baseline versions to build/test against in parallel (null means default,
   only available for Maven projects)
+* `allowEmptyTestResults` (default: `false`) - Do not fail the build if the test results are missing.
+   It may be used for components, which have no tests within the repo
```
jglick: Delete comma. You are saying something grammatically valid but quite different than what I think you meant: “components have no tests, so you may use this flag for them” → “if you have a component which has no tests, then you may use this flag for it” (subtle point of English punctuation; I know at least Czech lacks this rule)

vars/buildModule.groovy
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env groovy
+
+def call(Map params = [:]) {
+  buildPlugin(params)
+}
```
jglick: add trailing newline

oleg-nenashev:
> That is already supported with buildPlugin.

Maybe. But having a separate command is definitely reasonable for the future use

> This seems like a bad idea. You should always have InjectedTest at least.

Not for the Jenkins module

jglick:
> Not for the Jenkins module

Well I disagree with that, but will leave it to the downstream PR.

rtyler:
> Maybe. But having a separate command is definitely reasonable for the future use

@oleg-nenashev would you mind adding some commentary into the `buildModule.groovy` itself explaining what potential future use-cases there might be? I don't yet understand why a redundant API is necessary.

oleg-nenashev:
@rtyler Let's put in on hold for a while. I really need to come-up with a realistic use-case for the Jenkins module lifecycle, but my previous ideas seem to be not really relevant since Jenkins modules are too close to plugins now.

rtyler:
I'm going to close this out until we can figure out whether there's a strong use-case here. Feel free to re-open it if something comes up.

PR #10, opened 2017-06-08
Limit use of settings-azure.xml to Java 8+ due to cert issues

daniel-beck:
Untested.


orrc:
Context: https://issues.jenkins-ci.org/browse/INFRA-1214

PR #11, opened 2017-06-22
Minor cleanups in buildPlugin

jglick:
Untested, since we have no good mechanism for testing this. Really I would rather see the `implicit` flag turned off and `Jenkinsfile`s look like

```groovy
@Library('pipeline-library@abc1234d') _
buildPlugin()
```

so that we are using a well-defined version of the library for every repository, and updates can be rolled out incrementally in pull requests without risk of breaking everyone’s build.

@reviewbybees

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -24,14 +24,10 @@ def call(Map params = [:]) {
 
                         stage("Checkout (${stageIdentifier})") {
                             if (env.BRANCH_NAME) {
-                                timestamps {
-                                    checkout scm
-                                }
+                                checkout scm
                             }
                             else if ((env.BRANCH_NAME == null) && (repo)) {
-                                timestamps {
-                                    git repo
-                                }
+                                git repo
                             }
                             else {
                                 error 'buildPlugin must be used as part of a Multibranch Pipeline *or* a `repo` argument must be provided'
@@ -55,7 +51,7 @@ def call(Map params = [:]) {
                                         '-Dmaven.test.failure.ignore',
                                 ]
                                 if (jdk.toInteger() > 7) {
-                                    /* Azure mirror only works for Java 8+ due to Letsencrypt cert */
+                                    /* Azure mirror only works for sufficiently new versions of the JDK due to Letsencrypt cert */
                                     def settingsXml = "${pwd tmp: true}/settings-azure.xml"
                                     writeFile file: settingsXml, text: libraryResource('settings-azure.xml')
                                     mavenOptions += "-s $settingsXml"
@@ -78,15 +74,11 @@ def call(Map params = [:]) {
                             }
 
                             withEnv(env) {
-                                if (isUnix()) {
-                                    timestamps {
-                                        sh command
-                                    }
+                                if (isUnix()) { // TODO JENKINS-44231 candidate for simplification
+                                    sh command
                                 }
                                 else {
-                                    timestamps {
-                                        bat command
-                                    }
+                                    bat command
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


vars/buildPlugin.groovy
```diff
@@ -24,14 +24,10 @@ def call(Map params = [:]) {
 
                         stage("Checkout (${stageIdentifier})") {
                             if (env.BRANCH_NAME) {
-                                timestamps {
-                                    checkout scm
-                                }
+                                checkout scm
```
jglick: Simpler to wrap everything in `timestamps`, as done at the end.

vars/buildPlugin.groovy
```diff
@@ -55,7 +51,7 @@ def call(Map params = [:]) {
                                         '-Dmaven.test.failure.ignore',
                                 ]
                                 if (jdk.toInteger() > 7) {
-                                    /* Azure mirror only works for Java 8+ due to Letsencrypt cert */
+                                    /* Azure mirror only works for sufficiently new versions of the JDK due to Letsencrypt cert */
```
jglick: In fact some JDK 8 builds failed, and maybe there are some private JDK 7 updates which work.

vars/buildPlugin.groovy
```diff
@@ -102,14 +94,11 @@ def call(Map params = [:]) {
...
                             }
 
-                            timestamps {
-                                junit testReports
-                                if (failFast && currentBuild.result == 'UNSTABLE') {
-                                    error 'There were test failures; halting early'
-                                }
-                                archiveArtifacts artifacts: artifacts,
-                                            fingerprint: true
+                            junit testReports // TODO do this in a finally-block so we capture all test results even if one branch aborts early
```
jglick: Not feeling confident enough to do that refactoring without ability to test the effect.

jglick:
@reviewbybees done

Not sure who maintains this—@rtyler?

rtyler:
> Untested, since we have no good mechanism for testing this. 

There's not CI for this repository true, but there's no reason a local test-bed cannot be used to verify this. @bitwiseman and I have spent lots of time with local test beds to verify Shared Library edits, and thus far it seems to be the only viable testing approach (manual :frowning:)

> Really I would rather see the implicit flag turned off and Jenkinsfiles look like

The loading implicitly to my understanding doesn't preclude using versioned `@Library` statements, and I have yet to see any usages of versioned loading of the library from plugins. I also think that disabling load implicitly at this point would break everybody's Pipelines far more seriously than incremental changes.


I'll be in the office Monday and will take a more thorough look at this then before merging.

bitwiseman:
I can think of a couple ways to work around the local test-bad approach, at least for pull requests, but none of are very pretty.

jglick:
> The loading implicitly to my understanding doesn't preclude using versioned `@Library` statements

Well, yes and no. The problem is if I want to test a forked PR like this one. I cannot specify that as a version, since AFAICT `github-branch-source` (which I assume you are using here) has no mechanism for resolving a version string referring to a PR. (`https://github.com/origin-org/repo/commit/1234abcd` URLs *will* display a commit from a fork, but this is not what the internal APIs use.) So I would need to use the `@Library('github.com/jglick/pipeline-library@1234abcd')` syntax. This works, but if the library definition is implicit, the official version will _also_ be loaded, and they will clash (cannot recall offhand which one “wins”).

So to allow people to really test a forked PR properly you must turn off the implicit flag. I have run into this problem before; the workaround was to temporarily rename to `vars/buildPlugin2.groovy` in the PR, test it using `buildPlugin2(…)` in `Jenkinsfile`, then rename back once the tests pass. Possible but very clumsy.

> I have yet to see any usages of versioned loading of the library from plugins.

Well that is a matter of the expectations you have set so far. The official guide to using ci.jenkins.io does not suggest using versions.

> disabling load implicitly at this point would break everybody's Pipelines

If you decided to go this route, it would be necessary to have a transitional period during which at least the `master` branch of every plugin currently being built would have to be updated to specify an explicit `@Library` (or `library` step also works since there is no `src/` dir), and optionally also a version.

jglick:
> it would be necessary to have a transitional period

Since the “floodgates” were opened with no mention of this, I suppose we are not going this route, meaning that we need to think about alternate ways of testing patches to this library, especially as any apparently minor edit could suddenly break dozens of builds. Cf. https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/37.

PR #13, opened 2017-06-30
Static analysis

rsandell:
Added `checkstyle` and `findbugs` configuration.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -32,13 +32,13 @@ buildPlugin()
 * `findbugs`: (default: `null`) - a map of parameters to run findbugs and/or archive findbugs reports. (only available for Maven projects)
 ** `run`: set to `true` to add the `findbugs:findbugs` goal to the maven command.
 ** `archive`: set to `true` to collect the findbugs report.
-** `pattern`: (default: `'**/findbugsXml.xml'`) the file pattern for the report.
+** `pattern`: (default: `'**/target/findbugsXml.xml'`) the file pattern for the report.
 ** `unstableTotalAll`: (default: `null`) the warnings threshold for when to mark the build as unstable, see the `findbugs` step for additional details.
 ** `unstableNewAll`:  (default: `null`) the new warnings threshold for when to mark the build as unstable, see the `findbugs` step for additional details.
 * `checkstyle`: (default: `null`) - a map of parameters to run checkstyle and/or archive checkstyle reports. (only available for Maven projects)
 ** `run`: set to `true` to add the `checkstyle:checkstyle` goal to the maven command.
 ** `archive`: set to `true` to collect the checkstyle report.
-** `pattern`: (default: `'**/findbugsXml.xml'`) the file pattern for the report.
+** `pattern`: (default: `'**/target/checkstyle-result.xml'`) the file pattern for the report.
 ** `unstableTotalAll`: (default: `null`) the violations threshold for when to mark the build as unstable, see the `checkstyle` step for additional details.
 ** `unstableNewAll`:  (default: `null`) the new violations threshold for when to mark the build as unstable, see the `checkstyle` step for additional details.
 
@@ -49,7 +49,7 @@ Usage:
 .Jenkinsfile
 [source,groovy]
 ----
-buildPlugin(platforms: ['linux'], jdkVersions: [7, 8], findbugs: [run: true, archive: true, unstableTotalAll: '0'], checkstyle: [archive: true])
+buildPlugin(platforms: ['linux'], jdkVersions: [7, 8], findbugs: [archive: true, unstableTotalAll: '0'], checkstyle: [run: true, archive: true])
 ----
 
 === infra.isTrusted()
--- a/README.adoc
+++ b/README.adoc
@@ -29,13 +29,27 @@ buildPlugin()
   execute the steps against in parallel
 * `jenkinsVersions`: (default: `[null]`) - a matrix of Jenkins baseline versions to build/test against in parallel (null means default,
   only available for Maven projects)
+* `findbugs`: (default: `null`) - a map of parameters to run findbugs and/or archive findbugs reports. (only available for Maven projects)
+** `run`: set to `true` to add the `findbugs:findbugs` goal to the maven command.
+** `archive`: set to `true` to collect the findbugs report.
+** `pattern`: (default: `'**/findbugsXml.xml'`) the file pattern for the report.
+** `unstableTotalAll`: (default: `null`) the warnings threshold for when to mark the build as unstable, see the `findbugs` step for additional details.
+** `unstableNewAll`:  (default: `null`) the new warnings threshold for when to mark the build as unstable, see the `findbugs` step for additional details.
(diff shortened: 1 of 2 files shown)


rsandell:
@reviewbybees

rtyler:
whoa

README.adoc
```diff
@@ -29,13 +29,27 @@ buildPlugin()
   execute the steps against in parallel
 * `jenkinsVersions`: (default: `[null]`) - a matrix of Jenkins baseline versions to build/test against in parallel (null means default,
   only available for Maven projects)
+* `findbugs`: (default: `null`) - a map of parameters to run findbugs and/or archive findbugs reports. (only available for Maven projects)
+** `run`: set to `true` to add the `findbugs:findbugs` goal to the maven command.
```
oleg-nenashev: I see no reason in not running it if there are files matching the pattern

vars/buildPlugin.groovy
```diff
@@ -59,7 +64,23 @@ def call(Map params = [:]) {
...
+                                }
+                                if (runCheckstyle || archiveCheckstyle) {
+                                    mavenOptions += "-Dcheckstyle.failOnViolation=false -Dcheckstyle.failsOnError=false"
+                                }
+                                List<String> mavenGoals = [
+                                        'clean',
+                                        'install',
+                                ]
+                                if (runFindbugs) {
+                                    mavenGoals += "findbugs:findbugs"
```
jglick: Huh? FindBugs is run by default already.

vars/buildPlugin.groovy
```diff
@@ -59,7 +64,23 @@ def call(Map params = [:]) {
...
+                                }
+                                if (runCheckstyle) {
+                                    mavenGoals += "checkstyle:checkstyle"
+                                }
+                                command = "mvn ${mavenOptions.join(' ')} ${mavenGoals.join(' ')}"
```
jglick: Simplest to use a single list for everything.

vars/buildPlugin.groovy
```diff
@@ -95,6 +116,26 @@ def call(Map params = [:]) {
...
+                                }
+                                if (params?.findbugs?.unstableTotalAll) {
+                                    fp['unstableTotalAll'] ="${params.findbugs.unstableTotalAll}"
+                                }
+                                findbugs(fp)
+                            }
+                            if (isMaven && archiveCheckstyle) {
+                                def cp = [pattern: params?.checkstyle?.pattern ?: '**/checkstyle-result.xml']
+                                if (params?.checkstyle?.unstableNewAll) {
+                                    cp['unstableNewAll'] ="${params.checkstyle.unstableNewAll}"
```
jglick: `"${params.checkstyle.unstableNewAll}"` is a silly way of writing `params.checkstyle.unstableNewAll`.

README.adoc
jglick: Rather, you do not need to `run` it because it is already run by default. Rather we want to archive the results by default. (And allow that plugin to mark the build as `UNSTABLE` rather than `FAILURE`.)

vars/buildPlugin.groovy
```diff
@@ -95,6 +116,26 @@ def call(Map params = [:]) {
                             }
 
                             junit testReports // TODO do this in a finally-block so we capture all test results even if one branch aborts early
+                            if (isMaven && archiveFindbugs) {
+                                def fp = [pattern: params?.findbugs?.pattern ?: '**/findbugsXml.xml']
```
jglick: I think you can select a more specific pattern match. Certainly results will be inside `**/target/` somewhere.

vars/buildPlugin.groovy
```diff
@@ -95,6 +116,26 @@ def call(Map params = [:]) {
...
+                                if (params?.findbugs?.unstableNewAll) {
+                                    fp['unstableNewAll'] ="${params.findbugs.unstableNewAll}"
+                                }
+                                if (params?.findbugs?.unstableTotalAll) {
+                                    fp['unstableTotalAll'] ="${params.findbugs.unstableTotalAll}"
+                                }
+                                findbugs(fp)
+                            }
+                            if (isMaven && archiveCheckstyle) {
+                                def cp = [pattern: params?.checkstyle?.pattern ?: '**/checkstyle-result.xml']
```
jglick: ditto

README.adoc
```diff
@@ -29,13 +29,27 @@ buildPlugin()
...
+** `archive`: set to `true` to collect the findbugs report.
+** `pattern`: (default: `'**/findbugsXml.xml'`) the file pattern for the report.
```
jglick: Why is this even customizable? Maven always puts the results in a predictable place.

README.adoc
```diff
@@ -29,13 +29,27 @@ buildPlugin()
...
+** `unstableTotalAll`: (default: `null`) the warnings threshold for when to mark the build as unstable, see the `findbugs` step for additional details.
+** `unstableNewAll`:  (default: `null`) the new warnings threshold for when to mark the build as unstable, see the `findbugs` step for additional details.
+* `checkstyle`: (default: `null`) - a map of parameters to run checkstyle and/or archive checkstyle reports. (only available for Maven projects)
+** `run`: set to `true` to add the `checkstyle:checkstyle` goal to the maven command.
+** `archive`: set to `true` to collect the checkstyle report.
+** `pattern`: (default: `'**/findbugsXml.xml'`) the file pattern for the report.
```
jglick: ditto

vars/buildPlugin.groovy
oleg-nenashev: In Parent POM 2, but I doubt we care about anything else

rsandell: That's why I added the option to add it, but it's not added by default.

vars/buildPlugin.groovy
rsandell: The snippetizer wanted a String, so I felt this was less silly to do than `String.valueOf(arams.checkstyle.unstableNewAll)`

README.adoc
rsandell: Not if you tell it to put it someplace else ;)

rsandell:
From what I've seen there are plenty of plugin maintainers that don't care about their findbugs results, so if we turn on archiving by default (with mark as unstable) I see a risk that we will break a lot of builds. that's why I added opt-in on everything. `-Dfindbugs.failOnError=false` Is added if run *or* archive is enabled.

And yes I do want checkstyle since most of the plugins I maintain have checkstyle rules that fails the build if there are any violations, and I want to be able to read those violations without scrolling through the build log.

jglick:
> there are plenty of plugin maintainers that don't care about their findbugs results, so if we turn on archiving by default (with mark as unstable) I see a risk that we will break a lot of builds

Perhaps we can archive by default unless the POM specifies `findbugs.failOnError=false` (which you can checking via `help:evaluate`)?

rsandell:
I think it's better to just let people opt in to it.

rsandell:
@reviewbybees done

@rtyler anything more you need to merge this? I kinda wanna use it asap as most of my plugins currently have no PR verification because no Jenkinsfile :)

oleg-nenashev:
FYI I tested it manually a couple of weeks ago, and I confirm it works as expected. :shipit:

PR #17, opened 2017-07-22
[INFRA-1277] - Do not use settings-azure.xml when running outside Azure.

oleg-nenashev:
The script may be invoked from outside, e.g. during the patch development for it.
This change just disables the logic when it runs at one of Jenkins instances within our Azure infrastructure. Just a money saver.

@rtyler @jglick

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -1,7 +1,7 @@
 #!/usr/bin/env groovy
 
 Boolean isRunningOnJenkinsInfra() {
-    return env.JENKINS_URL == 'https://ci.jenkins.io' || isTrusted()
+    return env.JENKINS_URL == 'https://ci.jenkins.io/' || isTrusted()
 }
 
 Boolean isTrusted() {
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -1,11 +1,7 @@
 #!/usr/bin/env groovy
 
-Boolean isRunsOnJenkinsAzure() {
-    return isJenkinsIO() || isTrusted()
-}
-
-Boolean isJenkinsIO() {
-    return env.JENKINS_URL == 'https://ci.jenkins.io'
+Boolean isRunningOnJenkinsInfra() {
+    return env.JENKINS_URL == 'https://ci.jenkins.io' || isTrusted()
 }
 
 Boolean isTrusted() {
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -1,5 +1,13 @@
 #!/usr/bin/env groovy
 
+Boolean isRunsOnJenkinsAzure() {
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Also @reviewbybees, I suppose

vars/infra.groovy
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env groovy
 
+Boolean isRunsOnJenkinsAzure() {
```
daniel-beck: I like a different shade of red better: `isRunningOnJenkinsInfra` both for the nicer name in Java, still being a nice Groovy property (`runningOnJenkinsInfra`) and making less unnecessary assumptions over where the infra actually is hosted.

vars/infra.groovy
```diff
@@ -1,5 +1,13 @@
...
+    return isJenkinsIO() || isTrusted()
+}
+
+Boolean isJenkinsIO() {
```
daniel-beck: I don't think this one is useful. It should be rare enough that `runningOnJenkinsInfra && !trusted` works for this purpose just as well, and we don't introduce an ambiguity with the name: note how trusted.ci also is a `jenkins.io` domain.

vars/infra.groovy
oleg-nenashev: I was thinking about that, but I concluded that we may change the infra layout at some point (e.g. we move to JustAnotherCloudProvider). The `buildPipeline()` logic is specific to Azure, so I concluded it's better to have "Azure" in the method name

daniel-beck: >  The `buildPipeline()` logic is specific to Azure, so I concluded it's better to have "Azure" in the method name

Assuming you mean `buildPlugin()`, then yes, it has behavior specific to Azure, but that's limited to Jenkins project infra. This method becomes public API for the pipeline library. Notably, we can update the Pipeline library if we ever change project hosting as needed, without this method having to return `false` unconditionally (because there's no Jenkins Azure infra anymore).

oleg-nenashev:
@daniel-beck @rtyler Comments addressed

oleg-nenashev:
@reviewbybees

vars/infra.groovy
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env groovy
 
+Boolean isRunningOnJenkinsInfra() {
+    return env.JENKINS_URL == 'https://ci.jenkins.io' || isTrusted()
```
jglick: Missing trailing slash.

oleg-nenashev:
@reviewbybees ping

rtyler:
*shrug* let's hope this doesn't break a bunch of shit :smiling_imp:

PR #18, opened 2017-07-25
Artifact types

slide:
This adds the ability for people to define the artifact types that should be archived. Some "plugins" are actually libraries and generate jar files instead of jpi or hpi.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -92,7 +92,6 @@ def call(Map params = [:]) {
                                 for(m = 0; m < artifactTypes.size(); m++) {
                                     artifactTypes[m] = "**/target/*.${artifactTypes[m]}"
                                 }
-
                             } else {
                                 testReports = '**/build/test-results/**/*.xml'
                                 for(m = 0; m < artifactTypes.size(); m++) {


rtyler:
I'm not opposed to this change, though we're scope creeping what `buildPlugin` is supposed to do :smiling_imp:

Do you have some examples to reference as a use-case? I'm not clear on which projects this would be used by?

slide:
I'm more than happy to pull out the common elements for plugin and library into a "buildCommon" which is called by buildPlugin/buildLibrary if that is a more desired approach. One "plugin" that uses jar artifacts is publish-over-plugin, which is not really a plugin, but a common library for the other publish-over-XXX plugins.

oleg-nenashev:
Another example are Jenkins Modules, which are packaged as JARs. Since https://github.com/jenkins-infra/pipeline-library/pull/6 has been rejected (mostly my bad), right now we use `buildPlugin()` for their CI runs

uhafner:
Here is another example where the build fails because no hpi file is found: [](https://ci.jenkins.io/job/Plugins/job/analysis-model/job/master/1/console)

At least we should use the parameter `allowEmptyArchive: true`

oleg-nenashev:
I have created #56 for libraries support. Maybe it would be better to use a separate method since we are adding more and more logic around `buildPlugin()`.

WDYT @slide @uhafner ?

PR #19, opened 2017-08-03
Apply timeout only inside each node block

jglick:
https://github.com/jenkinsci/workflow-multibranch-plugin/pull/67 failed simply because the Windows node [took 52½m to be scheduled](https://ci.jenkins.io/job/Plugins/job/workflow-multibranch-plugin/job/PR-67/1/execution/node/18/log/), and then the remaining 7½m was not long enough to run the build, especially since most of that was downloading artifacts. We should apply the 1h timeout only to the actual build process once the executor slot is available.

@reviewbybees

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -20,6 +20,7 @@ def call(Map params = [:]) {
 
                 tasks[stageIdentifier] = {
                     node(label) {
+                        timeout(60) {
                         boolean isMaven
 
                         stage("Checkout (${stageIdentifier})") {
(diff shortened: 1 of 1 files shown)


vars/buildPlugin.groovy
```diff
@@ -20,6 +20,7 @@ def call(Map params = [:]) {
 
                 tasks[stageIdentifier] = {
                     node(label) {
+                        timeout(60) {
                         boolean isMaven
```
jglick: Deliberately not reindenting, to avoid a massive diff.

MarkEWaite:
Thank you, thank you, thank you!  

My infrastructure at home has 50% windows machines and 50% linux machines, but the Windows builds are so much slower that my git plugin and git client plugin jobs timeout consistently if I run more than 5 of them at once.

PR #21, opened 2017-12-03
Update builds: Refactoring, Parallel test support, Coverages, etc.

oleg-nenashev:
This is a WiP pull request, which updates the Pipeline library a bit. My main idea is to have a more comprehensive Pipeline library demo, which also offers more features on jenkins.io.

- [x] - Refactor buildPlugin and introduce buildMavenPluginPom and buildGradleComponent methods
- [x] - Introduce the `repoType` parameter, which enforces Maven or Gradle without doing the repo checkout. Ideally `buildMavenPluginPom()` should fallback to radle so there will be no extra `node()` block in the default logic
- [x] - Add support of parallel testing in Maven projects.
- [x] - Add stages
- [x] - Add Cobertura sample for coverage publishing
- [ ] - Add `jacoco` support forCoverage publishing, which is supported OOTB in Plugin POMs
- [ ] - Add tests using Pipeline Unit

Pipeline for a single platform:

screen shot 2017-12-03 at 19 05 54

Multi-platform Pipeline:

screen shot 2017-12-03 at 19 09 35

Multi-platform Pipeline with parallel tests:

screen shot 2017-12-03 at 19 19 47

The latter graph looks strange, CC @abayer @svanoort. I'd guess BlueOcean does not support parallel-in-parallel.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -9,7 +9,32 @@ def call(Map params = [:]) {
     def jenkinsVersions = params.containsKey('jenkinsVersions') ? params.jenkinsVersions : [null]
     def repo = params.containsKey('repo') ? params.repo : null
     def failFast = params.containsKey('failFast') ? params.failFast : true
+    def testParallelism = params.containsKey('testParallelism') ? params.testParallelism : 1
+    def projectType = params.containsKey('projectType') ? params.projectType : null
     Map tasks = [failFast: failFast]
+
+    boolean isMaven
+    //TODO: probably it makes sense to default to Maven instead
+    if (projectType == null) {
+        stage("Determine the project type") {
+            timeout(10) {
+                node("docker") {
+                    commonSteps.checkout(repo)
+                    if (fileExists('pom.xml')) {
+                        projectType = "maven"
+                        isMaven = true
+                    } else if (fileExists('build.gradle')) {
+                        projectType = "gradle"
+                    } else {
+                        error("Cannot determine the project type")
+                    }
+                }
+            }
+        }
+    }
+    echo "Project type is ${projectType}"
+
+    String firstStageIdentifier
     for (int i = 0; i < platforms.size(); ++i) {
         for (int j = 0; j < jdkVersions.size(); ++j) {
             for (int k = 0; k < jenkinsVersions.size(); ++k) {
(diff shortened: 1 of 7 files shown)


rtyler:
@oleg-nenashev This work looks really interesting, assuming it's working, are there any reasons not to merge this before you check off those other two checkboxes?

oleg-nenashev:
@rtyler It works even now (on https://github.com/oleg-nenashev/demo-jenkins-config-as-code). The main problem in the current implementation is that the Maven plugin build will require an extra `node()` block (and hence Docker provisioning) checkout to determine the project type before invoking the command.

It is not very efficient, but I want to keep Gradle projects running without breakages. So I consider adding a hack, which would fallback to Gradle from a Maven project if there is no pom.xml. In such case we would be able to keep `buildPlugin()` compatibility without extra `node()` clause.

Generally I had a plan to get CI for `pipeline-library` running before requesting the review. But we can make it iterative, of course

vars/buildPlugin.groovy
```diff
@@ -9,7 +9,32 @@ def call(Map params = [:]) {
...
     def failFast = params.containsKey('failFast') ? params.failFast : true
+    def testParallelism = params.containsKey('testParallelism') ? params.testParallelism : 1
+    def projectType = params.containsKey('projectType') ? params.projectType : null
     Map tasks = [failFast: failFast]
+
+    boolean isMaven
+    //TODO: probably it makes sense to default to Maven instead
+    if (projectType == null) {
+        stage("Determine the project type") {
+            timeout(10) {
```
oleg-nenashev: @rtyler The code below is my main concern about the current impl

oleg-nenashev:
CC @reviewbybees

svanoort:
@oleg-nenashev Very belated reply: Blue Ocean don't and apparently will not support nested parallels.  Bismuth does, explicitly, because I'd anticipated it -- but AFAICT nobody has offered visualization of such structures.

Wacky things may happen with nesting of parallels :-/

chriskilding:
I’m interested in progress on the code coverage front - I’ve thought for some time that plugins in the Jenkins project should have access to a standardised coverage reporter out of the box (if they want it), rather than individual plugin authors having to set up their own SaaS coverage reporter every time.

How close are you to enabling support for Jacoco / Cobertura reports?

oleg-nenashev:
Please see https://jenkins.io/blog/2018/08/17/code-coverage-api-plugin-1/ .
Unifying Code coverage reporters was one of the projects during GSoC 2018.
All APIs are there now, but some plugin integrations are needed

dduportal:
Hello 👋 
This pull request is dangling since 3 years, has the "draft" status since 11 month and has conflicting file.
That's why we are closing it. 

Of course, feel free to reopen if you are still working on it. Many thanks for the contribution proposal!

PR #25, opened 2018-02-14
Do not archive build products from nondefault Jenkins versions

jglick:
Someone (?) noticed that the `*.hpi` download links from PRs builds using the `jenkinsVersions` option (e.g., https://github.com/jenkinsci/tikal-multijob-plugin/pull/129) could randomly (depending on branch completion order) include a plugin built against a newer Jenkins core baseline, making the snapshot unusable in older versions including the baseline actually declared in the POM. We should only archive plugin builds from the official baseline.

As usual, untested.

@reviewbybees


jglick:
@reviewbybees done

PR #28, opened 2018-02-21
[WIP] [INFRA-1489] runATH step

raul-arabaolaza:
WORK IN PROGRESS

[INFRA-1489](https://issues.jenkins-ci.org/browse/INFRA-1489)

Created this PR for tracking purposes by request from @jglick Any kind of feedback is welcome and encouraged

Would be great if someone with permissions can assign the proper work-in-progress  label, I am not able

### DIFF
--- a/vars/runATH.groovy
+++ b/vars/runATH.groovy
@@ -5,10 +5,10 @@
  */
 
 def call(Map params = [:]) {
-    def String athUrl = params.get('athUrl', 'https://github.com/jenkinsci/acceptance-test-harness.git')
-    def String athRevision = params.get('athRevision', 'master')
-    def String metadataFile = params.get('metadataFile', 'essentials.yml')
-    def String jenkins = params.get('jenkins', 'latest')
+    def athUrl = params.get('athUrl', 'https://github.com/jenkinsci/acceptance-test-harness.git')
+    def athRevision = params.get('athRevision', 'master')
+    def metadataFile = params.get('metadataFile', 'essentials.yml')
+    def jenkins = params.get('jenkins', 'latest')
 
     def mirror = "http://mirrors.jenkins.io/"
     def defaultCategory = "org.jenkinsci.test.acceptance.junit.SmokeTest"
--- a/vars/runATH.groovy
+++ b/vars/runATH.groovy
@@ -24,7 +24,7 @@ def call(Map params = [:]) {
 
     def skipExecution = false
 
-    def localPluginsStashName = env.RUN_ATH_LOCAL_PLUGINS_STASH_NAME
+    def localPluginsStashName = env.RUN_ATH_LOCAL_PLUGINS_STASH_NAME ?: "localPlugins"
 
     ensureInNode(env, env.RUN_ATH_SOURCES_AND_VALIDATION_NODE ?: "linux", {
         List<String> env = [
@@ -146,7 +146,7 @@ def call(Map params = [:]) {
                 if (supportedBrowsers.contains(browser)) {
 
                     def currentBrowser = browser
-                    def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -e LOCAL_SNAPSHOTS=${localSnapshots} -e SHARED_DOCKER_SERVICE=true -e EXERCISEDPLUGINREPORTER=textfile -e PLUGIN_DIR=localPlugins -u ath-user"
+                    def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -e SHARED_DOCKER_SERVICE=true -e EXERCISEDPLUGINREPORTER=textfile -e PLUGINS_DIR=localPlugins -u ath-user"
                     def commandBase = "./run.sh ${currentBrowser} ./jenkins.war -Dmaven.test.failure.ignore=true -DforkCount=1 -B -Dsurefire.rerunFailingTestsCount=${rerunCount}"
                     if (infra.isRunningOnJenkinsInfra()) {
                         def settingsXml = "${pwd tmp: true}/settings-azure.xml"
@@ -160,7 +160,7 @@ def call(Map params = [:]) {
                                 unstash name: "athSources"
                                 unstash name: "jenkinsWar"
                                 dir("localPlugins") {
-                                    if (localPluginsStashName) {
+                                    if (localSnapshots && localPluginsStashName) {
                                         unstash name: localPluginsStashName
                                     }
                                 }
@@ -181,7 +181,7 @@ def call(Map params = [:]) {
                                 unstash name: "athSources"
                                 unstash name: "jenkinsWar"
                                 dir("localPlugins") {
-                                    if (localPluginsStashName) {
+                                    if (localSnapshots && localPluginsStashName) {
                                         unstash name: localPluginsStashName
                                     }
                                 }
--- a/vars/runATH.groovy
+++ b/vars/runATH.groovy
@@ -135,7 +135,8 @@ def call(Map params = [:]) {
             def browsers = metadata.browsers ?: ["firefox"]
             def failFast = metadata.failFast ?: false
             def rerunCount = metadata.rerunFailingTestsCount ?: 0
-            def localSnapshots = metadata.useLocalSnapshots ?: true
+            // Elvis fails in case useLocalSnapshots == false in metadata File
+            def localSnapshots = metadata.useLocalSnapshots != null ? metadata.useLocalSnapshots : true
 
             if (testsToRun == null && categoriesToRun == null) {
                 categoriesToRun = defaultCategory
@@ -146,7 +147,7 @@ def call(Map params = [:]) {
                 if (supportedBrowsers.contains(browser)) {
 
                     def currentBrowser = browser
-                    def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -e SHARED_DOCKER_SERVICE=true -e EXERCISEDPLUGINREPORTER=textfile -e PLUGINS_DIR=localPlugins -u ath-user"
+                    def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -e SHARED_DOCKER_SERVICE=true -e EXERCISEDPLUGINREPORTER=textfile -u ath-user"
                     def commandBase = "./run.sh ${currentBrowser} ./jenkins.war -Dmaven.test.failure.ignore=true -DforkCount=1 -B -Dsurefire.rerunFailingTestsCount=${rerunCount}"
                     if (infra.isRunningOnJenkinsInfra()) {
                         def settingsXml = "${pwd tmp: true}/settings-azure.xml"
@@ -157,41 +158,16 @@ def call(Map params = [:]) {
                     if (testsToRun) {
                         testingbranches["ATH individual tests-${currentBrowser}"] = {
                             dir("test${currentBrowser}") {
-                                unstash name: "athSources"
-                                unstash name: "jenkinsWar"
-                                dir("localPlugins") {
-                                    if (localSnapshots && localPluginsStashName) {
-                                        unstash name: localPluginsStashName
-                                    }
-                                }
-                                def command = commandBase + " -Dtest=${testsToRun}"
-
-                                athContainerImage.inside(containerArgs) {
-                                    realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
-                                        sh 'eval "$(./vnc.sh)" && ' + command
-                                    }
-
-                                }
+                                def discriminator = "-Dtest=${testsToRun}"
+                                test(discriminator, commandBase, localSnapshots, localPluginsStashName, containerArgs, athContainerImage)
                             }
                         }
                     }
                     if (categoriesToRun) {
                         testingbranches["ATH categories-${currentBrowser}"] = {
                             dir("categories${currentBrowser}") {
-                                unstash name: "athSources"
-                                unstash name: "jenkinsWar"
-                                dir("localPlugins") {
-                                    if (localSnapshots && localPluginsStashName) {
-                                        unstash name: localPluginsStashName
-                                    }
-                                }
-                                def command = commandBase + " -Dgroups=${categoriesToRun}"
-                                athContainerImage.inside(containerArgs) {
-                                    realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
-                                        sh 'eval "$(./vnc.sh)" && ' + command
-                                    }
-
-                                }
+                                def discriminator = "-Dgroups=${categoriesToRun}"
+                                test(discriminator, commandBase, localSnapshots, localPluginsStashName, containerArgs, athContainerImage)
                             }
                         }
                     }
@@ -204,6 +180,39 @@ def call(Map params = [:]) {
         }
 
     })
+}
+
+private void test(discriminator, commandBase, localSnapshots, localPluginsStashName, containerArgs, athContainerImage) {
+    unstashResources(localSnapshots, localPluginsStashName)
+    athContainerImage.inside(containerArgs) {
+        realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
+            sh 'eval "$(./vnc.sh)" && ' + prepareCommand(commandBase, discriminator, localSnapshots, localPluginsStashName)
+        }
+    }
+}
+
+private String prepareCommand(commandBase, discriminator, localSnapshots, localPluginsStashName ) {
+    def command = commandBase + " ${discriminator}"
+    if (localSnapshots && localPluginsStashName) {
+        command = "LOCAL_JARS=${getLocalPluginsList()} " + command
+    }
+    command
+}
+
+private void unstashResources(localSnapshots, localPluginsStashName) {
+    unstash name: "athSources"
+    unstash name: "jenkinsWar"
+    dir("localPlugins") {
+        if (localSnapshots && localPluginsStashName) {
+            unstash name: localPluginsStashName
+        }
+    }
+}
+
+private String getLocalPluginsList() {
+    dir("localPlugins") {
+       return sh(script : "ls -p -d -1 ${pwd()}/*.* | tr '\n' ':'| sed 's/.\$//'", returnStdout: true).trim()
+    }
 }
 
 private void ensureInNode(env, nodeLabel, body) {


vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env groovy
+
+/**
+ * Simple wrapper for running the ATH
+ */
+
+def call(Map params = [:]) {
+
+    def String athUrl = params.get('athUrl', 'https://github.com/jenkinsci/acceptance-test-harness.git')
+    def String athRevision = params.get('athRevision', 'master')
```
jglick: Recommend no default value. You should have to call out to reviewers that you are using a branch, so they can reject your PR until you fix that.

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+    def String metadataFile = params.get('metadataFile', 'essentials.yml')
+    def String jenkins = params.get('jenkins', 'latest')
```
jglick: Ditto. Specifying, say, `2.89.4` is hardly an imposition. When you decide to upgrade, the PR which proposes doing so will let you know whether the test still passes against that core.

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+
+    def isLocalATH = athUrl.startsWith("file://")
+    def isVersionNumber = (jenkins =~ /^(\d+\.)?(\d+\.)?(\*|\d+)$/).matches()
```
jglick: What is the `*` for?

And why are you hardcoding the number of digits? Just:

```regexp
\d+([.]\d+)*
```

would suffice I think.

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+    // Workaround for https://issues.jenkins-ci.org/browse/JENKINS-27092
+    def shouldStop = false
+    // End of workaround
+
+    def mirror = "http://mirrors.jenkins.io/"
```
jglick: Please do not use the mirror sites to download artifacts during CI builds. This will strain our servers.

Better to obtain `jenkins.war` via `mvn dependency:get`. There may or may not be any local caching happening (probably each build on ci.jenkins.io gets a fresh local repo) but at least we can make it go through

```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
    <mirrors>
        <mirror>
            <id>azure</id>
            <url>https://repo.azure.jenkins.io/public/</url> <!-- INFRA-1176 -->
            <mirrorOf>repo.jenkins-ci.org</mirrorOf>
        </mirror>
    </mirrors>
</settings>
```

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+    def defaultCategory = "org.jenkinsci.test.acceptance.junit.SmokeTest"
```
jglick: There should be no default for this, I think, because that would block you from running tests without this annotation IIUC.

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+            }
+            if (metadata.browsers == null) {
+                echo "The provided metadata file does not include the browsers property, using firefox as default"
+            }
+        }
+        if (!isLocalATH) {
+            def canConnectToGitRepo = sh(script: "git ls-remote --exit-code -h ${athUrl}", returnStatus: true) == 0
+            if (!canConnectToGitRepo) {
+                error "Is not possible to connect to the given athUrl, please review it. Current value for athUrl is ${athUrl}"
+            }
```
jglick:
```groovy
echo 'Checking connectivity to ATH sources…'
sh "git ls-remote --exit-code -h ${athUrl}"
```

would suffice I think.

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+        if (isVersionNumber) {
+            jenkinsURl = mirror + "war/${jenkins}/jenkins.war"
+        } else if (jenkins == "latest") {
+            jenkinsURl = mirror + "war/latest/jenkins.war"
+        } else if (jenkins == "latest-rc") {
+            jenkinsURl = mirror + "/war-rc/latest/jenkins.war"
+        } else if (jenkins == "lts") {
+            jenkinsURl = mirror + "war-stable/latest/jenkins.war"
+        } else if (jenkins == "lts-rc") {
+            jenkinsURl = mirror + "war-stable-rc/latest/jenkins.war"
```
jglick: I do not see any compelling reason to even offer these options.

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        def canFindJenkinsWar = sh(script: "curl --output /dev/null --silent --fail -r 0-0 -L '${jenkinsURl}'", returnStatus: true) == 0
```
jglick:
```groovy
echo 'Checking whether Jenkins WAR is available…'
sh "curl -ILf '$jenkinsURl'"
```

but again better to just use Maven.

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+            error "Is not possible to find the given jenkins war file, please review it. Current value for jenkins is ${jenkins}"
+        }
+        // Validation ended
+
+        // ATH
+        if (isLocalATH) { // Deal with already existing ATH sources
+            athSourcesFolder = athUrl - "file://"
+        } else {
+            dir(athSourcesFolder) {
+                checkout changelog: true, poll: false, scm: [$class           : 'GitSCM', branches: [[name: "${athRevision}"]],
```
jglick: `name: "${athRevision}"` is just a weird way of writing `name: athRevision`.

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        def testingbranches = ["failFast": failFast]
+        for (browser in browsers) {
+            // TODO remove this if else block once other browser are supported by infra
+            if (browser == "firefox") {
+
+                def currentBrowser = browser
+                def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -e LOCAL_SNAPSHOTS=${localSnapshots} -e SHARED_DOCKER_SERVICE=true -u ath-user"
+                def commandBase = "./run.sh ${currentBrowser} jenkins.war -Dmaven.test.failure.ignore=true -DforkCount=1 -B -Dsurefire.rerunFailingTestsCount=${rerunCount}"
```
jglick: BTW the `{}` are not necessary for most cases of string interpolation in Groovy.

```groovy
def commandBase = "./run.sh $currentBrowser jenkins.war -Dmaven.test.failure.ignore=true -DforkCount=1 -B -Dsurefire.rerunFailingTestsCount=$rerunCount"
```

should work the same.

vars/runATH.groovy
```diff
@@ -0,0 +1,156 @@
...
+                                realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
+                                    sh 'eval "$(./vnc.sh)" && ' + command
+                                }
+
+                            }
+                        }
+                    }
+                }
+                if (categoriesToRun) {
+                    testingbranches["ATH categories-${currentBrowser}"] = {
```
jglick: BTW see https://github.com/jenkinsci/parallel-test-executor-plugin/pull/34 and https://github.com/jenkinsci/acceptance-test-harness/pull/380. Not sure it matters for the use case here, which presumably involves only a small number of tests per build.

vars/runATH.groovy
raul-arabaolaza: That default is applied, unless a mistake on my part, only when you there are neither tests nor categories specified

```
stage("Running ATH") {
        def testsToRun = metadata.tests?.join(",")
        def categoriesToRun = metadata.categories?.join(",")
        def browsers = metadata.browsers ?: ["firefox"]
        def failFast = metadata.failFast ?: false
        def rerunCount = metadata.rerunFailingTestsCount ?: 0
        def localSnapshots = metadata.useLocalSnapshots ?: true

        if (testsToRun == null && categoriesToRun == null) {
            categoriesToRun = defaultCategory
        }
```

So you can specify any combination of individual tests and categories you want, and only in the case that you specify nothing the default is executed

vars/runATH.groovy
raul-arabaolaza: > Not sure it matters for the use case here, which presumably involves only a small number of tests per build.

Yeah, that is my understanding and the reason why I did not use parallel test executor plugin

vars/runATH.groovy
raul-arabaolaza: I find easier to read code when {} is explicitly used even if not needed. I can change that if there is a coding standard or technical reason, but I prefer to let it be like is if not.

Unless there are more people demanding the change...

vars/runATH.groovy
raul-arabaolaza: Yeah, I believe my version is more user-friendly but probably you are right and I am being too much user-friendly here without a compelling reason

vars/runATH.groovy
raul-arabaolaza: We can discuss that, but I prefer to default to master, the reason is that I want to encourage developers to take ownership of relevant ath tests, by always using a tag or commit that you know works tests may break without you noticing. And IMO that is bad, very very bad and one of the causes of the current status of the ATH (as it happened with PCT)

My ideal setup would be to use a fixed commit id for a plugin PR lifecycle (as I fully understand determinism is very important) but to use master of ath when you are building your plugin master branch

rtyler:
@raul-arabaolaza When you're comfortable with the runATH API, would you mind updating this pull request to include changes to the README which explain how to use runATH?

raul-arabaolaza:
@rtyler Of course, I am already taking notes of several examples and will copy the relevant parts of the design. Just waiting to have a final API version reviewed/approved by someone more than me

vars/runATH.groovy
```diff
@@ -0,0 +1,160 @@
...
+            categoriesToRun = defaultCategory
+        }
+
+        def testingbranches = ["failFast": failFast]
+        for (browser in browsers) {
+            // TODO remove this if else block once other browser are supported by infra
+            if (browser == "firefox") {
+
+                def currentBrowser = browser
+                def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -v \$HOME/.m2/repository:/home/ath-user/.m2/repository -e LOCAL_SNAPSHOTS=${localSnapshots} -e SHARED_DOCKER_SERVICE=true -u ath-user"
```
dduportal: The flag `-v \$HOME/.m2/repository:/home/ath-user/.m2/repository` will cause a lot of trouble across developers and CI setups, because it not portable at all (cf. https://docs.docker.com/storage/#choose-the-right-type-of-mount).
=> It expects that the path `$HOME/.m2/repository` is:

* Consistent across setups
* Always available on the file system of the machine where the Docker Engine is run. Which won't be the case with the "remote docker engine pattern (when you bind-mount `/var/run/docker.sock` or when you spawn a transient VM for hosting such an engine).

=> Those 2 assumptions are false, making this non portable.

Proposal, as we discussed this together @raul-arabaolaza , it is use an handy feature of docker, the "volumes" (cf. https://docs.docker.com/storage/volumes/).

Instead of providing an absolute file system path as source (on the left of the ':' on the string supplied to the `-v` flag), we will provide a "docker volume name":
`-v ath-maven-local-repo:/home/ath-user/.m2/repository`

* If the named volume that not exists when `docker run..` is invoked, then it is created on-the-fly by the Docker Engine, which will manage and store this data on the machine where it is running (which might be different than the machine where `docker` command is invoked).
* If the named volumes already exists, then it will be reused (which is your goal here as I understand it).



There are 2 attention points on my proposal:
* The name of the volume is a convention. But it MUST be documented for the end users whom will call this groovy code. It SHOULD be set as an argument with a default value to the `call()` function, and then the resulting variable is supplied to the -v flag. Then you provide an handy extension point to the actor in charge of populating a maven repo prior to call the ATH.
* Permissions on the volume: you cannot be 100% sure about the permission inside this maven local repo. Since Docker can "root", I propose to add a container call that will ensure the permission set is on the expected state. This would be the following code to run BEFORE the first `athContainerImage.inside()` invokation (~line 127):
```
echo "${dockerVolumeNameForMavenLocalRepo}" // We expect this variable to be set with the volume name
sh "docker run -v ${dockerVolumeNameForMavenLocalRepo}:/repo -u root --entrypoint chown ${athContainerImage} -R /repo ath-user" // the "ath-user is already known inside your built image, it is the convention so let's assume this is the default
```

vars/runATH.groovy
jglick: > tests may break without you noticing

That is why we have CI. To catch problems with…integrations. A **separate job** which builds lots of `master` branches together and looks for upcoming problems.

> use master of ath when you are building your plugin master branch

No no no, this will make builds unreproducible.

vars/runATH.groovy
jglick: Nope, just FYI.

vars/runATH.groovy
raul-arabaolaza: Thanks for this @dduportal !!

raul-arabaolaza: The named volume looks like the same approach followed in the [PCT docker image](https://github.com/jenkinsci/plugin-compat-tester) and it feels quite interesting, however, I need somehow to make sure that the node which actually builds the plugin ("linux" and "windows" in this case) use the aforementioned volume as storage and we may have version collisions if several jobs are running in parallel using the same maven repo...

dduportal: Right, but this is not related to this groovy file area of responsibilities. This is why I proposed to pass the volume name as an argument to the call method.
=> Then your Pipeline calling this groovy will play its role of "glue" by:

1. Generate a name for the docker volume, with enough discrimination to allow parallel builds and avoids conflicts on a single Docker Engine, and eventually create it (and ensure it is empty on the beginning of the build)
2. Pass this parameters to your "build plugin" function, to be sure the volume will be populated with the expected Maven cache. 
3. Pass this parameter when to the "runAth" step to have it reuse the volume.

N.B. for training, we use `"${GIT_BRANCH}-${GIT_COMMIT}"` for discrimination: it allows concurrents builds on a single Docker Engine, with still some "human readable" for when it is time to purge (example of policy: "if is does not start by "master-" and it is older than 5 days, then delete it).

vars/runATH.groovy
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Wrapper for running the ATH
+ */
+
+def call(Map params = [:]) {
+    def String athUrl = params.get('athUrl', 'https://github.com/jenkinsci/acceptance-test-harness.git')
+    def String athRevision = params.get('athRevision', 'master')
+    def String metadataFile = params.get('metadataFile', 'essentials.yml')
+    def String jenkins = params.get('jenkins', 'latest')
```
vilacides: Is this the latest LTS? master?

vars/runATH.groovy
```diff
@@ -0,0 +1,206 @@
...
+    ensureInNode(env, env.RUN_ATH_SOURCES_AND_VALIDATION_NODE ?: "linux", {
+        List<String> env = [
+                "JAVA_HOME=${tool 'jdk8'}",
+                'PATH+JAVA=${JAVA_HOME}/bin',
+                "PATH+MAVEN=${tool 'mvn'}/bin"
+
+        ]
+
+        if (!fileExists(metadataFile)) {
+            echo "Skipping ATH execution because the metadata file does not exist. Current value is ${metadataFile}."
```
vilacides: Why do you print the metadataFile value if the file doesn't exist? Isn't it going to be always null?

vars/runATH.groovy
```diff
@@ -0,0 +1,206 @@
...
+            skipExecution = true
+            return
+        }
+
+        stage("Getting ATH sources and Jenkins war") {
+            // Start validation
+            metadata = readYaml(file: metadataFile).ath
+            if (metadata == null) {
+                error "The provided metadata file seems invalid as it does not contain a valid ath section"
```
vilacides: Is it the same if it doesn't have a metadata section, if it's not valid or if it's empty? I think we should treat this as separate cases. For example if empty, maybe we want to suggest a default. If not valid we want to bring that to the developer's attention and if not existing the error message would be different as well.

vars/runATH.groovy
```diff
@@ -0,0 +1,206 @@
...
+ * Wrapper for running the ATH
+ */
+
+def call(Map params = [:]) {
+    def String athUrl = params.get('athUrl', 'https://github.com/jenkinsci/acceptance-test-harness.git')
+    def String athRevision = params.get('athRevision', 'master')
+    def String metadataFile = params.get('metadataFile', 'essentials.yml')
+    def String jenkins = params.get('jenkins', 'latest')
+
+    def mirror = "http://mirrors.jenkins.io/"
```
vilacides: Do we foresee any connectivity issues when using mirrors?

vars/runATH.groovy
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env groovy
+
```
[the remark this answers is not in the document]
raul-arabaolaza: Yeah, will do that and fully document this once is implementation complete

vars/runATH.groovy
raul-arabaolaza: latest weekly

vars/runATH.groovy
```diff
@@ -0,0 +1,206 @@
...
+                        def settingsXml = "${pwd tmp: true}/settings-azure.xml"
+                        writeFile file: settingsXml, text: libraryResource('settings-azure.xml')
+                        commandBase = commandBase + " -s ${settingsXml}"
+                    }
+
+                    if (testsToRun) {
+                        testingbranches["ATH individual tests-${currentBrowser}"] = {
+                            dir("test${currentBrowser}") {
+                                unstash name: "athSources"
+                                unstash name: "jenkinsWar"
```
vilacides: Maybe this is a silly question, but how do you map this with the already set jenkins.war passed in the commandBase?

vars/runATH.groovy
```diff
@@ -0,0 +1,206 @@
...
+                                    realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
+                                        sh 'eval "$(./vnc.sh)" && ' + command
+                                    }
+
+                                }
+                            }
+                        }
+                    }
+                } else {
+                    echo "${browser} is not yet supported"
```
vilacides: Haven't we already assessed this in the beginning and defaulted to firefox?

vars/runATH.groovy
vilacides: Maybe worth clarifying/documenting that. Thanks!

vars/runATH.groovy
vilacides: Ah wait, only if empty in line 51. Then I wonder if we should do all the validations there (empty and not supported)

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/env groovy
+
+/**
+ * Wrapper for running the ATH
+ */
+
+def call(Map params = [:]) {
+    def String athUrl = params.get('athUrl', 'https://github.com/jenkinsci/acceptance-test-harness.git')
```
burnMyDread: This is unnecessary on two levels. You never need both String and an object type and you don't need to call get on an element in a map in groovy. The preferred way to do it is via the groovy properties, http://groovy-lang.org/objectorientation.html has more details.

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+    def String athRevision = params.get('athRevision', 'master')
+    def String metadataFile = params.get('metadataFile', 'essentials.yml')
+    def String jenkins = params.get('jenkins', 'latest')
+
+    def mirror = "http://mirrors.jenkins.io/"
+    def defaultCategory = "org.jenkinsci.test.acceptance.junit.SmokeTest"
```
burnMyDread: You are being inconsistent in your style here, you are using concrete objects earlier but Object here.

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+                "JAVA_HOME=${tool 'jdk8'}",
+                'PATH+JAVA=${JAVA_HOME}/bin',
+                "PATH+MAVEN=${tool 'mvn'}/bin"
+
+        ]
+
+        if (!fileExists(metadataFile)) {
+            echo "Skipping ATH execution because the metadata file does not exist. Current value is ${metadataFile}."
+            skipExecution = true
+            return
```
burnMyDread: Why?

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        stage("Getting ATH sources and Jenkins war") {
+            // Start validation
+            metadata = readYaml(file: metadataFile).ath
+            if (metadata == null) {
```
burnMyDread: I would pull this out into a method, this is muddying the waters as to what the intent of the section is doing.

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+
+            // Allow override of jenkins version from metadata file
+            jenkins = metadata.jenkins ?: jenkins
+            isVersionNumber = (jenkins =~ /^\d+([.]\d+)*$/).matches()
+
+            if (!isLocalATH) {
+                echo 'Checking connectivity to ATH sources…'
+                sh "git ls-remote --exit-code -h ${athUrl}"
+            }
+            if (jenkins == "latest") {
```
burnMyDread: Pull this out into a method, so you can test it.

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+                                                                 userRemoteConfigs: [[url: athUrl]]]
+                }
+            }
+            dir(athSourcesFolder) {
+                // We may need to run things in parallel later, so avoid several checkouts
+                stash name: "athSources"
+            }
+
+            // Jenkins war
+            if (isVersionNumber) {
```
burnMyDread: This code needs more comments as to what this code is doing. It looks like you are downloading a war, but if that is the intent please say that.

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+                sh("curl -o jenkins.war -L ${jenkinsURl}")
+                stash includes: '*.war', name: 'jenkinsWar'
+            }
+
+        }
+    })
+
+    ensureInNode(env, env.RUN_ATH_DOCKER_NODE ?: 'docker && highmem', {
+        if (skipExecution) {
+            return
```
burnMyDread: Is this a code style thing? this is really not something you normally do in groovy.

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+        }
+        stage("Running ATH") {
+            dir("athSources") {
+                unstash name: "athSources"
+                def uid = sh(script: "id -u", returnStdout: true)
+                def gid = sh(script: "id -g", returnStdout: true)
+                athContainerImage = docker.build('jenkins/ath', "--build-arg=uid='$uid' --build-arg=gid='$gid' -f src/main/resources/ath-container/Dockerfile .")
```
burnMyDread: Is this generating a docker container on the fly? If so, I am -1000 this requires Docker in Docker which we need to get away from.

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+            }
+
+            def testsToRun = metadata.tests?.join(",")
```
burnMyDread: In general I like this code as you are defaulting values in a clear way. However, you are still not consistent in your object approach.

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+            def browsers = metadata.browsers ?: ["firefox"]
+            def failFast = metadata.failFast ?: false
+            def rerunCount = metadata.rerunFailingTestsCount ?: 0
+            def localSnapshots = metadata.useLocalSnapshots ?: true
+
+            if (testsToRun == null && categoriesToRun == null) {
+                categoriesToRun = defaultCategory
+            }
+
+            def testingbranches = ["failFast": failFast]
```
burnMyDread: What is this and why is it not with the other object you just defined?

vars/runATH.groovy
```diff
@@ -0,0 +1,205 @@
...
+                    def currentBrowser = browser
+                    def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -e LOCAL_SNAPSHOTS=${localSnapshots} -e SHARED_DOCKER_SERVICE=true -e EXERCISEDPLUGINREPORTER=textfile -u ath-user"
+                    def commandBase = "./run.sh ${currentBrowser} ./jenkins.war -Dmaven.test.failure.ignore=true -DforkCount=1 -B -Dsurefire.rerunFailingTestsCount=${rerunCount}"
+                    if (infra.isRunningOnJenkinsInfra()) {
+                        def settingsXml = "${pwd tmp: true}/settings-azure.xml"
+                        writeFile file: settingsXml, text: libraryResource('settings-azure.xml')
+                        commandBase = commandBase + " -s ${settingsXml}"
+                    }
+
+                    if (testsToRun) {
```
burnMyDread: These two sections seem very similar. And you have flow control code interspersed with the code that runs the code that does runs the scenario. It would be much more clear if you pull out the code that runs the scenario into a method that would be more clear an easier to test. However, I am less worried about this as some of the other issues.

vars/runATH.groovy
raul-arabaolaza: No, is a parameter so you can have a typo, the metadata file used is printed just for debugging purposes so if the step wrongly results in a no-op you can see which metadata file is trying to find.

No strong opinion on my side if you consider this not needed

vars/runATH.groovy
raul-arabaolaza: From my PoV a metadata file should contain an ath section to be considered valid.
 ATH section itself is validated in next lines of code

vars/runATH.groovy
raul-arabaolaza: I only expect that to happen with "latest" versions, the other ones are not going to be used too much and concrete versions are downloaded via maven. For "latest" I am using CURL and following redirects (as latest is not deterministic to be downloaded via maven) , we may have problems if the number of executions of runATH is big enough at a given instant of time to overhelm the download server, but I have not found an easy alternative, as crawling the download to get the latest version and use maven could potentially have the same problem

vars/runATH.groovy
raul-arabaolaza: `commandBase` tells the ath to use a jenkins.war file at root level, this `unstash` is responsible to put the file there.

vars/runATH.groovy
raul-arabaolaza: I thought about that, but this way is trivial to add support to new browsers in the future, just add it to the `supportedBrowsers` array and unsupported ones are still ignored without further code changes

oleg-nenashev:
FYI @rysteboe

vars/runATH.groovy
rysteboe: +1 on this approach

vars/runATH.groovy
```diff
@@ -0,0 +1,217 @@
...
+
+            if (testsToRun == null && categoriesToRun == null) {
+                categoriesToRun = defaultCategory
+            }
+
+            def testingbranches = ["failFast": failFast]
+            for (browser in browsers) {
+                if (supportedBrowsers.contains(browser)) {
+
+                    def currentBrowser = browser
```
rysteboe: Doesn't matter much, but why is this brought out into a separate variable?  Looks like it's only used for reads and never reassigned.

vars/runATH.groovy
rysteboe: +1 on not replicating this logic

rysteboe:
Caveat my review with the fact that I have basically no groovy/pipeline experience.

vars/runATH.groovy
raul-arabaolaza: `params.get(key, default)` is the recommended way of default in pipeline since [script-security#153](https://github.com/jenkinsci/script-security-plugin/pull/153)

vars/runATH.groovy
raul-arabaolaza: (+1)

vars/runATH.groovy
raul-arabaolaza: To skip execution, basically converting the entire step in a no-op in case no metadata file is found, for doing that in pipeline I have to return

vars/runATH.groovy
raul-arabaolaza: Yeah, exactly like is done in [the existing Jenkinsfile](https://github.com/jenkinsci/acceptance-test-harness/blob/master/Jenkinsfile) as the container definition is part of the sources and we want to be sure we are using the appropriate container definition in all cases.

A follow up ticket will be created to publish and use a released version once is stable enough

vars/runATH.groovy
raul-arabaolaza: This is part of the `parallel` step, basically the way to set the fail-fast behavior, see [here](https://github.com/jenkinsci/workflow-cps-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep/help.html)

vars/runATH.groovy
raul-arabaolaza: Done

vars/runATH.groovy
raul-arabaolaza: As I am defining closures instead of running code the looping over `browser` mutated the contents of the closure, resulting in having all the closures with strings interpolated to the last value of `browser`

raul-arabaolaza:
Just for the record, this merge has been to the `runATH-step` branch so we can test on the real instance, this is not yet finished

PR #30, opened 2018-03-08
[INFRA-1489] First implementation

raul-arabaolaza:
[INFRA-1489](https://issues.jenkins-ci.org/browse/INFRA-1489)

See README and runATH.txt for use details, see [here](https://wiki.jenkins.io/pages/viewpage.action?pageId=138449195&src=contextnavchildmode) for the design

See [here](https://ci.jenkins.io/job/Plugins/job/pipeline-githubnotify-step-plugin/job/INFRA-1489/) for some successful runs on Jenkins CI

@reviewbybees @jenkins-infra/essentials @oleg-nenashev

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -72,7 +72,8 @@ Runs the link:https://github.com/jenkinsci/acceptance-test-harness[Acceptance Te
 
 The configuration is divided into two parts, one related to the step itself and another related to how the ATH is run.
 To configure the step just use the step's parameters described below, to configure the ATH runs a metadata file (in YAML format)
-is used. Further sections describe the metadata file in detail. *Note that if the metadata file does not exist this step will do nothing at all*
+is used. Further sections describe the metadata file in detail. *Note that if the metadata file does not exist this step
+will do nothing at all.*
 
 The list of step's params and the related default values are:
 
@@ -182,3 +183,5 @@ In case you want to use the defaults for all properties you can use
 ----
 ath: "default"
 ----
+
+*Please note that a blank metadata file will result in an error*
--- a/README.adoc
+++ b/README.adoc
@@ -65,3 +65,120 @@ if (infra.isTrusted()) {
     /* perform some trusted action like a deployment */
 }
 ----
+
+=== runATH
+
+Runs the link:https://github.com/jenkinsci/acceptance-test-harness[Acceptance Test Harness] in a configurable way
+
+The configuration is divided into two parts, one related to the step itself and another related to how the ATH is run.
+To configure the step just use the step's parameters described below, to configure the ATH runs a metadata file (in YAML format)
+is used. Further sections describe the metadata file in detail. *Note that if the metadata file does not exist this step will do nothing at all*
+
+The list of step's params and the related default values are:
+
+
+* `athUrl`: The URL to get the ATH sources. It can point to a local path (by using the file:// protocol) or a github destination. Defaults to https://github.com/jenkinsci/acceptance-test-harness.git. *Can be overridden from the metadata file*
+* `athRevision`: The ATH revision to use, can be a branch or tag name or a commit id. Defaults to branch master. *Can be overridden from the metadata file*
+* `metadataFile`: A String indicating the file path (relative to the where this step is executed) to use as metadata file for the build, more details about the metadata file are provided belows. *Defaults to `essentials.yml` at the location where this step is invoked*
+* `jenkins`: URI to the jenkins.war, Jenkins version or one of "latest", "latest-rc", "lts" and "lts-rc". Defaults to "latest". For local war files use the file:// protocol in the URI. *Can be overriden from the metadata file*
+
+.Jenkinsfile
+[source,groovy]
+----
+runATH(metadataFile:"metadata.yml", athRevision: "master", athUrl:"https://github.com/jenkinsci/acceptance-test-harness.git", jenkins: "2.110")
+----
+
+
+To make it usable in PR builders this step allows users to run the ATH using custom (typically previously built in the same Jenkinsfile) versions of any plugin, for that you need to set the metadata file's `useLocalSnapshots` property to true and stash the
+plugins you want to use in the ATH run. By default you need to stash them with the name `localPlugins` the step will unstash them when appropriate and use the ATH`s `LOCAL_JARS` property to run the ATH. You can stash any number of plugins, all of
+them will be used. You can also stash under other name by setting the env variable `RUN_ATH_LOCAL_PLUGINS_STASH_NAME`
+
+.Jenkinsfile
+[source,groovy]
+----
+node("linux") {
+        dir("sources") {
+          checkout scm
+          List<String> mavenEnv = [
+                    "JAVA_HOME=${tool 'jdk8'}",
+                    'PATH+JAVA=${JAVA_HOME}/bin',
+                    "PATH+MAVEN=${tool 'mvn'}/bin"]
+          withEnv(mavenEnv) {
+            sh "mvn clean install -DskipTests"
+          }
+          dir("target") {
+           stash includes: '*.hpi', name: 'localPlugins'
+          }
+
+          runATH(metadataFile:"metadata.yml", athRevision: "master")
+        }
+    }
+----
+
+.Jenkinsfile
+[source,groovy]
+----
+    node("linux") {
+        dir("sources") {
+          checkout scm
+          List<String> mavenEnv = [
+                    "JAVA_HOME=${tool 'jdk8'}",
+                    'PATH+JAVA=${JAVA_HOME}/bin',
+                    "PATH+MAVEN=${tool 'mvn'}/bin"]
+          withEnv(mavenEnv) {
+            sh "mvn clean install -DskipTests"
+          }
+          dir("target") {
+           stash includes: '*.hpi', name: 'snapshots'
+          }
+
+          env.RUN_ATH_LOCAL_PLUGINS_STASH_NAME="snapshots"
+          runATH(metadataFile:"metadata.yml", athRevision: "master")
+        }
+    }
+----
+
+The metadata file is a YAML file with the following structure:
+
+.metadata
+[source,yaml]
+----
+ath:
+  athUrl: https://github.com/jenkinsci/acceptance-test-harness.git
+  athRevision: acceptance-test-harness-1.59
+  jenkins: 2.89.4
+  failFast: false
+  rerunFailingTestsCount: 0
+  useLocalSnapshots: true
+  browsers:
+    - firefox
+    - chrome
+  tests:
+    - Test1
+    - Test2
+    - Test3
+  categories:
+    - Category1
+    - Category2
+----
+
+Where:
+
+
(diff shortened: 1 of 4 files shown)


README.adoc
```diff
@@ -65,3 +65,120 @@ if (infra.isTrusted()) {
...
+    * `jenkins`: (Optional) URI to the jenkins.war file, Jenkins version or one of "latest", "latest-rc", "lts" and "lts-rc". If specified it will override the parameter in the runATH step
+    * `failFast`: If the run has to fail fast or not. Defaults to false if not specified
+    * `rerunFailingTestsCount`: The number of runs per failing test (a la maven). Defaults to zero</li>
+    * `useLocalSnapshots`: If the ATH should use local versions of the plugins. Defaults to true. *Note that if true the runATH expects the local plugins to be stashed, setting this to true without the stash will make the step fail*
+    * `browsers`: The list of browsers to use when running ATH Defaults to firefox. *Note that currently only firefox browser is supported, any other will be ignored*
+    * `tests`: The list of tests to run for the component that calls the step. If no particular set of tests or categories is defined the SmokeTest Category of the ATH will be run
+    * `categories`: The list of Categories to run. Defaults to nothing
+
+In case you want to use the defaults for all properties you can use
```
rysteboe: I would also note that an error will occur if the metadata file doesn't contain an `ath` section at all - a blank metadata file != `ath: "default"`

README.adoc
```diff
@@ -65,3 +65,123 @@ if (infra.isTrusted()) {
     /* perform some trusted action like a deployment */
 }
 ----
+
+=== runATH
+
+Runs the link:https://github.com/jenkinsci/acceptance-test-harness[Acceptance Test Harness] in a configurable way
```
batmat: NIT: missing period?

vars/runATH.groovy
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env groovy
```
batmat: This shebang is conceptually wrong IMO, and from a user standpoint we should probably not encourage people to think/remember this is Groovy. (this is not really, because of the subtle differences with groovy-cps)

vars/runATH.groovy
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * Wrapper for running the ATH see README and runATH.txt for full documentation
+ */
+
+def call(Map params = [:]) {
+    def athUrl = params.get('athUrl', 'https://github.com/jenkinsci/acceptance-test-harness.git')
+    def athRevision = params.get('athRevision', 'master')
+    def metadataFile = params.get('metadataFile', 'essentials.yml')
+    def jenkins = params.get('jenkins', 'latest')
```
batmat: Possibly rename this parameter to include some version in it? jenkins alone seems generic enough :)

vars/runATH.groovy
```diff
@@ -0,0 +1,221 @@
...
+
+        if (!fileExists(metadataFile)) {
+            echo "Skipping ATH execution because the metadata file does not exist. Current value is ${metadataFile}."
+            skipExecution = true
+            return
+        }
+
+        stage("Getting ATH sources and Jenkins war") {
+            // Start validation
+            metadata = readYaml(file: metadataFile).ath
```
batmat: Didn't you mean using `readYaml(file: metadataFile)?.ath` instead? I think it won't even reach the null check next line if the file is absent, but maybe it's not supposed to happen indeed?

vars/runATH.groovy
```diff
@@ -0,0 +1,221 @@
...
+            }
+            if (jenkins == "latest") {
+                jenkinsURl = mirror + "war/latest/jenkins.war"
+            } else if (jenkins == "latest-rc") {
+                jenkinsURl = mirror + "/war-rc/latest/jenkins.war"
+            } else if (jenkins == "lts") {
+                jenkinsURl = mirror + "war-stable/latest/jenkins.war"
+            } else if (jenkins == "lts-rc") {
+                jenkinsURl = mirror + "war-stable-rc/latest/jenkins.war"
+            }
```
batmat: should there be an else { error 'wat' } to avoid surprises?

vars/runATH.groovy
```diff
@@ -0,0 +1,221 @@
...
+                athSourcesFolder = athUrl - "file://"
+            } else {
+                dir(athSourcesFolder) {
+                    checkout changelog: true, poll: false, scm: [$class           : 'GitSCM', branches: [[name: athRevision]],
+                                                                 userRemoteConfigs: [[url: athUrl]]]
+                }
+            }
+            dir(athSourcesFolder) {
+                // We may need to run things in parallel later, so avoid several checkouts
+                stash name: "athSources"
```
batmat: Not sure this is not early optimization. Why not just `checkout scm` wherever we need sources?

vars/runATH.groovy
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+private String getLocalPluginsList() {
+    dir("localPlugins") {
+       return sh(script : "ls -p -d -1 ${pwd()}/*.* | tr '\n' ':'| sed 's/.\$//'", returnStdout: true).trim()
+    }
+}
+
+private void ensureInNode(env, nodeLabel, body) {
```
batmat: What is this trying to achieve? I think a comment would be great! :)

vars/runATH.txt
```diff
@@ -0,0 +1,128 @@
...
+  categories:
+    - Category1
+    - Category2
+</pre>
+
+<p>
+    Where:
+</p>
+
+<ul>
```
batmat: This whole duplicated block looks like a maintenance burden? Can't we just link to this from the README and avoid duplication?

README.adoc
```diff
@@ -65,3 +65,123 @@ if (infra.isTrusted()) {
...
+    - Test3
+  categories:
+    - Category1
+    - Category2
+----
+
+Where:
+
+
+    * `athUrl`: (Optional) The URL to get the ATH sources. It can point to a local path or a github destination. If specified it will override the parameter in the runATH step
```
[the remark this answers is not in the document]
raul-arabaolaza: Yeah, I agree with you :)

README.adoc
```diff
@@ -65,3 +65,123 @@ if (infra.isTrusted()) {
...
+    * `athRevision`: (Optional) The ATH revision to use can be a branch or tag name or a commit id. If specified it will override the parameter in the runATH step
+    * `jenkins`: (Optional) URI to the jenkins.war file, Jenkins version or one of "latest", "latest-rc", "lts" and "lts-rc". If specified it will override the parameter in the runATH step
```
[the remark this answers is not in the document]
raul-arabaolaza: Release candidate, I am just using the same versions that the local scripts in the ATH allow for coherence between local and CI runs

vars/runATH.groovy
raul-arabaolaza: Yeah, a blank file must fail, but better have a good error message than an NPE. If the file is absent the intention is to work as a no-op and is managed before this lines

vars/runATH.groovy
raul-arabaolaza: If none of this if's match I am assuming jenkinsUrl is a URI, not failing in that case because I want to support URI's like links to files generated by other jobs, there is a validation to check if the file is reachable (valid URL) on line 80

vars/runATH.groovy
raul-arabaolaza: Because potentially I can run in two nodes (one for categories and another for single tests) and in the future we may need to further parallelize, so I prefer to checkout in the linux node and stash and then unstash in all the parallel branches I may need. I also need sources to build the docker image and I want to do that only once

vars/runATH.txt
raul-arabaolaza: As a user, I greatly appreciate when the "Pipeline syntax"  section offers me all the needed doc instead of a link to an external page, yeah I agree is a maintenance burden but I am thinking in users here

vars/runATH.groovy
raul-arabaolaza: `buildPlugin` includes a shebang and is also cited in the  [doc|https://jenkins.io/doc/book/pipeline/getting-started/#defining-a-pipeline-in-scm] as a way to make sure you IDE understands the syntax.

So, for the moment I would prefer not to remove it, I understand your point but for me IDE support is a must in pipeline development given all the pending improvements.

vars/runATH.groovy
raul-arabaolaza: Keep in mind that you can pass a war file, so if I include `version` it will be misleading, also keeping the same format than the local scripts here.

I am open to discuss that, but that should be done in a separate PR which should also change the ATH local scripts, not here

vars/runATH.groovy
```diff
@@ -0,0 +1,221 @@
...
+
+def call(Map params = [:]) {
+    def athUrl = params.get('athUrl', 'https://github.com/jenkinsci/acceptance-test-harness.git')
+    def athRevision = params.get('athRevision', 'master')
+    def metadataFile = params.get('metadataFile', 'essentials.yml')
+    def jenkins = params.get('jenkins', 'latest')
+
+    def mirror = "http://mirrors.jenkins.io/"
+    def defaultCategory = "org.jenkinsci.test.acceptance.junit.SmokeTest"
+    def jenkinsURl = jenkins
```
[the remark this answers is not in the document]
raul-arabaolaza: I am going to transform this variable in a proper URL later if appropriate, this is just to make sure that if `jenkins` is a proper URL then is used as is as you can see in line 80

vars/runATH.txt
batmat: We should aim at both. Or at least find some way for instance to convert one source to some other format in an automated fashion or so. Copy-pasting always ends up badly.

vars/runATH.groovy
batmat: Well not sure I buy that: `checkout scm` where need is simpler than `stash` and `unstash` at the right places. So again looks like a premature optimization.

I would tend to agree if this was to not rebuild artifacts. Sources should never matter. I suspect `stash`ing might even be slower than just `git clone`ing again. 
For what it's worth, just checked, a fresh `git clone` took less than a second on my machine.

Anyway.

raul-arabaolaza:
@batmat FYI I have created a follow up ticket to track your concerns [INFRA-1533](https://issues.jenkins-ci.org/browse/INFRA-1533)

raul-arabaolaza:
@reviewbybees done

cc @rtyler This is ready to merge I believe

raul-arabaolaza:
@olivergondza Nothing production ready at this moment, as the intention of this step is to be run as part of PR and run only relevant tests. Of course, you can generate several different metadata files and call several times `runATH` in parallel but that is a hack IMO.

If you consider it necessary I guess it would be simple to create a `runFullATH` based on this one that gets the ATH, lists all tests there and generates some splits

lemeurherve:
> See README and runATH.txt for use details, see [here](https://wiki.jenkins.io/pages/viewpage.action?pageId=138449195&src=contextnavchildmode) for the design

Current wiki.jenkins.io link: https://wiki.jenkins-ci.org/JENKINS/runATH-and-runPCT-step-design.html

PR #31, opened 2018-03-09
Pass -Daccess-modifier-checker.failOnError=false when using a custom baseline

jglick:
Justification given in https://github.com/kohsuke/access-modifier/pull/10. As usual, this is untested.

@reviewbybees esp. @dwnusbaum


jglick:
> Do you think it would be useful to do the same with Findbugs

Probably yes; I have hit the situation you describe many times. Feel free to propose it.

jglick:
(And yes there are also cases where a newly introduced annotation alerts you to a problem that was there all along. But I think the top priority for someone adding a custom baseline to their build is to see if _tests_ detect any real integration problem.)

PR #35, opened 2018-04-03
INFRA-1534: Create a runPCT step

raul-arabaolaza:
[INFRA-1534](https://issues.jenkins-ci.org/browse/INFRA-1534)

Ready for review

@jenkins-infra/essentials

PR #36, opened 2018-04-03
Introduce infra.runWithMaven, infra.runWithJava and infra.checkout methods

oleg-nenashev:
It is a split-out of some refactoring code from #21, which we discussed with @raul-arabaolaza today.It should help to simplify the new scripts.

Documentation has not been created yet, I will do it once feasibility is confirmed.

@reviewbybees

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -34,25 +34,11 @@ def call(Map params = [:]) {
                         boolean isMaven
 
                         stage("Checkout (${stageIdentifier})") {
-                            if (env.BRANCH_NAME) {
-                                checkout scm
-                            }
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
Tested all methods on my instance. Now they work well with both Maven and Gradle flows

raul-arabaolaza:
@oleg-nenashev As this PR and #35 are somehow running in parallel I am going to suggest to just go ahead with them and create a follow up task once both are merged to perform further consolidation.

I mean, #35 is fully functional and this is code consolidation, I would like to not wait for more time than necessary waiting for downstream changes

oleg-nenashev:
I am not sure who can merge merge that. 
I pinged @rtyler yesterday about it, because I am also blocked with other PRs.

PR #38, opened 2018-04-06
Override tool names

olivergondza:
Make java and maven names overridable through environment variables.


olivergondza:
Ugh, rewritten on master completely.

PR #39, opened 2018-04-06
[FIXED JENKINS-50585] Run full ATH by default

olivergondza:
Run full ATH suite when no tests/categories specified. To be merged after https://github.com/jenkinsci/jenkins/pull/3381.

### DIFF
--- a/vars/runATH.groovy
+++ b/vars/runATH.groovy
@@ -11,7 +11,6 @@ def call(Map params = [:]) {
     def jenkins = params.get('jenkins', 'latest')
 
     def mirror = "http://mirrors.jenkins.io/"
-    def defaultCategory = "org.jenkinsci.test.acceptance.junit.SmokeTest"
     def jenkinsURL = jenkins
     def metadata
     def athContainerImage
@@ -131,10 +130,6 @@ def call(Map params = [:]) {
             // Elvis fails in case useLocalSnapshots == false in metadata File
             def localSnapshots = metadata.useLocalSnapshots != null ? metadata.useLocalSnapshots : true
 
-            if (testsToRun == null && categoriesToRun == null) {
-                categoriesToRun = defaultCategory
-            }
-
             def testingbranches = ["failFast": failFast]
             for (browser in browsers) {
                 if (supportedBrowsers.contains(browser)) {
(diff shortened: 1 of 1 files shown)


raul-arabaolaza:
@olivergondza I agree than the Smokes may be not make sense for most plugins but my concern about this is that anyone (for example a plugin maintainer) that tries to use this step with default values will get a failed build as the ATH as a whole is not green and it will take too much time to be run. Having a default that makes your build fail and is going to block the executor for more than 10 hours, is not a good idea IMO.

That said if there is a compeling reason for this and other people agrees with this change I am not going to block this as this is only a default and you can always specify the relevant tests via the metadata file

raul-arabaolaza:
@olivergondza Please see my comment in JENKINS-50585, that may be enough for your needs without the need of more code changes

olivergondza:
@raul-arabaolaza, I was not aware of the `-Dtest=*` trick, thanks! I see it triggers slightly different maven execution as there are profiles activated when the property is present presuming tests are being cherry-picked. Not that it would cause any real issue I could identify for now.

I do not have a strong objections against smoke tests being the default, provided there is a straightforward way to switch to full run. In any event, I would like `runAth` to handle full run explicitly so it can parallelize the tests (10 hours is a bit long) and get rid of passing in the `-Dtest` property if possible.

raul-arabaolaza:
@olivergondza  `I would like runAth to handle full run explicitly so it can parallelize the tests (10 hours is a bit long) and get rid of passing in the -Dtest property if possible`
Yeah, makes perfect sense, maybe something in the lines of 
```
ath:
  jenkins: 'latest'
  useLocalSnapshots: false
  tests: "all"
```
And make the `runATH` able to understand that and run in parallel only in that case? I would like to not run in parallel by default for the moment. For single runs (one plugin PR for example) tests should not take too much time unless you are abusing selenium and not creating enough unit and integration tests. WDYT?

olivergondza:
Fine with me. Should we dismiss https://github.com/jenkinsci/jenkins/pull/3381 if we go that way in your opinion?

raul-arabaolaza:
I liked that change, is a small one and makes it much more understandable for people which is not familiar with `runATH` step. So I would prefer to keep it

olivergondza:
@raul-arabaolaza, changed accordingly, please re-review.

olivergondza:
Well, this is a value with special handling. So we can either wrap the whole thing with yet another nested controls - which is what I tried to avoid - or used what I used. One more hting to do so would be to move the `all` handling before tests and categories with `continue` (not sure if works in pipeline).

I do not feel strongly.

raul-arabaolaza:
@olivergondza You are right, starting to use nested controls may also have disadvantages, ok. I do not feel so strongly about it to block you more time. 

And thanks for this :)

olivergondza:
Cool, thanks, both of you.

@rtyler, could you merge this?

olivergondza:
Merge conflict resolved

olivergondza:
@rtyler, Can we get this in please? This is a second merge conflict I've got to resolve to keep this mergeable...

oleg-nenashev:
I will sync-up with @raul-arabaolaza tomorrow regarding it

raul-arabaolaza:
@oleg-nenashev I am ok with merging

oleg-nenashev:
YOLO then, 🚢 🇮🇹

olivergondza:
Thanks guys!

PR #41, opened 2018-04-06
Remove maven download logs from output

carlossg:
They make output extra verbose for containerized agents


oleg-nenashev:
CC @raul-arabaolaza

PR #42, opened 2018-04-10
Add infra.retrieveMavenSettingsFile() + documentation

oleg-nenashev:
It is a follow-up to my comments in #35 (CC @raul-arabaolaza). In order to properly pass Maven Settings steps across all steps, we may need to add too much parallelization. In order to simplify implementing a default behavior, I have created a new library step so that `runPCT()` can use the same logic to retrieve settings as common Maven runners.

Also added docs for previously created utility methods as @rtyler asked

@reviewbybees

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -30,20 +30,48 @@ Object checkout(String repo = null) {
 }
 
 /**
+ * Retrieves Settings file to be used with Maven.
+ * If {@code MAVEN_SETTINGS_FILE_ID} variable is defined, the file will be retrieved from the specified
+ * configuration ID provided by Config File Provider Plugin.
+ * Otherwise it will fallback to a standard Jenkins infra resolution logic.
+ * @param settingsXml Absolute path to the destination settings file
+ * @return {@code true} if the file has been defined
+ */
+boolean retrieveMavenSettingsFile(String settingsXml) {
+    if (env.MAVEN_SETTINGS_FILE_ID != null) {
+        configFileProvider([configFile(fileId: env.MAVEN_SETTINGS_FILE_ID, variable: 'mvnSettingsFile')]) {
+            if (isUnix()) {
+                sh "cp ${mvnSettingsFile} ${settingsXml}"
+            } else {
+                bat "copy ${mvnSettingsFile} ${settingsXml}"
+            }
+            return true
+        }
+    } else if (jdk.toInteger() > 7 && isRunningOnJenkinsInfra()) {
+        /* Azure mirror only works for sufficiently new versions of the JDK due to Letsencrypt cert */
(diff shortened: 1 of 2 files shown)


vars/infra.groovy
```diff
@@ -29,29 +29,57 @@ Object checkout(String repo = null) {
...
+    if (env.MAVEN_SETTINGS_FILE_ID != null) {
+        configFileProvider([configFile(fileId: env.MAVEN_SETTINGS_FILE_ID, variable: 'mvnSettingsFile')]) {
+            if (isUnix()) {
+                sh "cp ${mvnSettingsFile} ${settingsXml}"
+            } else {
+                bat "copy ${mvnSettingsFile} ${settingsXml}"
+            }
+            return true
+        }
+    } else if (jdk.toInteger() > 7 && isRunningOnJenkinsInfra()) {
```
raul-arabaolaza: Is failing me with jdk is not defined

oleg-nenashev:
@raul-arabaolaza thanks for catching, should be fixed now. I will test it properly on my instance on Thursday

raul-arabaolaza:
@oleg-nenashev Any news on this? Is it ready for merge?

oleg-nenashev:
@raul-arabaolaza I think so

raul-arabaolaza:
Thanks @oleg-nenashev,  pinging @rtyler then

vars/infra.groovy
```diff
@@ -29,29 +29,58 @@ Object checkout(String repo = null) {
...
+/**
+ * Retrieves Settings file to be used with Maven.
+ * If {@code MAVEN_SETTINGS_FILE_ID} variable is defined, the file will be retrieved from the specified
+ * configuration ID provided by Config File Provider Plugin.
+ * Otherwise it will fallback to a standard Jenkins infra resolution logic.
+ * @param settingsXml Absolute path to the destination settings file
+ * @param jdk Version of JDK to be used
+ * @return {@code true} if the file has been defined
+ */
+boolean retrieveMavenSettingsFile(String settingsXml, String jdk = 8) {
```
rtyler: Is `jdk` supposed to be an integer or a String? It's weird to me that it's typed as a String when only numeric operations are done

vars/infra.groovy
rtyler: is this still a problem for you @raul-arabaolaza

vars/infra.groovy
oleg-nenashev: Maybe not-as-String at some point. e.g. 8u161 or 10.021-rc

vars/infra.groovy
raul-arabaolaza: @rtyler No, working properly for me now (via replay) in my testing environment

PR #44, opened 2018-04-20
[INFRA-1587] Fix return value when using custom maven settings

raul-arabaolaza:
[INFRA-1587](https://issues.jenkins-ci.org/browse/INFRA-1587)

The returning was exiting the block of configFileProvider but not the method itself, so in the end if you are using a custom maven settings the method was returning false

cc @oleg-nenashev

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -46,8 +46,8 @@ boolean retrieveMavenSettingsFile(String settingsXml, String jdk = 8) {
             } else {
                 bat "copy ${mvnSettingsFile} ${settingsXml}"
             }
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
I have tested it locally and...

* I confirm the original version breaks on newest plugin versions (my test env was 2.73.x with outdated Pipeline)
* I confirm that the fix works well on new and old versions 

:shipit:

oleg-nenashev:
@reviewbybees done

raul-arabaolaza:
@rtyler Ready to merge

PR #53, opened 2018-05-03
[INFRA-1609] - Fix runPCT() when multiple Java options are passed in args

oleg-nenashev:
https://issues.jenkins-ci.org/browse/INFRA-1609

@reviewbybees @raul-arabaolaza

### DIFF
--- a/vars/runPCT.groovy
+++ b/vars/runPCT.groovy
@@ -127,7 +127,7 @@ def call(Map params = [:]) {
                         }
                         def command = "JENKINS_WAR_PATH=${warAbsolutePath} run-pct ${pctBranchOptions.join(' ')}"
                         if (!javaOptions.isEmpty()) {
-                            command = "JAVA_OPTS=\"${javaOptions.join(' ')}\" ${command}"
+                            command = """JAVA_OPTS="${javaOptions.join(' ')}" ${command}"""
                         }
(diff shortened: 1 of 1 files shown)


vars/runPCT.groovy
```diff
@@ -127,7 +127,7 @@ def call(Map params = [:]) {
                         }
                         def command = "JENKINS_WAR_PATH=${warAbsolutePath} run-pct ${pctBranchOptions.join(' ')}"
                         if (!javaOptions.isEmpty()) {
-                            command = "JAVA_OPTS=${javaOptions.join(' ')} ${command}"
+                            command = "JAVA_OPTS=\"${javaOptions.join(' ')}\" ${command}"
```
batmat: NIT: generally I find that using the nice triple single/double quotes makes code more readable avoiding to use `\`.

Here this would be:

```groovy
command = '''JAVA_OPTS=${javaOptions.join(' ')} ${command} '''
```

batmat: Oh wait, we need the expression to be interpreted, so we want triple double `"""` instead

```groovy
command = """JAVA_OPTS="${javaOptions.join(' ')}" ${command} """
```

raul-arabaolaza: Yeah, I am with @batmat here, it makes it much more readable than escaping characters

PR #58, opened 2018-05-08
Added azure-incrementals mirror

jglick:
Not sure how to test this in context, though

```
$ curl -I https://repo.azure.jenkins.io/incrementals/org/jenkins-ci/main/jenkins-war/2.121-rc15727.3b61b96119b9/jenkins-war-2.121-rc15727.3b61b96119b9.war
HTTP/2 200 
…
content-length: 74618209
…
x-cache-status: MISS
…
```

works as expected. CC @raul-arabaolaza.

PR #60, opened 2018-05-09
Incrementals support

raul-arabaolaza:
Add the ability to get incrementals versions of jenkins to `infra.stashJenkinsWar` by making sure all proper repos are defined when running in ci.jenkins.io

Follow up of #58 and #59 

cc @oleg-nenashev @jglick


resources/settings-azure.xml
```diff
@@ -1,4 +1,39 @@
...
+    <profiles>
+        <profile>
+            <id>jenkins-repos</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <repositories>
+                <repository>
+                    <id>incrementals</id>
+                    <url>https://repo.jenkins-ci.org/incrementals/</url>
```
jglick: This gets overridden by the `mirror` IIUC.

PR #65, opened 2018-05-23
[JENKINS-51489] Use ATH pre configured plugins mode from essentialsTest

raul-arabaolaza:
[JENKINS-51489](https://issues.jenkins-ci.org/browse/JENKINS-51489)

*Downstream of:*
* #64
* [ATH#433](https://github.com/jenkinsci/acceptance-test-harness/pull/433) 

Modify `essentialsTest` so it uses the new pre configured plugins mode introduced in [ATH#433](https://github.com/jenkinsci/acceptance-test-harness/pull/433)

@reviewbybees @oleg-nenashev @jglick

### DIFF
--- a/vars/essentialsTest.groovy
+++ b/vars/essentialsTest.groovy
@@ -26,9 +26,7 @@ def call(Map params = [:]) {
             if (configData.ath != null && !configData.ath.disabled) {
                 stage("Run ATH") {
                     dir("ath") {
-                        def configFile = "ath-config.groovy"
-                        writeFile file: configFile, text: "pluginEvaluationOutcome='${testPluginResolution}'"
-                        runATH jenkins: customWarURI, metadataFile: metadataPath, configFile: configFile
+                        runATH jenkins: customWarURI, metadataFile: metadataPath, javaOptions: ["-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn", testPluginResolution]
                     }
                 }
             }
--- a/vars/essentialsTest.groovy
+++ b/vars/essentialsTest.groovy
@@ -2,7 +2,7 @@ def call(Map params = [:]) {
     def baseDir = params.containsKey('baseDir') ? params.baseDir : "."
     def metadataFile = params.containsKey('metadataFile') ? params.metadataFile : "essentials.yml"
     def labels = params.containsKey('labels') ? params.labels : "docker && highmem"
-    def pluginEvaluationOutcome = params.get("pluginEvaluationOutcome") ?: "failOnInvalid"
+    def testPluginResolution
 
     node(labels) {
         stage("Checkout") {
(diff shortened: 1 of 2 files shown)


vars/essentialsTest.groovy
```diff
@@ -2,6 +2,7 @@ def call(Map params = [:]) {
     def baseDir = params.containsKey('baseDir') ? params.baseDir : "."
     def metadataFile = params.containsKey('metadataFile') ? params.metadataFile : "essentials.yml"
     def labels = params.containsKey('labels') ? params.labels : "docker && highmem"
+    def pluginEvaluationOutcome = params.get("pluginEvaluationOutcome") ?: "failOnInvalid"
```
oleg-nenashev: The argument name cannot be understood bu a user without reading docs.
I would rather prefe...

```yaml
metadata:
   testPluginResolution:
     skipOnUnmetDependencies: true
```

with defaulting to `failOnInvalid`

raul-arabaolaza: Absolutely, that makes it much more nicer, will update the PR

vars/essentialsTest.groovy
```diff
@@ -10,7 +11,9 @@ def call(Map params = [:]) {
 
         dir(baseDir) {
             def metadataPath = "${pwd()}/${metadataFile}"
-            metadata = readYaml(file: metadataPath)
+            def configData = readYaml(file: metadataPath)
```
raul-arabaolaza: The variable was lacking a `def` and I have changed the name of it to avoid having a `metadata.metadata` in the code which may be misleading, and I like the idea of a `metadata`section in the yml file

vars/essentialsTest.groovy
```diff
@@ -10,7 +11,9 @@ def call(Map params = [:]) {
...
+            testPluginResolution = configData.metadata?.testPluginResolution?.skipOnUnmetDependencies ?  "skipOnInvalid" : "failOnInvalid"
```
raul-arabaolaza: Basically this is allways `failOnInvalid` unless you have:
```
metadata:
  testPluginResolution:
    skipOnUnmetDependencies: true
```

raul-arabaolaza:
Some reasonable changes requested in the upstream PR, I am going to modify this a little bit to accommodate to the new code

oleg-nenashev:
Regarding `metadata` entry, it was rather a mistake from my side. I was about proposing 

```ysml
ath:
  testPluginResolution:
    skipOnUnmetDependencies: true
```

Having a separate section sounds reasonable anyway, because the same setting may be required for `runATH()` at some point. I just doubt that `metadata` is a best name since it's still a source of confusion with the "metadata file" term. 

Not sure what would be a better name tho

raul-arabaolaza:
@oleg-nenashev Yeah, I am all for a separate section because this is not intended to work on ath step alone, this setting makes sense in a concrete flow (essentials in this case)  and the `runATH` already provides support for this kind customization via the custom `javaOptions` parameter.

I am thinking on a `flow` section or something on that lines that can be used by shared libraries or Jenkinsfiles to get some metadata in order to implement a high level flow. `essentialsTest` is one of those. In that section you can add some semantic metadata like the `testPluginResolution` or something more generic like:

```
flow: 
  javaOptions:
    pct:
      - "-D..."
    ath: 
     - "-Djfjefjf"
  ath:
    testPluginResolution:
      skipOnUnmetDependencies: true
```

raul-arabaolaza:
@oleg-nenashev I had to fix some conflicts, all needed PR are already merged, can you re-bee this if you feel is worthwile?

PR #68, opened 2018-06-19
[INFRA-1672] - Add option to skip tests in buildPlugin()

oleg-nenashev:
In order to deliver https://github.com/jenkinsci/workflow-support-plugin/pull/65 to Incrementals, we need to disable test publishing for a while.

Sample command: 
```
buildPlugin(platforms: ['linux'], tests: [skip: true])
```

Unfortunately now buildPlugin() just fails when there is no tests. It's a totally valid behavior by default, but I propose to introduce an option which allows disabling tests.

https://issues.jenkins-ci.org/browse/INFRA-1672

@raul-arabaolaza

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -30,6 +30,9 @@ buildPlugin()
   execute the steps against in parallel
 * `jenkinsVersions`: (default: `[null]`) - a matrix of Jenkins baseline versions to build/test against in parallel (null means default,
   only available for Maven projects)
+* `tests`: (default: `null`) - a map of parameters to run tests during the build
+** `skip` - If `true`, skipp all the tests by setting the `-skipTests` profile.
+  It will also skip FindBugs in modern Plugin POMs.
 * `findbugs`: (default: `null`) - a map of parameters to run findbugs and/or archive findbugs reports. (only available for Maven projects)
 ** `run`: set to `true` to add the `findbugs:findbugs` goal to the maven command.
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
I have checked the behavior on both enabled and disabled tests. Works well, no regressions

oleg-nenashev:
Thanks @raul-arabaolaza !

rtyler:
I'm **strongly** against this change being left in, and would like to see it reverted.

There's not an acceptable justification to allow broken code to the incrementals repository, and without tests, there is no way one could prove to me that the code is not broken :smile_cat: 

Disabling of tests also is against the _whole point_ of allowing Plugins to use ci.jenkins.io, which was to encourage better quality practices.

oleg-nenashev:
@rtyler Our use-case:
* We need to get release on Incrementals so that Java 10 hackers can start trying Pipeline
* Our tooling is not there yet to build/test with Java 10 patches, work in progress
* We know that the tests are broken, but we accept the risk

So, we need it at least as a temporary solution. I am happy to remove documentation and to add TODO comment for removal so the option does not get adopted elsewhere. WDYT?

rtyler:
@oleg-nenashev once this week is over, please remove it entirely.

oleg-nenashev:
Let's where we stand next week and discuss it. If we manage to fix tooling,
I am totally fine with removing

PR #74, opened 2018-09-20
Make the plugin build timeout configurable

batmat:
Plan to use this for configuration-as-code build while working on [JENKINS-53687](https://issues.jenkins-ci.org/browse/JENKINS-53687).
Not aiming at leaving this there forever, but I think that some plugins with a lot of tests could need a bit more than one hour.

Or some could even do the contrary to get a faster feedback: lower the default timeout value to have the build fail faster in case something is wrong.

cc @ndeloof @ewelinawilkosz


batmat:
@rtyler fixed, care to re-review? Thanks

PR #76, opened 2018-10-23
Permit buildPlugin jdk/jenkins/label matrix to be sparse

olivergondza:
This change provides an alternative way for specifying the environment for the build to run permitting to exclude combinations known not to work or not to make sense.

Thanks to @abayer for pairing while working on this.

Mostly whitespace change: https://github.com/jenkins-infra/pipeline-library/pull/76/files?w=1

PR #79, opened 2018-12-02
buildPlugin(): Allow passing Java level as a configuration argument

oleg-nenashev:
While testing configuration support patches from @olivergondza in https://github.com/jenkins-infra/pipeline-library/pull/76 , I tried to run the test automation against a plugin which still targets Java 7: https://github.com/jenkinsci/label-verifier-plugin/pull/4 . It failed, because the new Jenkins core version uses Bytecode Compatibility Transformer which is no longer compatible with Java 7.

Other static checkers may fail in various way as well. In order to prevent it, I propose to add an escape hatch to the library so that a custom `java.level` can be passed alongside a custom Jenkins version.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -39,7 +39,7 @@ buildPlugin()
 buildPlugin(/*...*/, configurations: [
   [ platform: "linux", jdk: "8", jenkins: null ],
   [ platform: "windows", jdk: "8", jenkins: null ],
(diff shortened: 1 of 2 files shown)


batmat:
Just added https://github.com/jenkins-infra/pipeline-library/pull/80 in complement FYI.

PR #80, opened 2018-12-05
--show-version aka -V should be default

batmat:
It will show Maven and Java versions all the time, which is IMO much more than a nice to have in a CI env :).

Will be especially useful when we start getting more Java 11 around.

cc @oleg-nenashev @mramonleon @alecharp

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -68,6 +68,7 @@ boolean retrieveMavenSettingsFile(String settingsXml, String jdk = 8) {
 Object runMaven(List<String> options, String jdk = 8, List<String> extraEnv = null, String settingsFile = null) {
     List<String> mvnOptions = [
         '--batch-mode',
+        '--show-version',
         '--errors',
         '-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn',
     ]


batmat:
> I don't have anything against this addition, but I don't think we should rely excessively on the content of the logs for the version of the JDK used: it's drown in too much details.

Not sure what you exactly mean. It's one of the only ways to be sure of the version of Java and which exact Maven version is in use. What there be an alternative? 
Also, yes, it's in the logs, but it's at the very beginning of the call and I personally look at these regularly.

alecharp:
> Not sure what you exactly mean. It's one of the only ways to be sure of the version of Java and which exact Maven version is in use.

It's just we should have an explicit way to say which version of the JDK was used. Not having to go in the build logs to figure this out.

> What there be an alternative?

I don't know any, that's why I approved this. Maybe the build description? (not convinced when I write this).

> Also, yes, it's in the logs, but it's at the very beginning of the call and I personally look at these regularly.

I understand, but having solution to see this without having to load the build console would be as great imo. But this is good.

olivergondza:
Note that in general, there can be a number of process invocations all of them using different JDK. Pulling all of that to build level would just be confusing...

batmat:
I suppose there were enough approvals, can someone merge? I'm trying to clean my list of open PRs :-). Thanks

batmat:
Thanks Raul. Thanks everyone.

PR #83, opened 2018-12-20
Add shm-size to avoid selenium webdriver exceptions

MRamonLeon:
See https://github.com/jenkinsci/jenkins/pull/3818 and related PRs on [pipeline-library](https://github.com/jenkins-infra/pipeline-library), [jenkins](https://github.com/jenkinsci/jenkins) and [acceptance-test-harness](https://github.com/jenkinsci/acceptance-test-harness) to get context.

I managed to reproduce the environment locally and got  the exception _Failed to decode response from marionette_. After a lot of research and looking for the failures over internet, I think we have the fix. It's just **--shm-size 2g** when running the docker container. It was set in _ath-container.sh_ and _Jenkinsfile_ on ATH, but not in _runATH.groovy_ here. I saw it but I thought it was due to Jenkins OSS CI restrictions, but it fix the problem actually.

In addition, we can revert the change on https://github.com/jenkinsci/acceptance-test-harness/pull/469, I tested in my local environment without this catch and it still works. But as it seems to be innocuous, we can leave it there.

I've added a comment to avoid forgetting the reason. As Jenkins CI gets the latest library, I think we don't need to change the _essentials.yml_ of Jenkins.

@olivergondza @batmat @oleg-nenashev @alecharp @raul-arabaolaza

### DIFF
--- a/vars/runATH.groovy
+++ b/vars/runATH.groovy
@@ -130,7 +130,8 @@ def call(Map params = [:]) {
                     def currentJdk = jdk
                     def javaOptions = defaultJavaOptions.clone()
                     def commandBaseWithFutureJava = ""
-                    def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -u ath-user"
+                    //Add shm-size to avoid selenium.WebDriverException exceptions like 'Failed to decode response from marionette' and webdriver closed
+                    def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -u ath-user --shm-size 2g"
 
                     if(configFile) {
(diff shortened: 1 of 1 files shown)


batmat:
@MRamonLeon do we have a way to test that by passing raw args in a Jenkins Core PR. I'd love to see us be able to check this before we make it a new default. 

If impossible currently, I think then it would be **very** valuable to have an envelope var for future such cases.

This would be similar to when we use timestamped SNAPSHOTs to test new parent poms for example.

batmat:
Can someone with permission merge this? @raul-arabaolaza @oleg-nenashev @rtyler  maybe? Thank you

PR #84, opened 2019-01-03
[INFRA-1953] Make pipeline-library#runATH step runs with old ATH sources and images

MRamonLeon:
See [Make pipeline-library#runATH step runs with old ATH sources and images](https://issues.jenkins-ci.org/browse/INFRA-1953)

This PR allows you to use the runATH step with old ATH sources and images. With this, your builds will not fail if the pipeline-library is not on sync with the ATH sources and images used. 

@batmat @alecharp @raul-arabaolaza @olivergondza

### DIFF
--- a/vars/runATH.groovy
+++ b/vars/runATH.groovy
@@ -131,11 +131,12 @@ def call(Map params = [:]) {
                     def javaOptions = defaultJavaOptions.clone()
                     def commandBaseWithFutureJava = ""
                     //Add shm-size to avoid selenium.WebDriverException exceptions like 'Failed to decode response from marionette' and webdriver closed
-                    def containerargs = "-v /var/run/docker.sock:/var/run/docker.sock -u ath-user --shm-size 2g"
+                    def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -u ath-user --shm-size 2g"
 
                     if(configFile) {
                         containerArgs += " -e CONFIG=../${configFile}" // ATH runs are executed in a subfolder, hence path needs to take that into account
                     }
+
                     // Add options for jdks
                     if ( currentJdk  > 8) {
                         // Add environment variable
--- a/vars/runATH.groovy
+++ b/vars/runATH.groovy
@@ -132,16 +132,15 @@ def call(Map params = [:]) {
                     def commandBaseWithFutureJava = ""
                     //Add shm-size to avoid selenium.WebDriverException exceptions like 'Failed to decode response from marionette' and webdriver closed
                     def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -u ath-user --shm-size 2g"
+                    containerArgs += " -e java_version=${currentJdk}"
 
                     if(configFile) {
                         containerArgs += " -e CONFIG=../${configFile}" // ATH runs are executed in a subfolder, hence path needs to take that into account
                     }
-
                     // Add options for jdks
                     if ( currentJdk  > 8) {
                         // Add environment variable
                         commandBaseWithFutureJava = "JENKINS_OPTS=\"--enable-future-java\" "
-                        containerArgs += " -e java_version=${currentJdk}"
 
                         // Add modules removed
                         javaOptions << "-p /home/ath-user/jdk11-libs/jaxb-api.jar:/home/ath-user/jdk11-libs/javax.activation.jar"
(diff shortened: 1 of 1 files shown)


vars/runATH.groovy
```diff
@@ -198,7 +198,9 @@ private void test(discriminator, commandBase, localSnapshots, localPluginsStashN
     unstashResources(localSnapshots, localPluginsStashName)
     athContainerImage.inside(containerArgs) {
         realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
-            def command = './set-java.sh $java_version && eval "$(./vnc.sh)" && export DISPLAY=$BROWSER_DISPLAY && export SHARED_DOCKER_SERVICE=true && export EXERCISEDPLUGINREPORTER=textfile && ' + prepareCommand(commandBase, discriminator, localSnapshots, localPluginsStashName)
+            // Allow call old images without the set-java.sh script. A message is showed if it doesn't exist to alert
+            // that you are using a new pipeline-library with an old ATH source/image (... type: ./set-java.sh: not found)
+            def command = 'if type ./set-java.sh; then ./set-java.sh $java_version; fi; eval "$(./vnc.sh)" && export DISPLAY=$BROWSER_DISPLAY && export SHARED_DOCKER_SERVICE=true && export EXERCISEDPLUGINREPORTER=textfile && ' + prepareCommand(commandBase, discriminator, localSnapshots, localPluginsStashName)
```
alecharp: have you tried `if -x "set-java.sh"`? It means "file exists && executable".

I fill that the command starts to become really hard to follow and understand. Could it be split in different instructions / lines?

vars/runATH.groovy
```diff
@@ -200,7 +200,25 @@ private void test(discriminator, commandBase, localSnapshots, localPluginsStashN
         realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
             // Allow call old images without the set-java.sh script. A message is showed if it doesn't exist to alert
             // that you are using a new pipeline-library with an old ATH source/image (... type: ./set-java.sh: not found)
-            def command = 'if type ./set-java.sh; then ./set-java.sh $java_version; fi; eval "$(./vnc.sh)" && export DISPLAY=$BROWSER_DISPLAY && export SHARED_DOCKER_SERVICE=true && export EXERCISEDPLUGINREPORTER=textfile && ' + prepareCommand(commandBase, discriminator, localSnapshots, localPluginsStashName)
+            def command = '''
+            if [ -x ./set-java.sh ]; then
+                ./set-java.sh $java_version;
+            else
+                if [ $java_version -gt 8 ]; then
+                    echo "Info: ./set-java.sh not found because you are using old ATH sources, $java_version cannot be used, instead java 8 will be used"
```
olivergondza: Can we fail the step here instead so problems gets the attention they need?

vars/runATH.groovy
```diff
@@ -132,16 +132,15 @@ def call(Map params = [:]) {
                     def commandBaseWithFutureJava = ""
                     //Add shm-size to avoid selenium.WebDriverException exceptions like 'Failed to decode response from marionette' and webdriver closed
                     def containerArgs = "-v /var/run/docker.sock:/var/run/docker.sock -u ath-user --shm-size 2g"
+                    containerArgs += " -e java_version=${currentJdk}"
```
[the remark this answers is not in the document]
olivergondza: Please document this is for old ATH versions to minimize the surprise for future readers.

vars/runATH.groovy
raul-arabaolaza: I agree with Oliver here, if we are using incoherent versions hidding the thing in a line of log is going to make much harder to find the issue when everything blow up in core. Failing by design in this case is going to make the problem evident and force users to fix it

MRamonLeon: We can also mark the build as UNSTABLE if set-java.sh is not found even when running on java8.

olivergondza: Well, we know what java version did the user wish to use and we know whether the switcher exists which leaves us with 4 possibilities:

switcher exists / java version | 8 | 11
-----------------------------------------|---|----
yes | switch to it | switch to it
no | java 8 will be used | _unspecified java used_

I believe it has to fail in case of java 11 and no switcher present. The question is if we want to rely the absence of the switcher file in the container is a sufficient guarantee the java version is 8.

What @batmat is suggesting would probably work, though I do not like the way it complicates the pipeline-library for such a minor use-case.

MRamonLeon: No we have:

|set-java.sh exists / java_version specified | 8 (or null) | 11
|---|---|---|
| Yes |	SUCCESS | SUCCESS |
| No | SUCCESS (1) | FAILURE |

My proposal is:
(1) What about changing SUCCESS to UNSTABLE in this case

If we keep it as SUCCESS, the user is not aware of a new version of ATH, and (s)he's not going to update it. We only write this in the log:
`INFO: ./set-java.sh not found because you are using old ATH sources, please consider to update ATH sources and docker image`
And most likely the admin is not going to review all the log messages, as the build was SUCCESS.

But doing the result as UNSTABLE, (s)he become more conscious. 

WDYT? _SUCCESS_ or _UNSTABLE_?

MRamonLeon: Can we use **githubNotify** to add this info in the GH status, instead of using the job status? 

Actually a @daniel-beck's proposal :+1:

olivergondza: I do not feel strongly. What you say makes sense.

vars/runATH.groovy
```diff
@@ -198,10 +198,33 @@ private void test(discriminator, commandBase, localSnapshots, localPluginsStashN
...
+            If other case (java8 and lack set-java), write a message in the log
+             */
+            def command = '''
+            if [ -x ./set-java.sh ] && [ -n "$java_version" ] && [ "$java_version" -gt "8" ]; then
+                ./set-java.sh $java_version;
+            elif [ -n "$java_version" ] && [ "$java_version" -gt "8" ] && [ ! -x ./set-java.sh ]; then
+                echo "ERROR: ./set-java.sh not found because you are using old ATH sources and \\$java_version = $java_version specified. We cannot run on this JDK";
+                exit 1;
+            elif [ ! -x ./set-java.sh ]; then
+                echo "INFO: ./set-java.sh not found because you are using old ATH sources, please consider to update ATH sources and docker image";
```
MRamonLeon: Due to the few valuable outcome and the time required to add a meanful GitHub status on this case, I opted to leave it as a message. Java8 with old ATH sources is not failing actually.

vars/runATH.groovy
MRamonLeon: Decision at: https://github.com/jenkins-infra/pipeline-library/pull/84#pullrequestreview-189421087

vars/runATH.groovy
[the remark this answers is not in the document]
MRamonLeon: done

vars/runATH.groovy
MRamonLeon: @alecharp I used **type** because it prints a message indicating that the file is not there (_type: ./set-java.sh: not found_). Finally we have improved a lot the messages written, so I changed the if to your proposal.

MRamonLeon:
@olivergondza @raul-arabaolaza Does these changes look good to you? Thank you.

MRamonLeon:
@varyvol No specific reason. Probably I forgot to add the && after the if. Good point.

PR #85, opened 2019-02-12
[JENKINS-56018] Remove jaxb libraries and enable-future-java

MRamonLeon:
See [[JENKINS-56018] ATH Docker image should not be using Module patches starting from 2.163](https://issues.jenkins-ci.org/browse/JENKINS-56018).

Remove the extra params and the enable-future-java param to run the tests on Java 11.

It's not needed on Jenkins >=2.164, and it's the recommended version to test Java 11.

@jenkins-infra/java11-support @olivergondza @raul-arabaolaza


MRamonLeon:
AFAIK the stable versions are running only on Java8 with an old ATH docker image (1.62). This change shouldn't impact them.

PR #86, opened 2019-03-08
Offer recommended configurations for buildPlugin()

oleg-nenashev:
I am a bit tired of copy-pasting configurations for Java 11 between plugins, and I would like to add a utility method for it to the Pipeline library.

Question to reviewers:

*  Is it fine to use a generic `recommendedConfigurations()` method with a disclaimer that the configurations may change at any moment and break CI?
* If no, what would be your recommendations?

CC @batmat @alecharp @MRamonLeon @MarkEWaite @olivergondza @raul-arabaolaza

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -219,3 +219,21 @@ static List<Map<String, String>> getConfigurations(Map params) {
     }
     return ret
 }
+
+/**
+ * Get recommended configurations for testing.
+ * Includes testing Java 8 and 11 on the newest LTS.
+ */
+static List<Map<String, String>> recommendedConfigurations() {
+    //TODO: replace by 2.164.1 once it is released
+    def recentLTS = "2.164"
+    def configurations = [
+        [ platform: "linux", jdk: "8", jenkins: null ],
+        [ platform: "windows", jdk: "8", jenkins: null ],
+        [ platform: "linux", jdk: "8", jenkins: recentLTS, javaLevel: "8" ],
+        [ platform: "windows", jdk: "8", jenkins: recentLTS, javaLevel: "8" ],
+        [ platform: "linux", jdk: "11", jenkins: recentLTS, javaLevel: "8" ],
+        [ platform: "windows", jdk: "11", jenkins: recentLTS, javaLevel: "8" ]
+    ]
+    return configurations
+}
--- a/README.adoc
+++ b/README.adoc
@@ -31,9 +31,10 @@ buildPlugin()
 * `jenkinsVersions`: (default: `[null]`) - a matrix of Jenkins baseline versions to build/test against in parallel (null means default,
   only available for Maven projects)
 * `configurations`: An alternative way to specify `platforms`, `jdkVersions` and `jenkinsVersions` (that can not be combined
-  with any of them). Those options will run the build for all combinations of their values. While that is desirable in
+  with any of them).   
+** Those options will run the build for all combinations of their values. While that is desirable in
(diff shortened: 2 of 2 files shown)


vars/buildPlugin.groovy
```diff
@@ -219,3 +219,21 @@ static List<Map<String, String>> getConfigurations(Map params) {
     }
     return ret
 }
+
+/**
+ * Get recommended configurations for testing.
+ * Includes testing Java 8 and 11 on the newest LTS.
+ */
+static List<Map<String, String>> recommendedConfigurations() {
```
batmat: I'm sorry, I should have told you before, but I've been thinking about this, and I really think this function must take an argument so we can evolve it over time without breaking existing usages.

IOW, would become something like:

```groovy
static List<Map<String, String>> recommendedConfigurations(configDate) {
  if(version="2019-03-08") {
    //TODO: replace by 2.164.1 once it is released
    def recentLTS = "2.164"
    def configurations = [
        [ platform: "linux", jdk: "8", jenkins: null ],
        [ platform: "windows", jdk: "8", jenkins: null ],
        [ platform: "linux", jdk: "8", jenkins: recentLTS, javaLevel: "8" ],
        [ platform: "windows", jdk: "8", jenkins: recentLTS, javaLevel: "8" ],
        [ platform: "linux", jdk: "11", jenkins: recentLTS, javaLevel: "8" ],
        [ platform: "windows", jdk: "11", jenkins: recentLTS, javaLevel: "8" ]
    ]
    return configurations
  } else if ( version=...) {
    ...
  } else {
     throw new UnknownConfigException("blah");
  }
}
```

This way, we will be able to evolve the recommended config without suddenly breaking plugins builds out there.

WDYT?

oleg-nenashev:
Groovy supports default parameters, so we will be able to extend it later

batmat:
Yeah, why not. That should be OK.

vars/buildPlugin.groovy
jglick: > evolve the recommended config without suddenly breaking plugins builds out there

I have long advocated that plugins not simply import the `master` version of this library, but only a specific version, so that updates to the library can be rolled out in a controlled fashion in tested PRs. (With @dependabot, even!) @rtyler never did so; not sure if he specifically argued against it, or just did not want to spend the time.

vars/buildPlugin.groovy
```diff
@@ -219,3 +219,21 @@ static List<Map<String, String>> getConfigurations(Map params) {
...
+    //TODO: replace by 2.164.1 once it is released
+    def recentLTS = "2.164"
+    def configurations = [
+        [ platform: "linux", jdk: "8", jenkins: null ],
+        [ platform: "windows", jdk: "8", jenkins: null ],
+        [ platform: "linux", jdk: "8", jenkins: recentLTS, javaLevel: "8" ],
+        [ platform: "windows", jdk: "8", jenkins: recentLTS, javaLevel: "8" ],
+        [ platform: "linux", jdk: "11", jenkins: recentLTS, javaLevel: "8" ],
+        [ platform: "windows", jdk: "11", jenkins: recentLTS, javaLevel: "8" ]
```
jglick: Specifying `javaLevel` here seems wrong. It is overriding whatever it is defined in `pom.xml`. Is this just here as a workaround for old parent POMs which do not require an explicit `java.level`? If so, can we make it detect that version; or at least include a TODO comment that this should be reconsidered?

batmat: Yes, I believe we simply underestimated the speed at which at some point some plugins would start having their `jenkins.version` ≥ the current LTS specified here.

PR #87, opened 2019-03-18
buildPlugin() - Switch to the LTS release

oleg-nenashev:
CC @jenkins-infra/java11-support


oleg-nenashev:
@batmat I agree with you, bit this is how te method is documented now (explicit disclaimers). My plan is to add an option to put a warning instead of failing the build by default. Or we can refactor it later

PR #89, opened 2019-05-15
`buildPlugin()` - Archive Maven Failsafe Plugin reports

chriskilding:
Archive Failsafe reports as well as Surefire reports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 *.sw*
 build/
-
+.idea
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -115,7 +115,7 @@ def call(Map params = [:]) {
                         if (!skipTests) {
                             String testReports
                             if (isMaven) {
-                                testReports = '**/target/surefire-reports/**/*.xml'
+                                testReports = '**/target/surefire-reports/**/*.xml,**/target/failsafe-reports/**/*.xml'
                             } else {
                                 testReports = '**/build/test-results/**/*.xml'
                             }


chriskilding:
Questions:

- This project doesn’t have any tests. How can I test this change?
- Is `(surefire|failsafe)-reports` supported by the pattern matcher?

chriskilding:
Hi Oleg

I’ve tried to set up both of those test projects, but neither is working for me.

demo-Jenkins-config-as-code builds the master image, but then fails when running it because some plugin classes cannot be found:

    IOException: Branch API Plugin version 2.5.2 failed to load
    - Folders Plugin version 6.7 is older than required. To fix, install version 6.8 or later.

Meanwhile ci.jenkins.io-runner fails on the `make clean build` step because almost none of the plugins can download, and then the following error appears:

    INFO: Dependency <whatever> was found in the non-HPI cache. Delete <whatever>.hpi to attempt another resolution attempt.

All I’ve done is check out the latest master branch of each project locally and then followed the build instructions in each README, so would appreciate any ideas you have.

chriskilding:
After updating all the outdated plugin dependencies in plugins.txt to their latest versions, I got `demo-jenkins-config-as-code` to work.

chriskilding:
I can also now confirm that the change in this pull request works: it successfully archived the Warnings NG Plugin Failsafe test results.

oleg-nenashev:
Sorry, I missed the updates somehow. Will try to review and integrate the merge conflicts

oleg-nenashev:
Sorry for that taking so long

PR #90, opened 2019-05-31
[INFRA-2028] `buildPlugin()` - Remove the timestamps step which is no longer necessary

jglick:
[INFRA-2028](https://issues.jenkins-ci.org/browse/INFRA-2028)


jglick:
@olblak @rtyler

oleg-nenashev:
Will merge after #99

PR #91, opened 2019-06-10
[INFRA-2118] Add function to run JMH benchmarks

AbhyudayaSharma:
Uses the added Maven profile from https://github.com/jenkinsci/plugin-pom/pull/206 and https://github.com/jenkinsci/jenkins-test-harness/pull/135
 and adds a utility function to run benchmarks in Jenkinsfiles.

@oleg-nenashev @runzexia @supun94

### DIFF
--- a/vars/runBenchmarks.groovy
+++ b/vars/runBenchmarks.groovy
@@ -6,23 +6,24 @@
  * @since TODO
  */
 def call(String artifacts = null) {
-    // TODO: Use Lockable Resources plugin
-    node('highmem') {
+    lock('runBenchmarks') {
+        node('highmem') {
 
-        stage('Checkout repo') {
-            infra.checkout()
-        }
+            stage('Checkout repo') {
+                infra.checkout()
+            }
 
-        stage('Run Benchmarks') {
-            List<String> mvnOptions = ['test', '-Dbenchmark']
-            infra.runMaven(mvnOptions)
-        }
+            stage('Run Benchmarks') {
+                List<String> mvnOptions = ['test', '-Dbenchmark']
+                infra.runMaven(mvnOptions)
+            }
 
-        stage('Archive reports') {
-            if (artifacts != null) {
-                archiveArtifacts artifacts: artifacts
-            } else {
-                echo 'No artifacts to archive, skipping...'
+            stage('Archive reports') {
+                if (artifacts != null) {
+                    archiveArtifacts artifacts: artifacts
+                } else {
+                    echo 'No artifacts to archive, skipping...'
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


vars/runBenchmarks.groovy
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env groovy
+
+/**
+ * Run JMH benchmarks and archives results
+ * @param artifacts archive these JMH reports
+ * @since TODO
+ */
+def call(String artifacts = null) {
+    // TODO: Use Lockable Resources plugin
```
oleg-nenashev: I am not sure we can merge it without it. Maybe it makes sense to create an INFRA ticket to @olblak to get it set up. Maybe 2 parallel runs is a good limit for now.

vars/runBenchmarks.groovy
```diff
@@ -0,0 +1,29 @@
...
+            infra.checkout()
+        }
+
+        stage('Run Benchmarks') {
+            List<String> mvnOptions = ['test', '-Dbenchmark']
+            infra.runMaven(mvnOptions)
+        }
+
+        stage('Archive reports') {
+            if (artifacts != null) {
```
oleg-nenashev: Undefined variable

vars/runBenchmarks.groovy
```diff
@@ -0,0 +1,29 @@
...
+                archiveArtifacts artifacts: artifacts
```
oleg-nenashev: Would it also make sense to generate a http://jmh.morethan.io/ viewer link and put it to the console log for now?

vars/runBenchmarks.groovy
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env groovy
```
oleg-nenashev: There should be also a `runBenchmarks.txt` HTML text file. It is used for generating built-in Documentation in Jenkins IIRC

vars/runBenchmarks.groovy
AbhyudayaSharma: `artifacts` is a parameter to the function

vars/runBenchmarks.groovy
AbhyudayaSharma: Created https://issues.jenkins-ci.org/browse/INFRA-2141

olblak: the lockable plugin is already installed

AbhyudayaSharma: @olblak We need you to create a resource that we can use for throttling benchmarks running on `highmem` nodes.

olblak: so something like lock('runBenchmarks') {
}

should be fine

AbhyudayaSharma:
Yes, that would be great!

oleg-nenashev:
" IMHO we can ship it once JTH and Plugin POM patches are out" @olblak They are not :) No worries, we will get there soon

PR #92, opened 2019-06-11
Update the recommended LTS version from 2.164.1 to 2.176.1 in `buildPlugin.recommendedConfigurations`

jglick:
See https://github.com/jenkinsci/kubernetes-plugin/pull/461/commits/e71a5c5950ee976da51bf3046775ff2b22e4c170 for example: cannot use `recommendedConfigurations()` if your `jenkins.baseline` is _newer_ than the `recentLTS`! (If it were the _same_, it would be possible, though wasteful.)

In general I am not a fan of `recommendedConfigurations`. It means that a critical aspect of a plugin’s build is controlled by a library it consumes in a `master` revision, making the build nondeterministic. See https://github.com/jenkins-infra/pipeline-library/pull/86#discussion_r263697425 for discussion. It would be OK if the `recentLTS` were a parameter passed to the method, `null` meaning skip those branches, so you could begin testing against a new LTS line in a controlled fashion in a PR. @oleg-nenashev @batmat

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -225,7 +225,7 @@ static List<Map<String, String>> getConfigurations(Map params) {
  * Includes testing Java 8 and 11 on the newest LTS.
  */
 static List<Map<String, String>> recommendedConfigurations() {
-    def recentLTS = "2.176.1"
+    def recentLTS = "2.176.3"
     def configurations = [
         [ platform: "linux", jdk: "8", jenkins: null ],
         [ platform: "windows", jdk: "8", jenkins: null ],
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -225,7 +225,7 @@ static List<Map<String, String>> getConfigurations(Map params) {
  * Includes testing Java 8 and 11 on the newest LTS.
  */
 static List<Map<String, String>> recommendedConfigurations() {
-    def recentLTS = "2.176.3"
+    def recentLTS = "2.222.3"
     def configurations = [
         [ platform: "linux", jdk: "8", jenkins: null ],
         [ platform: "windows", jdk: "8", jenkins: null ],
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -225,7 +225,7 @@ static List<Map<String, String>> getConfigurations(Map params) {
  * Includes testing Java 8 and 11 on the newest LTS.
  */
 static List<Map<String, String>> recommendedConfigurations() {
-    def recentLTS = "2.164.1"
+    def recentLTS = "2.176.1"
     def configurations = [
         [ platform: "linux", jdk: "8", jenkins: null ],
         [ platform: "windows", jdk: "8", jenkins: null ],


jglick:
> allowing users to select a "version" of the recommended configuration

My recommendation from the start of this library has been that we turn off the implicit load and `master` version, so that every `Jenkinsfile` which wants explicitly asks for it in a particular version. Then we would use Dependabot to _propose_ updates to `Jenkinsfile`s.

rtyler:
> My recommendation from the start of this library

@jglick I am never sure how you mean these comments, but more than others you consistently make them, and every time I read one it seems like nothing more than "I told you so" or "Everybody should be doing things how I want"

I decided to simply disengage with any topic (or INFRA ticket) where a developer makes a comment like this.

jglick:
Well I cannot pretend that I did _not_ make this recommendation long ago, or that I would not want everybody to be doing things the way I want (which is tautological), but yes my attitude was poor. So maybe I should start over. I am not really sure whom I am addressing, since I have no visibility into who actually maintains ci.jenkins.io on a day-to-day basis, or what its roadmap looks like. Clearly it needs a lot of things overhauled, and it is hard to see which priorities are the most urgent.

----

The principle of reproducibility in software builds dictates that, to the extent possible given the realities of available technology, the outcome of a build in a given software revision—which in the context of things managed on GitHub means a commit in a repository—should be a deterministic function of the contents of that revision (typically, tree of files). It should not depend on who ran the build, or in what environment, or when, or what changes may have been happening in other repositories. There are many aspects of this—for example, Docker images with immutable tags should be preferred as a build environment to vaguely named tools on some unspecified build VM—but in this context the particular issue is that a (Groovy) library is being imported from one repository (this) into another (`*-plugin`) without pinning the version of the import. That is almost guaranteed to cause abrupt build failures (or, worse, passing builds with unwanted behaviors), even for example in a plugin pull request that merely corrects a typo in `README.md` and thus could not possibly have been responsible.

#94 is better than nothing, as it controls for one important source of variation (future Jenkins core cross-testing), but there are many other subtle changes to this library that would not be covered. It would be better to really solve the problem without a lot of special-case parameters.

The main argument I am aware of _for_ having a nondeterministic import is to minimize the work required of a plugin maintainer in getting the latest recommended behaviors. This is a serious consideration, and Dependabot is the best available solution, which we already advocate for example to update the `plugin-pom` import. We would presumably need to offer a patch for it to support `workflow-cps-global-lib`, but this is feasible and would benefit the whole Jenkins community.

vars/buildPlugin.groovy
```diff
@@ -225,7 +225,7 @@ static List<Map<String, String>> getConfigurations(Map params) {
  * Includes testing Java 8 and 11 on the newest LTS.
  */
 static List<Map<String, String>> recommendedConfigurations() {
-    def recentLTS = "2.164.1"
+    def recentLTS = "2.176.1"
```
oleg-nenashev:
```suggestion
    def recentLTS = "2.176.3"
```

oleg-nenashev:
https://github.com/search?q=org%3Ajenkinsci+recommendedConfigurations&type=Code shows that 70 plugins now use the `recommendedConfigurations()`. I'd guess it actually worth asking in the developer mailing list

oleg-nenashev:
https://groups.google.com/forum/#!topic/jenkinsci-dev/lmZ_e7eFh2k

bitwiseman:
@jenkins-infra/ci-maintainers 
Any chance we can go ahead on this now.  It's been quite a while and there have been no comments or further responses to the email.

vars/buildPlugin.groovy
```diff
@@ -225,7 +225,7 @@ static List<Map<String, String>> getConfigurations(Map params) {
  * Includes testing Java 8 and 11 on the newest LTS.
  */
 static List<Map<String, String>> recommendedConfigurations() {
-    def recentLTS = "2.164.1"
+    def recentLTS = "2.176.3"
```
bitwiseman, line 228: Months later now... 
Perhaps it we can change this to a currently recent LTS? 
```suggestion
    def recentLTS = "2.204.6"
```

jglick: Or 2.222.3?

jglick:
My personal preference is to deprecate `recommendedConfigurations` as a failed experiment, and find some other mechanism for guiding plugin developers to build against both newer LTS versions and Java 11.

vars/buildPlugin.groovy
olblak:
```suggestion
    def recentLTS = "2.222.3"
```

bitwiseman:
@jglick 

> My personal preference is to deprecate `recommendedConfigurations` as a failed experiment, and find some other mechanism for guiding plugin developers to build against both newer LTS versions and Java 11.

Perhaps this is something that should be tracked in a JIRA? It may not get done, but filing it there would be more effective that tacking it on as comment on tail end of this PR.

jglick:
`INFRA` tickets are not processed reliably, I would not bother. More of a topic for discussion than a specific task.

PR #94, opened 2019-07-06
Allow to use a more recent LTS version

batmat:
This allows mainly plugins whose baseline is > 2.164.x to still use `recommendedConfigurations`.

We do this by adding a `version` parameter.

Alternative (untested) to https://github.com/jenkins-infra/pipeline-library/pull/92 by @jglick 

cc @jenkins-infra/java11-support

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -224,8 +224,13 @@ static List<Map<String, String>> getConfigurations(Map params) {
  * Get recommended configurations for testing.
  * Includes testing Java 8 and 11 on the newest LTS.
  */
-static List<Map<String, String>> recommendedConfigurations() {
-    def recentLTS = "2.164.1"
+static List<Map<String, String>> recommendedConfigurations(version=1) {
+    def recentLTS = null
+    if(version==1) {
+        recentLTS = "2.164.1"
+    } else if (version==2) {
+        recentLTS = "2.176.1"
+    }
     def configurations = [
         [ platform: "linux", jdk: "8", jenkins: null ],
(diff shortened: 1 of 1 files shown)


jglick:
We could, but would that you really buy you much over just not using `recommendedConfiguration` to begin with and specifying your matrix explicitly?

I do not want to be blocking progress here, but I advocate trying to satisfy these goals:

* A given textual revision of a `Jenkinsfile` has a deterministic behavior. (To the best of our ability to control the environment.)
* It is easy for a plugin author to request testing against a fairly recent version of core, and against Java 11.
* The default configuration, or the configuration of least resistance, should not consume excessive cloud resources. (I think we could pare down the number of branches to just two: Windows with default POM; and Linux with Java 11 and a newer core.)
* Dependabot can be used to upgrade tests to the newest recommendations, without breaking `master` builds.

Number 3 is perhaps independent. The most straightforward way to accomplish numbers 1, 2, and 4 that I can think of is to switch `Jenkinsfile`s to pinning a version of the library and patch Dependabot to grok Jenkins libraries. #92 would be compatible with that approach but of course more work would be needed.

oleg-nenashev:
> We could, but would that you really buy you much over just not using recommendedConfiguration to begin with and specifying your matrix explicitly?

Mote that we specify 6 configurations and pass 4 parameters in the recommendedConfigurations(). So it is not just about core version. It just keeps the things simple

oleg-nenashev:
https://groups.google.com/forum/#!topic/jenkinsci-dev/lmZ_e7eFh2k

jglick:
Superseded.

PR #95, opened 2019-07-30
Use ACI agents for Linux based builds.

slide:
This will use ACI (Azure Container Instances) agents instead of VM's. The default is to use ACI agents, with the ability to override in the plugin configuration.

CC @rtyler, @jglick, @daniel-beck

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -23,6 +23,7 @@ buildPlugin()
 ==== Optional arguments
 
 * `repo` (default: `null`  inherit from Multibranch) - custom Git repository to check out
+* `useAci` (default: `false`) - uses link:https://azure.microsoft.com/en-us/services/container-instances/[Azure Container Instances] for build instead of link:https://azure.microsoft.com/en-us/services/virtual-machines/[Azure Virtual Machines].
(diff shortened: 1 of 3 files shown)


rtyler:
This looks reasonable to me, I could have sworn I saw a similar change from @jglick somewhere already but now cannot find it.

Would it perhaps make sense to swap the default to _not_ use ACI, and then test against a few plugin repos?

Or would you rather #yolo and see what happens? :smile_cat:

slide:
I was planning on #yolo :-)

slide:
I believe @jglick's changes were to the Jenkinsfile for jenkinsci/jenkins

slide:
If everyone would rather play it safe, I can switch the default.

oleg-nenashev:
Our infrastructure breaks every week anyway, so I do not mid YOLO if it is
documented

jglick:
Switching the default would break a bunch of builds in my experience:

* no `docker` CLI
* tests run as `root`

So let us make it opt-in.

jglick:
[INFRA-2165](https://issues.jenkins-ci.org/browse/INFRA-2165) FTR

PR #93, opened 2019-06-12
Add Jenkins Pipeline Unit to the repository

v1v:
This PR is just a PoC using the https://github.com/jenkinsci/JenkinsPipelineUnit for the `vars/infra.groovy` steps. 

## Highlights
- UTs based on https://github.com/jenkinsci/JenkinsPipelineUnit
- Jenkinsfile to test the shared library

## Tasks
- [ ] Agree whether to move forward with the rest of steps.
- [x] Ask for MB in ci.jenkins.io [#INFRA-2147](https://issues.jenkins-ci.org/browse/INFRA-2147)

## Proposed
- Create simple vars/steps for the methods within the `vars/infra.groovy` to simplify the UTs and the maintenance in the future.

PR #96, opened 2019-07-31
Fix issues with previous PR for ACI

slide:
There was a syntax issue with the previous PR. This fixes it as well as fixes the buildPlugin for non-linux builds.


oleg-nenashev:
i think we should finish #93 and require tests for new changes :D

slide:
Agreed! Much less YOLO with that approach

rtyler:
Hah, I was about to ask "what happened to #yolo?"

PR #78, opened 2018-12-01
Add support of ci.jenkins.io-runner + introduce `PIPELINE_LIBRARY_SKIP_WINDOWS`, `PIPELINE_LIBRARY_USE_DEFAULT_MAVEN_REPO` and `JFR_LOCAL_WORKSPACE` variables

oleg-nenashev:
These patches reflect changes needed in order to run this library inside Jenkinsfile Runner.

Here is my sandbox project for such mode: https://github.com/oleg-nenashev/ci.jenkins.io-runner

PR #99, opened 2019-08-13
`buildPlugin()` - Cleanup plugin workspace and Docker images after build

timja:
Review with no whitespace:
https://github.com/jenkins-infra/pipeline-library/pull/99/files?w=1

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -41,132 +41,140 @@ def call(Map params = [:]) {
 
         tasks[stageIdentifier] = {
             node((useAci && label == 'linux') ? (jdk == '8' ? 'maven' : 'maven-11') : label) {
-                timeout(timeoutValue) {
-                    boolean isMaven
-                    // Archive artifacts once with pom declared baseline
-                    boolean doArchiveArtifacts = !jenkinsVersion && !archivedArtifacts
-                    if (doArchiveArtifacts) {
-                        archivedArtifacts = true
-                    }
-                    boolean incrementals // cf. JEP-305
-
-                    stage("Checkout (${stageIdentifier})") {
-                        infra.checkout(repo)
-                        isMaven = fileExists('pom.xml')
-                        incrementals = fileExists('.mvn/extensions.xml') &&
-                                readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
-                    }
-
-                    String changelistF
-                    String m2repo
-
-                    stage("Build (${stageIdentifier})") {
-                        String command
-                        if (isMaven) {
-                            m2repo = "${pwd tmp: true}/m2repo"
-                            List<String> mavenOptions = [
-                                    '--update-snapshots',
-                                    "-Dmaven.repo.local=$m2repo",
-                                    '-Dmaven.test.failure.ignore',
-                            ]
-                            if (incrementals) { // set changelist and activate produce-incrementals profile
-                                mavenOptions += '-Dset.changelist'
-                                if (doArchiveArtifacts) { // ask Maven for the value of -rc999.abc123def456
-                                    changelistF = "${pwd tmp: true}/changelist"
-                                    mavenOptions += "help:evaluate -Dexpression=changelist -Doutput=$changelistF"
-                                }
-                            }
-                            if (jenkinsVersion) {
-                                mavenOptions += "-Djenkins.version=${jenkinsVersion} -Daccess-modifier-checker.failOnError=false"
-                            }
-                            if (javaLevel) {
-                                mavenOptions += "-Djava.level=${javaLevel}"
-                            }
-                            if (params?.findbugs?.run || params?.findbugs?.archive) {
-                                mavenOptions += "-Dfindbugs.failOnError=false"
-                            }
-                            if (skipTests) {
-                                mavenOptions += "-DskipTests"
-                            }
-                            if (params?.checkstyle?.run || params?.checkstyle?.archive) {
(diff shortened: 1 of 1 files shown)


vars/buildPlugin.groovy
```diff
@@ -41,132 +41,140 @@ def call(Map params = [:]) {
...
+                                        artifacts = '**/build/libs/*.hpi,**/build/libs/*.jpi'
+                                    }
+                                    archiveArtifacts artifacts: artifacts, fingerprint: true
                                 }
-                                archiveArtifacts artifacts: artifacts, fingerprint: true
                             }
                         }
+                    }            
+                } finally {
+                    deleteDir()
```
oleg-nenashev: Might be better to use https://jenkins.io/doc/pipeline/steps/ws-cleanup/

timja: Is that available on ci.jenkins.io? can change if you think it's worth it, I think it would be:
```suggestion
                    cleanWs
```

jglick: Would be `cleanWs()`. Anyway I see no need for a special plugin, `deleteDir` works fine.

Do we actually need to do this? Just means that cloning will be slower (no update) if we do happen to reuse workspaces. I doubt that dirty workspaces are actually an issue (I have never seen any such problems), but if so I would rather recommend

```sh
git clean -f -f -d -x
```

vars/buildPlugin.groovy
```diff
@@ -41,132 +41,140 @@ def call(Map params = [:]) {
...
+                                    } else {
+                                        artifacts = '**/build/libs/*.hpi,**/build/libs/*.jpi'
+                                    }
+                                    archiveArtifacts artifacts: artifacts, fingerprint: true
                                 }
-                                archiveArtifacts artifacts: artifacts, fingerprint: true
                             }
                         }
+                    }            
+                } finally {
```
jglick: Probably does not really need a `finally` block, since most builds will pass and this is just a best effort anyway.

vars/buildPlugin.groovy
```diff
@@ -41,132 +41,140 @@ def call(Map params = [:]) {
...
+                    deleteDir()
+
+                    if (env?.NODE_LABELS?.contains("docker")) {
```
jglick: Or would suffice to just

```groovy
if (isUnix()) {
    sh 'docker system prune --force --all || :'
}
```

and let it just fail if Docker is not present.

vars/buildPlugin.groovy
oleg-nenashev: @olblak says the plugin is not installed, so it is a separate story

jglick: By the way another good practice for CI (to run before `clean`):

```groovy
if (isUnix()) {
    sh 'git add . && git diff --exit-code HEAD'
}
```

which will fail if the build modified any versioned files, or created build products not in the ignore list.

vars/buildPlugin.groovy
timja: does it do any harm though? should be more reliable?

vars/buildPlugin.groovy
timja: > Would be `cleanWs()`. Anyway I see no need for a special plugin, `deleteDir` works fine.
> 
> Do we actually need to do this? Just means that cloning will be slower (no update) if we do happen to reuse workspaces. I doubt that dirty workspaces are actually an issue (I have never seen any such problems), but if so I would rather recommend
> 
> ```shell
> git clean -f -f -d -x
> ```

Happy to integrate if required, but not sure if needed?

timja:
not sure if the CI failure is related?

oleg-nenashev:
@timja Nope. I just merged the CI pull request and it broke. https://github.com/jenkins-infra/pipeline-library/pull/100 for the fix

vars/buildPlugin.groovy
olblak: Every linux machine have docker installed

timja: does isUnix pass on ACI?
I've pushed @jglick 's suggested change but a little worried that it will just always fail on ACI, not that it matter's much as the command has `||` on it to make it never fail the build

vars/buildPlugin.groovy
```diff
@@ -181,6 +189,10 @@ def call(Map params = [:]) {
     }
 }
 
+static boolean hasDockerLabel() {
+    env?.NODE_LABELS?.contains("docker")
```
jglick: Unnecessary null checks.

```suggestion
    env.NODE_LABELS.contains('docker')
```

vars/buildPlugin.groovy
jglick: Only brought it up since it forced a lot of reindentation.

oleg-nenashev:
I tested the patch on https://github.com/oleg-nenashev/demo-jenkins-config-as-code , and the build still passes there. No Windows checks applied. Once https://github.com/jenkins-infra/pipeline-library/pull/78 is integrated we can do some integration testing using https://github.com/jenkinsci/ci.jenkins.io-runner

oleg-nenashev:
A bit YOLO, but 🚢 🇮🇹

timja:
```
Total reclaimed space: 6.754GB
```
👍

vars/buildPlugin.groovy
```diff
@@ -181,6 +189,10 @@ def call(Map params = [:]) {
     }
 }
 
+static boolean hasDockerLabel() {
```
jglick: amended in #101

PR #102, opened 2019-08-15
Add Unit Tests for `buildPlugin` and `runBenchmarks`

v1v:
### Highlights
- Add UTs for `buildPlugin` and `runBenchmarks`
- Refactor mock classes to be reused
- Required to remove the `static` for the method `getConfigurations` otherwise the UTs won't run as expected

### Actions
- [ ] Confirm whether `static` is mandatory or no.


v1v:
That's indeed a great approach, I'll explore it then

oleg-nenashev:
I should have a working example soon.

PR #107, opened 2019-08-15
Switch `infra.maybePublishIncrementals()` to use an Azure Container Instance agent

jglick:
To allow https://github.com/jenkinsci/jenkins/pull/4153#discussion_r314395511 to be reverted.

(Why `maven`? Well, because it has `curl` installed, and plain `alpine` does not.)

@oleg-nenashev @rtyler

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -242,7 +242,7 @@ void ensureInNode(env, nodeLabels, body) {
 void maybePublishIncrementals() {
     if (isRunningOnJenkinsInfra() && currentBuild.currentResult == 'SUCCESS') {
         stage('Deploy') {
-            node('linux') {
+            node('maven') {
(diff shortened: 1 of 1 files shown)


jglick:
Note that I did _not_ add some sort of TODO comment to revert this once the Azure VM outage is fixed, because this patch would be desirable anyway—we do not want Incrementals publishing to have to wait for a limited pool of VMs. ACI agents start a bit faster, and this is just a super simple task. (There is even a plugin with an HTTP step we could use instead, but I prefer CLI for testability.)

oleg-nenashev:
Merging to unblock Plugins using Incrementals

jglick:
I have no real way of testing effectiveness of this beyond checking that the Docker image used by this agent class does indeed have `curl` installed. So we will need to merge it and hope for the best.

PR #110, opened 2019-08-27
Add forceAci parameter to `buildPlugin()`

slide:
Adds a forceAci parameter to allow testing on Windows ACI agents. Setting forceAci to true will override the label check for `== 'linux'`


oleg-nenashev:
It also looks like we should create an ACI EPIC in JIRA. Too many topics to cover in the library/infra

PR #112, opened 2019-09-02
[INFRA-2243] - JUnit archiver was taking a non-existent property in `buildPluginWithGradle()`

oleg-nenashev:
My mistake during the last rounds of refactoring


oleg-nenashev:
Merging just because it is a stupid mistake from my side, should not impact any existing flow

PR #113, opened 2019-09-12
feat: archive tests even when the build failed

v1v:
## Highlights
- I've found the JUnit step doesn't happen when there is a build error.
- The `Archive` stage has been moved within the finally block as suggested in https://github.com/jenkins-infra/pipeline-library/blob/478d9a971e23d8ec5ca57019cd3f478f42b6968e/vars/buildPlugin.groovy#L137
- Add UTs to validate whether a build failure triggers the `Archive` stage.

### DIFF
--- a/src/test/groovy/BuildPluginStepTests.groovy
+++ b/src/test/groovy/BuildPluginStepTests.groovy
@@ -188,6 +188,12 @@ class BuildPluginStepTests extends BasePipelineTest {
     }.any { call ->
       callArgsToString(call).contains('windows')
     })
+    // then the archive stage happens
+    assertTrue(helper.callStack.findAll { call ->
+      call.methodName == 'stage'
+    }.any { call ->
+      callArgsToString(call).contains('Archive')
+    })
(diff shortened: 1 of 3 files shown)


jglick:
Could also use `realtimeJUnit`.

v1v:
> Could also use realtimeJUnit.

@jglick , If that's the case then `tests: [ skip: true]` should be deprecated, right? I mean, currently, the junit happens at the very end of the pipeline and can be skipped with the above parameter.

I'd rather prefer to keep this change as simple as possible rather than using the suggested step, if possible.

WDYT?

jglick (approved):
Merge conflicts, caused by reindentation probably.

dduportal:
Hello @v1v , may I push on this branch to updatge (and land) this PR?
Or should we close the PR instead?

dduportal:
Closing this PR given the age. Please feel free to reopen it (and rebase + fix conflicts) if this is still current and should be merged.

PR #114, opened 2019-09-17
Enable ACI for Linux by default

slide:
This enables using ACI agents for Linux by default, there is still the ability to opt out by setting `useAci: false` in the call to `buildPlugin`

### DIFF
--- a/src/test/groovy/BuildPluginStepTests.groovy
+++ b/src/test/groovy/BuildPluginStepTests.groovy
@@ -181,7 +181,7 @@ class BuildPluginStepTests extends BasePipelineTest {
       call.methodName == 'node'
     }.any { call ->
       def args = callArgsToString(call)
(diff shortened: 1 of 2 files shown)


slide:
Any feedback?

src/test/groovy/BuildPluginStepTests.groovy
```diff
@@ -180,7 +180,8 @@ class BuildPluginStepTests extends BasePipelineTest {
     assertTrue(helper.callStack.findAll { call ->
       call.methodName == 'node'
     }.any { call ->
-      callArgsToString(call).contains('linux')
+      def args = callArgsToString(call)
+      args.contains('linux') || args.contains('maven')
```
v1v: I'd say the label should be `maven` only and a new test case might be required when `useAci: true`.

PR #115, opened 2019-09-24
Stop trying to deleteDir in buildPlugin

jglick:
See https://github.com/jenkins-infra/pipeline-library/pull/99#discussion_r313400182 by @timja. Seems to cause [errors](https://ci.jenkins.io/job/Plugins/job/pam-auth-plugin/job/PR-12/2/console) such as

```
java.nio.file.FileSystemException: C:\Jenkins\workspace\Plugins_pam-auth-plugin_PR-12\.git\objects\pack\pack-5e6e1b080765685b35cce003cf48670eaea3a58e.pack: The process cannot access the file because it is being used by another process.
```

Since I do not see a compelling reason to delete the workspace after a build—pointless on a one-off agent, possibly harmful to performance on a reused agent—and it can apparently fail, better to just revert this aspect.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -182,8 +182,6 @@ def call(Map params = [:]) {
                         }
                     }            
                 } finally {
-                    deleteDir()
-
                     if (hasDockerLabel()) {
                         if(isUnix()) {
                             sh 'docker system prune --force --all || echo "Failed to cleanup docker images"'


vars/buildPlugin.groovy
```diff
@@ -182,8 +182,6 @@ def call(Map params = [:]) {
                         }
                     }            
                 } finally {
-                    deleteDir()
-
                     if (hasDockerLabel()) {
```
timja: We could try this now that quite a few places are using ACI which I believe are one time agents? but previously we were re-using agents on every build and they were filling up constantly, (the docker prune change may be enough to fix this)

I'm ±0 it's easy to try out, but may bring back disk space issues

some other options would be:
```suggestion
                    try {
                      deleteDir()
                    } catch(ignored) {}
                    
                    if (hasDockerLabel()) {
```

```suggestion
                    if (isUnix()) {
                      deleteDir()
                    }
                    
                    if (hasDockerLabel()) {
```

PR #116, opened 2019-10-07
Try and debug Windows file locking

slide:
Currently, there are issues with deleteDir() on ci.j.io which is causing builds to fail. In the case that deleteDir() fails, we download handle.exe and run it on the .git directory to see what has the directory open. This is mainly a debug step to determine what is going on.


slide:
deleteDir was removed

PR #119, opened 2019-11-07
[INFRA-2402] - Skip tests if changeset content not relevant

MarkEWaite:
## Skip tests if changeset content not relevant

If the changeset does not include files that end with ".java" or a file with name that ends with "pom.xml", skip the tests.

Reduce the CI load while still performing simple evaluation of changes which don't modify Java files or pom files.

Do not skip tests on first build unless the skip test argument is passed explicitly.

@oleg-nenashev is this a useful short-term reduction in CI load when the tests are unlikely to tell us more about the change?

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -289,5 +289,6 @@ boolean skipTestsIfNoRelevantChanges(skipTestsInitialValue) {
             }
         }
     }
+    echo "Skipping tests because changeset does not contain java or pom.xml changes"
     return true
 }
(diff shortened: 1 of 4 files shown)


vars/buildPlugin.groovy
```diff
@@ -261,3 +262,42 @@ static List<Map<String, String>> recommendedConfigurations() {
     ]
     return configurations
 }
+
+/**
+ * Return true if tests should be skipped because the changeset
+ * contains no changes that will reasonably affect a test.
+ */
+boolean skipTestsIfNoRelevantChanges(skipTestsInitialValue) {
```
oleg-nenashev: It should be NonCPS to avoid massive performance overhead

oleg-nenashev:
Created https://issues.jenkins-ci.org/browse/INFRA-2402 for this story

MarkEWaite:
Need rework and more accuracy deciding which tests need to be run based on the contents of the changeset.

PR #120, opened 2019-11-07
Reduce recommended configurations

MarkEWaite:
## Reduce recommended configurations to best bug finders

Reduce recommended configurations to best bug finders

Assure that the recommended configuration will test most likely bug finding combinations:

* Windows, JDK 8, recent LTS
* Linux, JDK 11, recent LTS
* Linux, JDK 8, minimum Jenkins version

Don't need to cover every combination, just the most interesting combinations.

Will reduce the load on ci.jenkins.io and may decrease pull request build time.

@oleg-nenashev and @olblak this will reduce the load on ci.jenkins.io by running fewer configurations.  Reduces tested configurations by 50%.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -252,12 +252,16 @@ List<Map<String, String>> getConfigurations(Map params) {
 static List<Map<String, String>> recommendedConfigurations() {
     def recentLTS = "2.164.1"
     def configurations = [
+        // Intentionally test configurations which have detected the most problems
+        // Linux - Java 8 with plugin specified minimum Jenkins version
+        // Windows - Java 8 with recent LTS
+        // Linux - Java 11 with recent LTS
         [ platform: "linux", jdk: "8", jenkins: null ],
-        [ platform: "windows", jdk: "8", jenkins: null ],
-        [ platform: "linux", jdk: "8", jenkins: recentLTS, javaLevel: "8" ],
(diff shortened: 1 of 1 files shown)


MarkEWaite:
I disagree that Windows should be tested for the minimum Jenkins core version supported by the plugin.  I've never encountered a case where that combination would detect something that Linux and plugin selected minimum Jenkins core version did not also detect.  

I've not seen as many complex environments as you have through your work on remoting and other platform areas, so I'm open to learn of cases where Windows with minimum core version detected a problem that Linux with minimum Jenkins core version did not detect.

Windows runs tests more slowly (at least in my experience).  Retaining a Windows test configuration that is not a strong benefit will keep our resource usage higher than necessary without significant benefit to plugin maintainers or users.

MarkEWaite:
I'm open to consider a keystone build.  It seems like an interesting concept.

I am concerned by the number of times that I've waited while a ci.jenkins.io job paused to acquire a new Windows machine for a build or paused to acquire a new `maven` agent for its deploy stage.  With those pauses for agents, I'm hesitant to encourage adding more stages.  It means that what is now described as a parallel process (validate configurations in parallel) will become a serial step followed by a parallel process.  Won't that double the average elapsed time of the job?

oleg-nenashev:
depends on how we implement it, but indeed it would almost double the build
speed if done incorrectly.
We can reduce the impact by doing spotchecks as a keystone build on quickly
provisioned ACI containers

PR #121, opened 2019-11-26
[INFRA-2294][INFRA-2000] Enable Warnings Next Generation Plugin

uhafner:
This PR will remove the old static analysis configuration and replace it with a configuration of the 
[Warnings Next Generation plugin](https://github.com/jenkinsci/warnings-ng-plugin/). See [example output](https://ci.jenkins.io/job/Plugins/job/analysis-model/view/change-requests/job/PR-455/lastBuild/) to get an impression of the new configuration.

I removed the old configuration of Checkstyle and Findbugs and replaced it with the corresponding Warnings plugin configuration. There are two open things that need to be discussed (and may be added in another PR):
- I removed the maven configuration on how to start the tools. I'm not sure if I am using Maven the wrong way but I think that this part should belong to the pom and not into the `buildPlugin` step. In my jobs I configure all analysis tools in the pom so that `maven verify` will invoke automatically the static analysis tools that I want (with my configuration files and options). The warnings plugin is smart enough to not break the build if there are no reports written. 
- I did not yet activate the quality gates. I'm not sure if  someone already is using that feature? A total number of 0 warnings  quite hard to achieve.

Detailed changes summary:
- Removed FindBugs and CheckStyle maven configuration.
- Removed old FindBugs and CheckStyle parameters for old plugins.
- Enabled warnings-ng reporters for Maven, Java and JavaDoc.
- Enabled warnings-ng reporters for CheckStyle, PMD, CPD, and SpotBugs.
- Enabled warnings-ng reporters for Open Tasks.
- Add jar files to the archive as well (otherwise the [example job](https://ci.jenkins.io/job/Plugins/job/analysis-model/view/change-requests/job/PR-455/lastBuild/) will fail).


jtnord:
> I removed the maven configuration on how to start the tools

spotbugs is run as part of `verify` at least in semi recent versions of the plugin-pom

However it will fail the build and we expect an unstable build for testish (of which spotbugs is one) failures, so I think you still need to tell the build not to fail for spotbugs issues.

checkstyle is not as actually getting a consensus on any style is problematic :)

> I did not yet activate the quality gates.
> I did not yet activate the quality gates. I'm not sure if someone already is using that feature

There are at least some plugins that do.  Actually this is an issue I had recently with wanrings ng - the quality gate is acrross all tools not just specific ones that it used to be (which allows marking unstable for any spotbugs issue, but just reporting TODOs checkstyle etc).

https://github.com/search?p=2&q=org%3Ajenkinsci+findbugs+filename%3AJenkinsfile&type=Code  shows quite a few (on page 2 onwards)

zbynek:
@jtnord you can set a quality gates for a subset of tools: https://github.com/geogebra/geogebra/blob/5.0.570.0/Jenkinsfile#L44

uhafner:
Yes, indeed, you can specify a quality gate for each step, this was one of the main requirements for the warnings-ng redesign.

uhafner:
Does that mean it is ok to remove the `spotbugs:spotbugs` part but we need to leave the `failOnError=false` flags?

uhafner:
What still sounds for me as pipeline newbie kind of strange is that we need to remodel the  parameters of my steps as parameters for the `buildPlugin` method. As a plugin author I would feel more comfortable with some kind of closure that I am adding to the step (but that might not be possible with pipelines).

So in pseudo-code it would be nice to have:
```groovy
buildPlugin(
    post: () -> {
          recordIssues tools: [java(), javaDoc()], aggregatingResults: 'true', id: 'java', name: 'Java'
          publishCoverage adapters: [jacocoAdapter('target/site/jacoco/jacoco.xml')]
          ...
    }
)
```

jtnord:
> @jtnord you can set a quality gates for a subset of tools using two `recordIssues` steps: https://github.com/geogebra/geogebra/blob/5.0.570.0/Jenkinsfile#L44

but if you want an aggregated report... maybe we where doing something else wrong. I will try and reproduce and ask on the users list

> What still sounds for me as pipeline newbie kind of strange is that we need to remodel the parameters of my steps as parameters for the buildPlugin method

you can do that but it's the old way (not sure of all the reasons) but it can mean a plugin upgrade change could require reconfig of many jobs rather than a central place.
also does not help with making findings not fail the build and getting Jenkins to do it via warnings-ng.  (a rather contrived example would be if the unsupported findings plugin had a security vuln that would not be fixed as it's replaced by warnings-ng that meant we would uninstall that plugin)

jtnord:
> Does that mean it is ok to remove the `spotbugs:spotbugs` part but we need to leave the `failOnError=false` flags?

yes.  the failOnError should be the spotbugs flag not findings as that is about to be removed (or use both)

darxriggs:
The [buildPluginWithGradle](https://github.com/jenkins-infra/pipeline-library/blob/1.5.0/vars/buildPluginWithGradle.groovy) step should also be adapted.

uhafner:
Does someone know how to fix the unit test failure
```
No signature of method: buildPlugin.mavenConsole() is applicable for argument types: () values: []
```
Do I need to register my steps somewhere?

uhafner:
The step works perfectly in our CI build

v1v:
@uhafner regarding `buildPlugin.mavenConsole()` , you need to mock it with something like the below snippet:

```
class BuildPluginStepTests extends BasePipelineTest {
...
  void setUp() throws Exception {
...
    helper.registerAllowedMethod('checkStyle', [], { true })
    helper.registerAllowedMethod('java', [], { true })
    helper.registerAllowedMethod('javaDoc', [], { true })
    helper.registerAllowedMethod('mavenConsole', [], { true })
    helper.registerAllowedMethod('pmdParser', [], { true })
    helper.registerAllowedMethod('spotBugs', [], { true })
    helper.registerAllowedMethod('recordIssues', [Map.class], { true })
    helper.registerAllowedMethod('taskScanner', [Map.class], { true })
...

```

Cheers

vars/buildPlugin.groovy
```diff
@@ -142,29 +126,23 @@ def call(Map params = [:]) {
...
-                                if (params?.checkstyle?.unstableTotalAll) {
-                                    cp['unstableTotalAll'] = "${params.checkstyle.unstableTotalAll}"
-                                }
-                                checkstyle(cp)
-                            }
                             if (failFast && currentBuild.result == 'UNSTABLE') {
                                 error 'There were test failures; halting early'
                             }
+
+                            if (first) {
```
v1v: IIUC, the current condition will be always `true`, so I wonder if this condition is redundant somehow

uhafner: Isn't the condition only true for the first configuration we build? E.g. for linux it is true but for windowsit isfalse`?

v1v: Gotcha!! Thanks for the clarification, I actually have to look it at twice, thanks again :)

timja:
`JOB_NAME` is available for the `referenceJobName` field

timja:
@markewaite could you take a look at this please?

timja:
ping @jenkins-infra/ci-maintainers

vars/buildPlugin.groovy
```diff
@@ -152,29 +136,39 @@ def call(Map params = [:]) {
...
+
+                                recordIssues enabledForFailure: true,
+                                        tool: mavenConsole(),
+                                        referenceJobName: referenceJobName
+                                recordIssues enabledForFailure: true,
+                                        tools: [java(), javaDoc()],
+                                        sourceCodeEncoding: 'UTF-8',
+                                        filters:[excludeFile('.*Assert.java')],
+                                        referenceJobName: referenceJobName
+                                recordIssues tools: [spotBugs(pattern: '**/target/spotbugsXml.xml'),
```
oleg-nenashev, line 155: It would make sense to read `findbugsXml` as well

vars/buildPlugin.groovy
```diff
@@ -103,22 +99,10 @@ def call(Map params = [:]) {
...
                                 }
-                                if (params?.findbugs?.run || params?.findbugs?.archive) {
-                                    mavenOptions += "-Dfindbugs.failOnError=false"
-                                }
                                 if (skipTests) {
-                                    mavenOptions += "-DskipTests"
-                                }
-                                if (params?.checkstyle?.run || params?.checkstyle?.archive) {
-                                    mavenOptions += "-Dcheckstyle.failOnViolation=false -Dcheckstyle.failsOnError=false"
+                                    mavenOptions += "-DskipTests -DskipITs"
```
oleg-nenashev, line 103: `-DskipITs` is unrelated to the change subject. YAGNI?

vars/buildPlugin.groovy
```diff
@@ -103,22 +99,10 @@ def call(Map params = [:]) {
                                 if (javaLevel) {
                                     mavenOptions += "-Djava.level=${javaLevel}"
                                 }
-                                if (params?.findbugs?.run || params?.findbugs?.archive) {
-                                    mavenOptions += "-Dfindbugs.failOnError=false"
```
oleg-nenashev, line 107: `failOnError` is not longer added to the build. With the current job implementation the build will not really reach the publishing stage with warnings NG if there is a FindBugs issue

timja: confirmed in https://ci.jenkins.io/job/Plugins/job/slack-plugin/job/PR-707/1/

vars/buildPlugin.groovy
```diff
@@ -89,6 +85,7 @@ def call(Map params = [:]) {
                                         '--update-snapshots',
                                         "-Dmaven.repo.local=$m2repo",
                                         '-Dmaven.test.failure.ignore',
+                                        '-Dfindbugs.failOnError=false'
```
timja, line 88: https://github.com/jenkinsci/plugin-pom/blob/master/pom.xml#L577
```suggestion
                                        '-Dspotbugs.failOnError=false'
```

vars/buildPlugin.groovy
timja: Fixed that but now the build is green, it should be unstable instead:

https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fslack-plugin/detail/PR-707/4/pipeline

vars/buildPlugin.groovy
timja: I think a quality gate needs to be defined for spotbugs so that the build gets marked as unstable, currently it's just green which isn't good.

vars/buildPlugin.groovy
uhafner: Yes, this flag is from failsafe plugin, I'll remove it.

vars/buildPlugin.groovy
uhafner: I think the maven configuration of FindBugs/SpotBugs should not be done in the Jenkinsfile. This is something that should be part of the Plugin POM.

vars/buildPlugin.groovy
uhafner: Yes, I can add quality gates. But which ones should we use? How many new, how many total warnings?

vars/buildPlugin.groovy
timja: it's configured in the pom already:

https://github.com/jenkinsci/plugin-pom/blob/master/pom.xml#L68-L70

vars/buildPlugin.groovy
timja: 0 to both and make it configurable so it can be overridden?

vars/buildPlugin.groovy
uhafner: What is the reason that one should change the flag in the `Jenkinsfile` and not in the pom? 
Here is an example of a properly configured pom: https://github.com/jenkinsci/analysis-pom-plugin/blob/master/pom.xml#L415

timja: I believe it's so that builds will fail locally but on CI the full build completes so as much information as possible is presented to the developer to fix in one iteration.

uhafner: Hmm, I think this is the wrong way. This is typically something for a Maven profile.

vars/buildPlugin.groovy
uhafner: After thinking a little bit about the old style with a property I came to another idea: it might be more powerful if we let the developers decide on how to configure the whole step: 

https://github.com/jenkins-infra/pipeline-library/pull/121/commits/e4d86e8dcd43cc4e213a62b6af0696dbcd5080c8

What do you think about such an option? Then we can have meaningful defaults but the developers can override everything as they like.  Otherwise we would have a lot of duplicate properties that are part of the `buildPlugin` code and my plugin. 

If this is ok for you then I also add such a `map` for the other tools (checkstyle, etc.).

timja: Great idea and works for me

timja: Tested https://ci.jenkins.io/job/Plugins/job/slack-plugin/job/master/ and https://ci.jenkins.io/job/Plugins/job/slack-plugin/view/change-requests/job/PR-707/

Works 👍

timja:
Looks like it's just documentation updates blocking this in the README now.

And if you want to allow quality gate overrides for non spotbugs as well.

uhafner:
Yes, I am working on it - I will update the documentation if the syntax is ready.

uhafner:
> One quirk that I noticed is that depending on timing and when the parallel stages finish if a quality gate fails you may get a failed build or an unstable one

In the warnings plugin Jenkins file I am also halting the build with a proper error message: 
https://github.com/jenkinsci/warnings-ng-plugin/blob/master/Jenkinsfile#L131

Should I add it here as well?

timja:
> Should I add it here as well?
> 
> (It still can cause a problem if the check is too late, but at least in most cases the error message is correct)

That looks like it would fix the issue, yes I would suggest adding it here too

timja:
> It breaks compatibility for plugins which define findbugs settings in Jenkinsfile. Does not look to be optimal

won't they just be ignored now?

timja:
@uhafner the last commit doesn't seem to be working right, maybe it's not set to unstable in time?

https://ci.jenkins.io/job/Plugins/job/slack-plugin/view/change-requests/job/PR-707/

I ran two builds

uhafner:
Hmm, I did not test it yet. I just copied the test block. Did you activate `failFast`? I'll try in my plugins...

timja:
It's on by default: https://github.com/uhafner/pipeline-library/blob/warnings-ng/vars/buildPlugin.groovy#L17

uhafner:
Hmm, in my build it works: https://ci.jenkins.io/job/Plugins/job/analysis-model/job/env-job-name/11/console

timja:
> Hmm, in my build it works: [ci.jenkins.io/job/Plugins/job/analysis-model/job/env-job-name/11/console](https://ci.jenkins.io/job/Plugins/job/analysis-model/job/env-job-name/11/console)

My mistake, I pushed to the wrong remote, it works now: https://ci.jenkins.io/job/Plugins/job/slack-plugin/job/PR-707/14/

The only question left now is what do we do about existing users of the spotbugs / checkstyle configuration, do we want to map their configuration? ignore it? fail their builds and tell them to adapt it?

uhafner:
I think mapping is too complex as this also requires to adapt the maven call (which is now not altered anymore). Moreover, the old calls also use `findbugs` and not `spotbugs`.

slide:
Is there anything I can help with in getting this merged? It would be nice to have all the plugins be able to use buildPlugin directly.

oleg-nenashev:
Update from my side: I am working on
https://github.com/jenkinsci/ci.jenkins.io-runner update to the new
Jenkinsfile Runner version. Once done, I will try to add real integration
tests to this repository in addition to unit tests. Such change takes a
while, but I do not want to block this change.

I can prioritize review instead of test automation if it helps

timja:
Review would be great, there’s test builds of this setup

https://ci.jenkins.io/job/Plugins/job/slack-plugin/job/PR-707/14/

timja:
> There are many usages of the FindBugs and Checkstyle definitions:
> 
> * [github.com/search?q=org%3Ajenkinsci+buildPlugin+findbugs&type=Code](https://github.com/search?q=org%3Ajenkinsci+buildPlugin+findbugs&type=Code)
> * [github.com/search?q=org%3Ajenkinsci+buildPlugin+checkstyle&type=Code](https://github.com/search?q=org%3Ajenkinsci+buildPlugin+checkstyle&type=Code)
> 
> I am not 100% comfortable about merging it without a compatibility layer. An alternative to a compatibility layer would be to have a new `buildPluginWithMaven()` with new arguments (and removed gradle support, yes)
> 
> I could probably help with one of these options on the weekend

The majority of those look like defaults or very close to it,

`buildPluginWithMaven` is way to overkill, for like ~10-20 repos who use these flags...

imo for compatibility as long as providing checkstyle configuration with the old config doesn't break it should be fine.

another option is to just send PRs as there isn't many...

slide:
I think buildPlugin is becoming overly complex as it is, adding a compatibility layer will only continue to increase the complexity. I think making PR's to the plugins that need it would be a good solution.

vars/buildPlugin.groovy
```diff
@@ -103,22 +101,10 @@ def call(Map params = [:]) {
...
                                     mavenOptions += "-Djava.level=${javaLevel}"
                                 }
-                                if (params?.findbugs?.run || params?.findbugs?.archive) {
-                                    mavenOptions += "-Dfindbugs.failOnError=false"
-                                }
                                 if (skipTests) {
                                     mavenOptions += "-DskipTests"
                                 }
-                                if (params?.checkstyle?.run || params?.checkstyle?.archive) {
-                                    mavenOptions += "-Dcheckstyle.failOnViolation=false -Dcheckstyle.failsOnError=false"
```
timja, line 113: should we be setting these flags in the maven run?

vars/buildPlugin.groovy
```diff
@@ -89,6 +86,7 @@ def call(Map params = [:]) {
                                         '--update-snapshots',
                                         "-Dmaven.repo.local=$m2repo",
                                         '-Dmaven.test.failure.ignore',
+                                        '-Dspotbugs.failOnError=false'
```
timja, line 89: should we added the checkstyle vars here so the build doesn't fail but gets marked as unstable?
```suggestion
                                        '-Dspotbugs.failOnError=false',
                                        '-Dcheckstyle.failOnViolation=false',
                                        '-Dcheckstyle.failsOnError=false'
```

timja:
I've added the checkstyle fail on violation / warning vars by default and introduced a default quality gate for it.

I can't see any harm in including it and I think we should have sensible defaults and not ask maintainers to have to include anything in order to have the build failed, they can override it if they want, thoughts?

timja:
@oleg-nenashev @slide @uhafner I've opened PRs to every single plugin that uses the findbugs or checkstyle configurations.

are we good to go ahead?

slide:
@oleg-nenashev goes above and beyond again! I'm happy to move forward

vars/buildPlugin.groovy
```diff
@@ -152,28 +140,79 @@ def call(Map params = [:]) {
...
+                            if (first) {
+                                referenceJobName = env.JOB_NAME.substring(0, env.JOB_NAME.lastIndexOf("/") + 1) + "master"
+                                echo "Recording static analysis results on '${stageIdentifier}' using reference job '${referenceJobName}'"
+
+                                recordIssues enabledForFailure: true,
+                                        tool: mavenConsole(),
+                                        trendChartType: 'TOOLS_ONLY',
+                                        referenceJobName: referenceJobName
+                                recordIssues enabledForFailure: true,
+                                        tools: [java(), javaDoc()],
```
jglick, line 156: With the addition of Checks, we are suddenly getting tons and tons of warnings. For example, https://github.com/jenkinsci/ldap-plugin/pull/48/files is practically illegible because of irrelevant, low-priority warnings like

```
NORMAL:
no @param for rootDN
```

Worse, `checks-api` apparently submits annotations even for lines of code which are not modified by a PR and GH displays every annotation, so the **Files** tab is lost in annotations which are not even related to the diff.

Can we at least turn off display of low-priority Javadoc warnings?

jglick: Hmm, do not seem to be getting these Checks results in PRs on other plugins. Not sure what would be different here.

timja: You could work around this by doing one of:

* fixing the javadoc issues
* adjusting the maven javadoc plugin not to complain about these
* extending the javadoc / java parts of buildPlugin to allow overriding the configuration like you can do for spotbugs, checkstyle, pmd already and then do something like `buildPlugin(javadoc: skipPublishingChecks: true`) in your plugin
* turning checks off by default for javadoc

thoughts?

cc @XiongKezhi

jglick: > adjusting the maven javadoc plugin not to complain about these

Could be done in `plugin-pom` I guess. I care about genuine problems—`@link` target missing, mistyped tag, etc.—but not just silliness like missing `@param` which is true for _most_ methods in the Jenkins source base.

XiongKezhi: Those warnings are also reported by the warnings-ng plugin: https://ci.jenkins.io/job/Plugins/job/ldap-plugin/job/PR-48/3/, they are just scanned from Java or JavaDoc log files,  we can't really do much about the scan and collect process. But you can control the minimal severity of issues to report:
```
recordIssues tool: java(pattern: '*.log'), healthy: 10, unhealthy: 100, minimumSeverity: 'HIGH'
```
But this is the pipeline grammar of warnings-ng, I'm not sure how to override the configurations in `buildPlugin`, maybe 
`buildPlugin(java: minimumSeverity: 'LOW')`? @timja 

Also, do you think we should provide another configuration for minimal severity beside the warnings-ng one to only control the published checks?

timja: I don’t think so, just need to configure warnings ng / the Pom appropriately

jglick: see #154

PR #124, opened 2019-12-23
Activate the benchmark profile directly rather than indirectly via a property

jtnord:
do not be scared of profiles - activate them directly so you know what is happening (also removed in plugin-pom 4.0)
https://github.com/jenkinsci/plugin-pom/pull/269

### DIFF
--- a/vars/runBenchmarks.groovy
+++ b/vars/runBenchmarks.groovy
@@ -14,7 +14,7 @@ def call(String artifacts = null) {
             }
 
             stage('Run Benchmarks') {
-                List<String> mvnOptions = ['test', '-Dbenchmark']
+                List<String> mvnOptions = ['test', '-P jmh-benchmark']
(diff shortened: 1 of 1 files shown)


vars/runBenchmarks.groovy
```diff
@@ -14,7 +14,7 @@ def call(String artifacts = null) {
             }
 
             stage('Run Benchmarks') {
-                List<String> mvnOptions = ['test', '-Dbenchmark']
+                List<String> mvnOptions = ['test', '-P jmh-benchmark']
```
oleg-nenashev: will it work with a space?

jtnord: yes.

jtnord: I would expect it is all going down via bash so likely yes - but I have fixed it as it should be a separate arg :)

PR #126, opened 2019-12-24
Skip test result collection if tests are skipped

darxriggs:
It was missed in commit 14ca5fe16ff95d8c4f39f05015c93addee49b0ce.


darxriggs:
I will also create another pull request to add tests for `buildPluginWithGradle` when this one is merged.

darxriggs:
@oleg-nenashev could you already release this and use it on ci.jenkins.io?

I am asking because while migrating Gradle-based plugins from `buildPlugin` to `buildPluginWithGradle`, it shows that some do not have tests at all.

PR #129, opened 2019-12-30
Use explicit types for Custom WAR Packager settings

darxriggs:
This better describes the settings.
In case of `jdk` this also improves the user experience.
If an `Integer` is provided, it is converted to a `String`.
This is required for the `runMaven(...)` call to work correctly.


darxriggs:
This change is also required for another upcoming pull request to simplify the `Infra` mock for testing. The goal there is to use the same types in the method signatures as in the real implementation.

oleg-nenashev:
Restarted the build

PR #130, opened 2019-12-31
Improve Infra mock in PiplineUnit

darxriggs:
Use same types in method signatures as in the real implementation.
Use default maps constructor to get named parameters.

It requires pull request #129 to be merged beforehand.


darxriggs:
@v1v I am not sure if @oleg-nenashev is waiting for your reply. So I think it would be a good idea that you provide a quick reply.

I am waiting for this to be merged because I have another pull request in preparation that depends on it.

oleg-nenashev:
Merging it since to response from @v1v

PR #135, opened 2020-01-27
[INFRA-2435] Remove repo.azure mirrors

olblak:
`repo.azure.jenkins.io` is deprecated and we are reverting it to repo.jenkins-ci.org
More information [here](https://issues.jenkins-ci.org/browse/INFRA-2435?focusedCommentId=383771&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-383771)

### DIFF
--- a/resources/settings-azure.xml
+++ b/resources/settings-azure.xml
@@ -42,17 +42,4 @@
             </properties>
         </profile>
     </profiles>
-
-    <mirrors>
-        <mirror>
-            <id>azure</id>
-            <url>https://repo.azure.jenkins.io/public/</url> <!-- INFRA-1176 -->
-            <mirrorOf>repo.jenkins-ci.org</mirrorOf>
-        </mirror>
-        <mirror>
-            <id>azure-incrementals</id>
-            <url>https://repo.azure.jenkins.io/incrementals/</url> <!-- IEP-9, JEP-305 -->
-            <mirrorOf>incrementals</mirrorOf>
-        </mirror>
-    </mirrors>
 </settings>


olblak:
I made some tests from ci.jenkins.io, where the bottom terminal is repo.azure.jenkins-ci.org
If you want to take more time to discuss about it, then keep in mind that the cluster will be gone in two days.

So either we change every occurrence of repo.azure.jenkins.io to repo.jenkins-ci.org or we redeploy repo.azure.jenkins.io on the new kubernetes cluster.


image

oleg-nenashev:
I am fine with that as a last resort, but we may need to find a solution quickly after Jan 31

olblak:
I am merging this to see how it behaves

jglick:
This PR does not make sense to me. The only real purpose for `settings-azure.xml` is to define this mirror. If we are getting rid of the mirror, we should delete the whole file, and the usages of it.

jglick:
(All Jenkins POMs already define the Artifactory repository locations, so you do not need a special settings file for this; the Incrementals repository is already activated conditionally in a profile, so it is less than worthless to define it here; and the log level trick is already done by `mvn` command lines, and anyway obsolete as of Maven 3.6.1’s `-ntp`.)

oleg-nenashev:
> This PR does not make sense to me. ... If we are getting rid of the mirror, we should delete the whole file, and the usages of it.

I believe the approach is right. If we ever restore the mirror or introduce other shared profiles for ci.jenkins.io, this file will be useful. So +100 for keeping the stub and requiring using it in ci.jenkins.io runs. Also, usage cleanup requires time which could be better spent elsewhere

jglick:
> If we ever restore the mirror or introduce other shared profiles for ci.jenkins.io, this file will be useful.

Perhaps, but then we should delete all sections from it and replace its body with

```xml
<!-- Stub in case we ever wish to reintroduce mirrors (see INFRA-2435) -->
```

PR #141, opened 2020-04-11
Clean workspace before building incrementals

MarkEWaite:
## Clean workspace before building incrementals

The incrementals build process will abort if `git status -s` is not empty.  Since the maven step will invoke `mvn clean`, a `git clean` is safe.  In case command line git is not installed, the script does its very best to not fail with an error and to report that it intentionally did not fail with an error.

Also includes a spelling fix.

This change is needed until ci.jenkins.io agents become 100% ephemeral.  I've been hit repeatedly by tainted agent workspaces with the git plugin.  This will reduce the failure rate of git plugin builds and git client plugin builds on ci.jenkins.io.

See [my git plugin pull request](https://github.com/jenkinsci/git-plugin/pull/870) for an example of this pull request being used in a repository.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -67,11 +67,11 @@ def call(Map params = [:]) {
                                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
                             if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
                                 if (isUnix()) {
-                                    sh(script: 'git clean -xffd || true',
+                                    sh(script: 'git clean -xffd > /dev/null 2>&1',
                                        label:'Clean for incrementals',
                                        returnStatus: true) // Ignore failure if CLI git is not available
                                 } else {
-                                    bat(script: 'git clean -xffd || echo "git clean failure intentionally ignored in incrementals setup"',
+                                    bat(script: 'git clean -xffd 1> null 2>&1',
                                         label:'Clean for incrementals',
                                         returnStatus: true) // Ignore failure if CLI git is not available
                                 }
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -66,6 +66,17 @@ def call(Map params = [:]) {
                             isMaven = fileExists('pom.xml')
                             incrementals = fileExists('.mvn/extensions.xml') &&
                                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
+                            if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
+                                if (isUnix()) {
+                                    sh(script: 'git clean -xffd || true',
+                                       label:'Clean for incrementals',
+                                       returnStatus: true) // Ignore failure if CLI git is not available
+                                } else {
+                                    bat(script: 'git clean -xffd || echo "git clean failure intentionally ignored in incrementals setup"',
+                                        label:'Clean for incrementals',
+                                        returnStatus: true) // Ignore failure if CLI git is not available
+                                }
+                            }
                         }
 
                         String changelistF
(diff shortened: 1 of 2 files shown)


bitwiseman:
@jenkins-infra/ci-maintainers  Can we merge this?

vars/buildPlugin.groovy
```diff
@@ -66,6 +66,17 @@ def call(Map params = [:]) {
                             isMaven = fileExists('pom.xml')
                             incrementals = fileExists('.mvn/extensions.xml') &&
                                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
+                            if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
```
jglick, line 69: Could we not just do this unconditionally?

vars/buildPlugin.groovy
```diff
@@ -66,6 +66,17 @@ def call(Map params = [:]) {
...
+                                if (isUnix()) {
+                                    sh(script: 'git clean -xffd || true',
+                                       label:'Clean for incrementals',
+                                       returnStatus: true) // Ignore failure if CLI git is not available
```
jglick, line 73: Does the `|| true` not sufice?

vars/buildPlugin.groovy
MarkEWaite: We certainly could, but I wanted to keep the impact of the change to only those specific cases where I know it is an issue.  I'm willing to change the pull request to do it in all cases if that will help with its approval.

vars/buildPlugin.groovy
MarkEWaite: It should, but I was practicing an "abundance of caution" in hopes that a peripheral cleanup operation would have even less chance of inadvertently breaking a build.

jglick: So then `returnStatus: true` should suffice (both both platforms)?

MarkEWaite: Yes, should. Same "abundance of caution" logic.  If ignoring the return value inside the script somehow did not work, also take the active measure of ignoring the return value from the step.  Both [`sh`](https://www.jenkins.io/doc/pipeline/steps/workflow-durable-task-step/#sh-shell-script)` and [`bat`](https://www.jenkins.io/doc/pipeline/steps/workflow-durable-task-step/#bat-windows-batch-script) step documentation say:

> Normally, a script which exits with a nonzero status code will cause the step to fail with an exception.  If this option is checked, the return value of the step will instead be the status code.

MarkEWaite: I've changed to hide the output from git clean and ignore the return code in c05194b38dddced80f031ed365ca9274950ab7dc .  Let me know if other changes are needed.

MarkEWaite:
> Fine if these commands have been tested and seem to work. Currently there is no mechanism for testing `buildPlugin` AFAIK.

I tested them by running them on my home CI server.  The CI tests at home (SCM API, Branch API, Git plugin, Git client plugin, Platform labeler plugin) detected the mistake that I had made in c05194b38dddced80f031ed365ca9274950ab7dc of using the wrong name for the nul device on Windows.  That is corrected in 09edbc1a56992e0e355c3837e8f6c2e23a36ae02.  It was quite awkward to have a file named `null` on Windows causing 100% failure of all jobs that used incrementals.

bitwiseman:
@jenkins-infra/ci-maintainers 
Can we merge this now?  My CI jobs continue to fail due to files left in the workspace.

MarkEWaite:
> @jenkins-infra/ci-maintainers
> Can we merge this now? My CI jobs continue to fail due to files left in the workspace.

I had to resolve a conflict.  I'm not sure that I chose the correct direction for the resolution of the conflict.  We'll know when the tests run.

jglick:
> I'm not sure that I chose the correct direction for the resolution of the conflict.

Pro tip (cannot fathom why this is not the default): add to `~/.gitconfig`

```ini
[merge]
	conflictstyle = diff3
```

MarkEWaite:
I continue to encounter this problem in plugin builds that use incrementals.  Is there anything else blocking it from being included in the pipeline shared library?

timja:
@markewaite imo you should request to join that team, you’re doing work on ci.jenkins.io regularly

PR #144, opened 2020-04-30
Update ACI usage

slide:
This will enable Windows ACI usage when `useAci: true` is used. It will also only enable ACI when the label matches linux or windows so that if a different platform is specified (arm64, ppc64le, etc), it will work correctly.

/cc @MarkEWaite

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -15,8 +15,8 @@ def call(Map params = [:]) {
     def repo = params.containsKey('repo') ? params.repo : null
     def failFast = params.containsKey('failFast') ? params.failFast : true
     def timeoutValue = params.containsKey('timeout') ? params.timeout : 60
-    def useAci = params.containsKey('useAci') ? params.useAci : false
     def forceAci = params.containsKey('forceAci') ? params.forceAci : false
+    def useAci = params.containsKey('useAci') ? params.useAci : forceAci
     if(timeoutValue > 180) {
       echo "Timeout value requested was $timeoutValue, lowering to 180 to avoid Jenkins project's resource abusive consumption"
       timeoutValue = 180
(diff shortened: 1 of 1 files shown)


dwnusbaum:
I think that the Git installation is somewhat broken on the Windows ACI agents, see the build history on https://github.com/jenkinsci/workflow-cps-plugin/pull/355. Tests that use the Git plugin internally fail with this error:

> C:\mingit\usr\bin\sh.exe: *** fatal error - unable to load netapi32.dll, Win32 error 126

Here is the stack trace:

```
hudson.plugins.git.GitException: Command "git.exe fetch --tags --progress C:\Users\jenkins\Work\workspace\ugins_workflow-cps-plugin_PR-355\target\tmp\junit8163935707156005829\junit5722207468178687480 +refs/heads/*:refs/remotes/origin/*" returned status code 128:
 ... above error message ...
 at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:1793)
 at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandWithCredentials(CliGitAPIImpl.java:1519)
 at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.access$300(CliGitAPIImpl.java:64)
 at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$1.execute(CliGitAPIImpl.java:315)
 at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$2.execute(CliGitAPIImpl.java:512)
 at hudson.plugins.git.GitSCM.retrieveChanges(GitSCM.java:1061)
 at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1101)
 at org.jenkinsci.plugins.workflow.steps.scm.SCMStep.checkout(SCMStep.java:109)
 at org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition.create(CpsScmFlowDefinition.java:155)
 at org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition.create(CpsScmFlowDefinition.java:69)
 at org.jenkinsci.plugins.workflow.job.WorkflowRun.run(WorkflowRun.java:293)
 at hudson.model.ResourceController.execute(ResourceController.java:97)
 at hudson.model.Executor.run(Executor.java:428)
```

As a workaround, I can switch my builds back to using non-ACI agents, but I'd prefer not to do that for the Linux builds if possible. Any idea on what would be needed to fix this? Would it make sense to revert this PR temporarily, or maybe to expose a separate `useAciWindows` option that can be disabled without also disabling ACI agents on Linux?

slide:
I will look into this ASAP

PR #145, opened 2020-05-11
Deprecate recommendedConfigurations

jglick:
Amends #92 (which broke some builds) and supersedes #94. Better to specify configurations you want in your `Jenkinsfile`, and adjust those in PRs.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -51,15 +51,6 @@ buildPlugin(/*...*/, configurations: [
 ])
 ----
 
-** It is also possible to use a `buildPlugin.recommendedConfigurations()` method to get recommended configurations for testing.
-Note that the recommended configuration may change over time,
-and hence your CI may break if the new recommended configuration is not compatible
-+
-[source,groovy]
-----
-buildPlugin(/*...*/, configurations: buildPlugin.recommendedConfigurations())
-----
-
 * `tests`: (default: `null`) - a map of parameters to run tests during the build
 ** `skip` - If `true`, skip all the tests by setting the `-skipTests` profile.
   It will also skip FindBugs in modern Plugin POMs.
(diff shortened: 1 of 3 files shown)


jglick:
See discussion in https://github.com/jenkins-infra/pipeline-library/pull/94#issuecomment-522657761 and elsewhere.

olblak (dismissed):

jglick:
Updated to subsume #146.

oleg-nenashev (dismissed):
I disagree with removing it. The whole reason of `buildPlugin.recommendedConfigurations()` was to remove the need to manually manage versions for plugin developers. Definitely the suggested approach did not get well accepted once the version was updated, but the behavior is documented explicitly.

We need better reporting to ensure that the Pipeline acts as an advisory tool without failing the build for recent versions, but IMO the feature itself makes sense and should be retained

jglick:
> The whole reason of `buildPlugin.recommendedConfigurations()` was to remove the need to manually manage versions for plugin developers.

Yes; and I am saying this experiment failed. By contrast, Dependabot on `pom.xml` is (by and large) working. We could try to find a way to make Dependabot offer `jenkins.version` bumps, if we wanted to guide plugins to just build against one recent LTS version (critical bug fixes could be backported on demand).

> better reporting to ensure that the Pipeline acts as an advisory tool without failing the build for recent versions

I do not see how that could work. Either the build passes or it fails; developers need to be able to tell whether their changes are broken or not. If you want to run PCT on `master` of a few dozen popular plugins in another job and have someone trawling through failures, fine, but that would not be part of `*-plugin/Jenkinsfile` (that is actually pretty close to what `bom` builds are doing).

jglick:
Continues to be a problem in various plugins, and people who know continue to remove it.

timja (approved):

basil (approved):
+`UINT64_MAX`

PR #148, opened 2020-05-26
JDK version parameters in `buildPlugin()` are supposed to be strings not integers

jglick:
Noticed after https://github.com/jenkinsci/workflow-job-plugin/pull/165#discussion_r430589682

Perhaps only matters in `useAci` mode: https://github.com/jenkins-infra/pipeline-library/blob/7d45d79ec5ef1ee15558568e012f130faff91456/vars/buildPlugin.groovy#L44


jglick:
also note apparent implicit coercion of param defaults: https://github.com/jenkins-infra/pipeline-library/blob/7d45d79ec5ef1ee15558568e012f130faff91456/vars/infra.groovy#L69-L99

PR #149, opened 2020-06-09
Add step to push test results to Codecov

stellargo:
CC @oleg-nenashev


stellargo:
If this change is not needed here, I am happy to remove it. If it looks good, Ill file a JIRA and document it

oleg-nenashev:
I may have misunderstood the discussion in Jira, but IIRc the intent was to publsh JaCoCo coverage reports. We can indeed deploy it to a service, but we could also publish the reports in Jenkins directly. have you considered using https://plugins.jenkins.io/jacoco/ for publishing reports right inside Jenkins?

oleg-nenashev:
Or maybe even https://plugins.jenkins.io/code-coverage-api/

stellargo:
Ok I will use Jacoco plugin. I just wanted those PR checks, but no issues this works too. Thanks :) Ill close the PR.

PR #150, opened 2020-06-11
Add Jacoco step in buildPlugin()

stellargo:
CC @oleg-nenashev

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -37,6 +37,7 @@ def call(Map params = [:]) {
         boolean runCheckstyle = first && params?.checkstyle?.run
         boolean archiveFindbugs = first && params?.findbugs?.archive
         boolean archiveCheckstyle = first && params?.checkstyle?.archive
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
maybe it should be opt-in, similar to FindBugs & Co?

vars/buildPlugin.groovy
```diff
@@ -172,6 +173,9 @@ def call(Map params = [:]) {
                                 }
                                 checkstyle(cp)
                             }
+                            if (isMaven && archiveJacoco) {
```
timja, line 176: does warnings-ng cover this or do both need to be used?

timja: shouldn't this be the code coverage api plugin?
https://plugins.jenkins.io/code-coverage-api/

uhafner: See #153

timja:
replaced by https://github.com/jenkins-infra/pipeline-library/pull/153

PR #151, opened 2020-07-20
Only use ACI agents on Windows when forceAci is true

dwnusbaum:
@slide and @timja and I were recently chatting in `#jenkins-infra` and they mentioned that `useAci` is not supposed to affect Windows agents, you are expected to use `forceAci` if you want ACI agents on Windows (also discussed in [the documentation here](https://github.com/jenkins-infra/pipeline-library#optional-arguments)), but from a quick look at `buildPlugin` I think that `forceAci` doesn't actually do anything right now other than set the default value of `useAci`, and `useAci` _does_ affect Windows.

Maybe I am missing something and `forceAci` does affect platform-specific behavior (for example I do not know the actual labels for ACI and non-ACI agents on ci.jenkins.io), but I figured I'd file this just in case.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -40,12 +40,10 @@ def call(Map params = [:]) {
         boolean skipTests = params?.tests?.skip
         boolean addToolEnv = !useAci
 
-        if(useAci && (label == 'linux' || label == 'windows')) {
-            String aciLabel = jdk == '8' ? 'maven' : 'maven-11'
-            if(label == 'windows') {
-                aciLabel += "-windows"
-            }
-            label = aciLabel
+        if (useAci && label == 'linux') {
+            label = jdk == '8' ? 'maven' : 'maven-11'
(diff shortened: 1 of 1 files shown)


dwnusbaum:
Also though, if `useAci` really has been affecting all platforms for some time now, maybe the better solution is to leave the behavior as-is, update the docs, and deprecate `forceAci`?

slide:
I couldn't remember if we had changed to use useAci, or not. The original code is what is correct. We left forceAci because some people were using it. The existing code is correct, so I don't think this PR is needed.

slide:
Yes, we should deprecate forceAci and update docs.

jglick:
> update docs

https://github.com/jenkins-infra/pipeline-library/search?q=forceaci&unscoped_q=forceaci specifically

slide:
See #152

dwnusbaum:
Closing, see #152.

Qadeem411 (approved):

PR #152, opened 2020-07-20
Remove forceAci - all usages updated

slide:


dwnusbaum:
It looks like the only uses of `forceAci` are in `email-ext` and `azure-keyvault`, although various plugins have copied the contents of `buildPlugin.groovy` into their Jenkinsfile, see [this search](https://github.com/search?utf8=%E2%9C%93&q=user%3Ajenkinsci+forceAci&type=Code). If we update those repositories we can just delete the parameter entirely.

slide:
> It looks like the only uses of `forceAci` are in `email-ext` and `azure-keyvault`, although various plugins have copied the contents of `buildPlugin.groovy` into their Jenkinsfile, see [this search](https://github.com/search?utf8=%E2%9C%93&q=user%3Ajenkinsci+forceAci&type=Code). If we update those repositories we can just delete the parameter and its documentation entirely.

I can update email-ext easily (I maintain that one). I'll file a PR against azure-keyvault as well.

slide:
email-ext is updated (https://github.com/jenkinsci/email-ext-plugin/commit/a5757a69b3cecf467b9ae94a6935adee4d6c4cb1), PR for azure-keyvault submitted (https://github.com/jenkinsci/azure-keyvault-plugin/pull/62)

timja:
Merged azure-keyvault, I think we were using it for testing windows ci 🤷 
or an ACI fix =/

slide:
I'm submitting PR's to the projects that copied the contents of buildPlugin. It's mostly the warnings-ng and it's dependencies. I need to compare what they have in buildPlugin vs what we have in the global library to see what they are needing so we can move them to use buildPlugin from the global library.

timja:
It's this PR https://github.com/jenkins-infra/pipeline-library/pull/121 they need

README.adoc
```diff
@@ -29,7 +29,7 @@ buildPlugin()
 
 * `repo` (default: `null`  inherit from Multibranch) - custom Git repository to check out
 * `useAci` (default: `false`) - uses link:https://azure.microsoft.com/en-us/services/container-instances/[Azure Container Instances] for build instead of link:https://azure.microsoft.com/en-us/services/virtual-machines/[Azure Virtual Machines].
-* `forceAci` (default: `false`) - forces usage of ACI (see `useAci` above) even on platforms that are not enabled when `useAci` is set to `true` (Windows).
+* `forceAci` (default: `false`) - this parameter is deprecated, please use `useAci` instead.
```
timja, line 32: why not just remove it?

```suggestion
```

slide: It is still in the buildPlugin.groovy script, so I thought it was best to actually notate that it is deprecated instead of just removing it.

dwnusbaum: Now that no plugins are using it though, you could remove it from `buildPlugin.groovy` as well.

slide:
Good point...

Qadeem411 (approved):

PR #153, opened 2020-07-29
Add optional recording of code coverage

uhafner:
Add a new configuration option `withCoverage` to enable the computation and recording of code coverage results using [JaCoCo maven plugin](https://www.eclemma.org/jacoco/trunk/doc/maven.html) and [Jenkins Code Coverage API plugin](https://github.com/jenkinsci/code-coverage-api-plugin).

See https://ci.jenkins.io/job/Plugins/job/analysis-model/job/PR-484/3/ for an example.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -32,7 +32,6 @@ def call(Map params = [:]) {
         String stageIdentifier = "${label}-${jdk}${jenkinsVersion ? '-' + jenkinsVersion : ''}"
         boolean first = tasks.size() == 1
         boolean skipTests = params?.tests?.skip
-        boolean enableCoverage = params?.tests?.withCoverage
         boolean addToolEnv = !useAci
 
         if(useAci && (label == 'linux' || label == 'windows')) {
@@ -106,12 +105,11 @@ def call(Map params = [:]) {
                                 }
                                 mavenOptions += "clean install"
                                 try {
-                                    echo "Running maven with options " + mavenOptions
                                     infra.runMaven(mavenOptions, jdk, null, null, addToolEnv)
                                 } finally {
                                     if (!skipTests) {
                                         junit('**/target/surefire-reports/**/*.xml,**/target/failsafe-reports/**/*.xml,**/target/invoker-reports/**/*.xml')
-                                        if (first && enableCoverage) {
+                                        if (first) {
                                             publishCoverage calculateDiffForChangeRequests: true, adapters: [jacocoAdapter('**/target/site/jacoco/jacoco.xml')]
                                         }
                                     }
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -1,6 +1,4 @@
-#!/usr/bin/env groovy
-import org.codehaus.groovy.runtime.DefaultGroovyMethods
-
+#!/usr/bin/env groovy 
 //TODO(oleg_nenashev): This thing is not simple anymore. I suggest reworking it to a config YAML
 // which would be compatible with essentials.yml (INFRA-1673)
 /**
@@ -107,9 +105,6 @@ def call(Map params = [:]) {
                                     mavenOptions += "-DskipTests"
                                 }
                                 mavenOptions += "clean install"
-                                if (!skipTests && first && enableCoverage) {
-                                    mavenOptions += "jacoco:prepare-agent test jacoco:report"
-                                }
                                 try {
                                     echo "Running maven with options " + mavenOptions
                                     infra.runMaven(mavenOptions, jdk, null, null, addToolEnv)
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -34,6 +34,7 @@ def call(Map params = [:]) {
         String stageIdentifier = "${label}-${jdk}${jenkinsVersion ? '-' + jenkinsVersion : ''}"
         boolean first = tasks.size() == 1
         boolean skipTests = params?.tests?.skip
+        boolean enableCoverage = params?.tests?.withCoverage
         boolean addToolEnv = !useAci
 
         if(useAci && (label == 'linux' || label == 'windows')) {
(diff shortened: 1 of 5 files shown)


uhafner:
Replaces #150.

vars/buildPlugin.groovy
```diff
@@ -106,11 +107,18 @@ def call(Map params = [:]) {
                                     mavenOptions += "-DskipTests"
                                 }
                                 mavenOptions += "clean install"
+                                if (!skipTests && first && enableCoverage) {
+                                    mavenOptions += "jacoco:prepare-agent test jacoco:report"
```
timja, line 111: shouldn't this be done automatically as part of the maven build, using the recommended configuration in the code-coverage-api? 

(skip to jacoco, no anchor unfortunately)
https://github.com/jenkinsci/code-coverage-api-plugin#1-config-coverage-tool-to-generate-reports

vars/buildPlugin.groovy
```diff
@@ -106,11 +107,18 @@ def call(Map params = [:]) {
...
+                                }
                                 try {
+                                    echo "Running maven with options " + mavenOptions
                                     infra.runMaven(mavenOptions, jdk, null, null, addToolEnv)
                                 } finally {
                                     if (!skipTests) {
                                         junit('**/target/surefire-reports/**/*.xml,**/target/failsafe-reports/**/*.xml,**/target/invoker-reports/**/*.xml')
+                                        if (first && enableCoverage) {
+                                            publishCoverage adapters: [jacocoAdapter('target/site/jacoco/jacoco.xml')]
```
timja, line 120: should we enable the `calculateDiffForChangeRequests: true` ?

https://www.jenkins.io/blog/2019/07/21/Jenkins-code-coverage-diff-for-pull-request/#how-to-enable-code-coverage-diff-for-pull-requests

timja: what happens if this is a multi module project?

does the path need to be configurable or should we locate all the jacoco reports?

or YAGNI and enhance later?

vars/buildPlugin.groovy
uhafner: Yes, this would be another way to do it. But how do we toggle the execution of Jacoco then? Should we assume that we always record the coverage and remove all toggles?

vars/buildPlugin.groovy
uhafner: Yes, adding the diff would make sense, yes. I never used that before but it sounds quite useful.

vars/buildPlugin.groovy
timja: is there any reason not to?

uhafner: Normally, I run test twice: once without coverage in order to make sure that the unpatched class files are tested, and a second time with coverage. This duplicates the run time for the tests. But maybe this approach is too academic since in practice I never noticed tests that failed only due the coverage modifications off the class files.

timja: I come from a gradle world where everything like this is just bound to the `check` phase.

It seems odd to me to have to directly invoke plugins, I don't really understand your answer on why you run it without first.

If it adds quite a bit of time to the build I would understand it, although I tried it out on my plugins and it seemed fine.

uhafner: > I don't really understand your answer on why you run it without first.

Because the coverage step changes the class files. I.e. the tests may theoretically fail (or succeed) because of the changed class files. Normally we should test the same executable that we are shipping. (As noted, this might be a too academical approach.(

uhafner: Do we already have a helper method that detects if files of a given pattern exist?

timja: Not that I’m aware of

timja:
LGTM pending testing it out on ci.jenkins.io

timja:
I've pushed it as an origin branch to `code-coverage`

Anyone can test this out by adding:

```
@Library('pipeline-library@code-coverage')
```

to their Jenkinsfile.

I'm interested in:

* a pipeline without code coverage
* one with code coverage
* any pom changes required to make it work

uhafner:
I added the following snippet to my parent pom to get jacoco reporting the results:
https://github.com/uhafner/codingstyle/blob/master/pom.xml#L498

We should adapt our parent pom accordingly.

uhafner:
> I've pushed it as an origin branch to `code-coverage`
> 
> Anyone can test this out by adding:
> 
> ```
> @Library('pipeline-library@code-coverage')
> ```
> 
> to their Jenkinsfile.

Do you have a whole example that is working? I tried https://github.com/jenkinsci/analysis-model/pull/495/files but this seems not to compile.

timja:
https://github.com/jenkinsci/analysis-model/pull/495/files#r475479363

uhafner:
Here a first example for a Java library: 
- [PR](https://github.com/jenkinsci/analysis-model/pull/495)
- [GitHub Checks](https://github.com/jenkinsci/analysis-model/pull/495/checks?check_run_id=1021082244)
- [Jenkins CI](https://ci.jenkins.io/job/Plugins/job/analysis-model/job/PR-495/2/)

uhafner:
And here an example of a build that does not produce the coverage results:
https://ci.jenkins.io/job/Plugins/job/forensics-api-plugin/job/serialization-0.7/1/console

```
15:01:00  Publishing Coverage report....
15:01:01  No reports were found
```

uhafner:
BTW: currently the source code visualization is not enabled. Should I enable it for the latest build? (Requires some file system space.)

timja:
> BTW: currently the source code visualization is not enabled. Should I enable it for the latest build? (Requires some file system space.)

May as well enable it

uhafner:
Here is a PR that builds a plugin with coverage: https://github.com/jenkinsci/forensics-api-plugin/pull/203.

MarkEWaite:
> > BTW: currently the source code visualization is not enabled. Should I enable it for the latest build? (Requires some file system space.)
> 
> May as well enable it

I'm `-1` for enabling it at this time because we're 80% full  on the disc that holds the JENKINS_HOME file system on ci.jenkins.io.  Before we enable something that significantly increases disc use on JENKINS_HOME of ci.jenkins.io, we either need to reduce current disc use or add more disc space.

uhafner:
It will not work anyway: https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java#L282

The coverage result checks for workspace permission which I do not have.

timja (approved):

timja:
@jenkins-infra/ci-maintainers 🚢 ?

MarkEWaite:
I've freed enough space on ci.jenkins.io that the Jenkins HOME directory is now only 71% full and with some additional time we may get it below 70% full.  I'm OK with shipping this change.

PR #154, opened 2020-08-08
Skip publishing checks for java and javadoc tool

timja:
https://github.com/jenkinsci/analysis-model/pull/486 has been released for fixing the duplicate java and javadoc checks.

But not sure if we want checks showing up for these?

The javadoc ones can be quite spammy, and we don't have a quality gate setup for it atm.

Thoughts?

@uhafner @jglick @daniel-beck @XiongKezhi

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -156,7 +156,8 @@ def call(Map params = [:]) {
                                         filters: [excludeFile('.*Assert.java')],
                                         sourceCodeEncoding: 'UTF-8',
                                         trendChartType: 'TOOLS_ONLY',
-                                        referenceJobName: referenceJobName
+                                        referenceJobName: referenceJobName,
+                                        skipPublishingChecks: true
 
                                 // Default configuration for SpotBugs can be overwritten using a `spotbugs`, `checkstyle', etc. parameter (map).
                                 // Configuration see: https://github.com/jenkinsci/warnings-ng-plugin/blob/master/doc/Documentation.md#configuration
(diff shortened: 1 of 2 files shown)


XiongKezhi:
I will take a look at the javadoc warnings since the checks plugins are new projects and we tried to keep it documented from the start, but I'm not sure those warnings are useful, or even annoying for others since many of them are much larger, have a longer history and with different policies.

vars/buildPlugin.groovy
```diff
@@ -156,7 +156,8 @@ def call(Map params = [:]) {
                                         filters: [excludeFile('.*Assert.java')],
                                         sourceCodeEncoding: 'UTF-8',
                                         trendChartType: 'TOOLS_ONLY',
-                                        referenceJobName: referenceJobName
+                                        referenceJobName: referenceJobName,
+                                        skipPublishingChecks: true
```
uhafner, line 160: Would it make sense to add a customization as well (see SpotBugs and Checkstyle) here? Then we can leave the default unchanged and add the possibility to change. (Or change the default but let users with a more restrict rule set override) 

I think if someone gets too much warnings it would be better to change the build script (e.g. if we do not want to flag `@param` warnings then we should either disable the javadoc step (or at least disable this specific check if that is possible in the javadoc tool).

timja: I had a look into that, from what I could all it was all or nothing with the javadoc config =/.

customisation could be good, but I don't think that these are likely to be used much.

jglick:
> The javadoc ones can be quite spammy

I think this is ultimately something that can and should be corrected in `plugin-pom`. From [documentation](https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html#CHDCIBFC) it seems like `-Xdoclint:all,-missing` would fix the most common spam (`no @param for whatEver` etc.), while continuing to report genuine mistakes.

jglick (changes requested):
I think https://github.com/jenkinsci/plugin-pom/pull/356 should fix the root issue.

PR #155, opened 2020-08-08
buildDockerImage.groovy - hadolint and record issues

halkeye:


vars/buildDockerImage.groovy
```diff
@@ -20,6 +20,17 @@ def call(String imageName, Map config=[:]) {
...
 
     stages {
+      stage("Lint") {
+        agent {
+          docker { image "hadolint/hadolint" }
+        }
+        steps {
+          script {
+            writeFile(file: 'hadolint.json', text: sh(returnStdout: true, script: "/bin/hadolint --format json ${config.dockerfile} || true").trim())
+            recordIssues(tools: [hadoLint(pattern: 'hadolint.json')])
```
timja, line 30: Which ci servers does this buildDockerImage step run on? I.e does warnings-ng need installing anywhere else?

And did you want a quality gate for this or to just see what it looks like?

Is there a repo you want to test this on as well before we merge? I can push an origin branch for testing as that seems to be the easiest way unless you’ve tested it locally

halkeye: > Which ci servers does this buildDockerImage step run on? I.e does warnings-ng need installing anywhere else?

At the moment, as far as I know, its only used by jenkins-wiki-exporter, which is currently only built on ci.jenkins.io, working on infra.ci

> And did you want a quality gate for this or to just see what it looks like?

Quality gate fails the build right? I think it could be made optional, but i don't think hard fails on hadolint is worth it yet

> Is there a repo you want to test this on as well before we merge? I can push an origin branch for testing as that seems to be the easiest way unless you’ve tested it locally

I tested the same code on my install
https://github.com/halkeye/jenkins-shared-library/blob/master/vars/buildDockerfile.groovy#L24-L34
https://ci.g4v.dev/job/halkeye-docker/job/docker-jenkins/job/master/15/hadolint/

timja:
anyone else want to take a look?

PR #156, opened 2020-08-09
buildDockerImage but with k8s support so we can run it on infra.ci.jenkins.io

halkeye:


halkeye:
Sadly unit testing library doesn't have kubernetes support, only docker

https://github.com/jenkinsci/JenkinsPipelineUnit/blob/8330c417586442bc5d57fcac90294ea9e8a3d49b/src/main/groovy/com/lesfurets/jenkins/unit/declarative/AgentDeclaration.groovy

README.adoc
```diff
@@ -444,6 +444,25 @@ Name of the docker image to build
 buildDockerImage('plugins-site-api')
 ----
 
+=== buildDockerImage_k8s
+
+Builds a standard docker image and on master/tags pushes it to dockerhub but runs on the k8s servers (infra, releases, etc)
```
oleg-nenashev, line 449: Reference https://github.com/jenkinsci/jep/blob/master/jep/217/README.adoc and other architecture aspects under the hood_

README.adoc
```diff
@@ -444,6 +444,25 @@ Name of the docker image to build
 buildDockerImage('plugins-site-api')
 ----
 
+=== buildDockerImage_k8s
```
oleg-nenashev, line 447:
```suggestion
=== buildDockerAndPublishImage
```

K8s is an implementation detail and rather belongs to docs IMHO

halkeye: its the k8s (infra, release, trusted?) version of jenkins. buildDockerImage is the one used in the non k8s versions of infra.

I suspect the difference will be short lived, as we probably don't need to run the shared library on other installs (its only used by wiki exporter right now, and ci.jenkins can probably just docker build and not worry about all the pushing and stuff)

README.adoc
halkeye: I don't understand what this comment means

README.adoc
halkeye: I was unable to figure out how to make the shared library work in both environments. Totally open to ideas. The limiting thing is kubernetes{} agent and container{}

https://github.com/jenkins-infra/pipeline-library/pull/156/files#diff-d62a5e7da2898a017f06db97b3974b15R16

halkeye: gonna rename, don't need both anymore

vars/buildDockerAndPublishImage.groovy
```diff
@@ -0,0 +1,159 @@
...
+  if (!config.dockerfile) {
+    config.dockerfile = "Dockerfile"
+  }
+
+  pipeline {
+    agent {
+      kubernetes {
+        label 'helmfile'
+        inheritFrom 'jnlp-linux'
+        yaml """
```
olblak, line 19: It would be nice to have the YAML file in a dedicated *.yaml file in this repository so we easily automate file update using something similar to [updateCliPodTemplates.tpl](ttps://github.com/jenkins-infra/charts/blob/b79e9e7298c755a34f807626c1d7aa01501e1d55/updateCli/updateCli.d/updateCliPodTemplate.tpl#L42)

halkeye: I wasn't sure how the template would work with the shared library. Does it need to be in each repo or just here?

So played it safe

PR #161, opened 2020-08-17
Fixing building docker images on infra.ci.jenkins.io

halkeye:


vars/buildDockerAndPublishImage.groovy
```diff
@@ -97,42 +101,45 @@ spec:
...
-              GIT_SCM_URL = sh(returnStdout: true, script: "git remote show origin | grep 'Fetch URL' | awk '{print \$3}'").trim()
-              SCM_URI = GIT_SCM_URL.replace("git@github.com:", "https://github.com/")
-              BUILD_DATE = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssXXX").format(new Date())
               sh """
+                  export GIT_COMMIT_REV=\$(git log -n 1 --pretty=format:'%h')
+                  export GIT_SCM_URL=\$(git remote show origin | grep 'Fetch URL' | awk '{print \$3}')
+                  export SCM_URI=\$(echo \$GIT_SCM_URL | awk '{print gensub("git@github.com:","https://github.com/",\$3)}')
+
                   img build \
                       -t ${config.registry}${imageName} \
```
daniel-beck, line 110: Would use `withEnv` these  and then use single quotes for the `sh` argument.

vars/buildDockerAndPublishImage.groovy
```diff
@@ -97,42 +101,45 @@ spec:
         steps {
           container('img') {
             script {
```
daniel-beck, line 103: Unnecessary now?

halkeye: being a script? probably safe to delete now yea

vars/buildDockerAndPublishImage.groovy
halkeye: how would withEnv make this clearer? i'd still need to run sh, and its only used in this one block

That being said, I can revert now though, since the problem was actually credentials breaking with the wrong type

daniel-beck: Single-quoting let's you skip the escaping of `$`.

halkeye: OH! expose the one non bash one (which I think i did now) and convert it all to single quotes, yea totally! i thought you were saying the other way around.

PR #163, opened 2020-09-06
Switch to using the new incrementals publisher url and auth

halkeye:
@slide @timja 

as per discussions on irc, we decided to move the incremental publisher from azure functions to the k8s cluster. We have more control and logging and such.


timja (approved):
Tested here:
https://github.com/jenkinsci/theme-manager-plugin/pull/30

1st build had intentional incorrect token, 2nd had updated and it worked.

slide:
Awesome, thanks for this!

jglick:
Great, thanks!

PR #166, opened 2020-10-12
Update recentLTS

slide:
A few plugins that are common dependencies do not work with the previous recentLTS value. This updates to a more recent LTS version number.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -305,7 +305,7 @@ List<Map<String, String>> getConfigurations(Map params) {
  * Includes testing Java 8 and 11 on the newest LTS.
  */
 static List<Map<String, String>> recommendedConfigurations() {
-    def recentLTS = "2.164.1"
+    def recentLTS = "2.176.4"
     def configurations = [
         // Intentionally test configurations which have detected the most problems
         // Linux - Java 8 with plugin specified minimum Jenkins version


oleg-nenashev:
FTR https://github.com/jenkins-infra/pipeline-library/pull/145 . Looks like we have no consensus on whether it makes sense to retain the method at all

slide:
I'm fine with either method, updating the version or removing recommended configurations.

markyjackson-taulia (approved):
/lgtm

oleg-nenashev (approved):
I approve it, because it actually does what is documented in README: _It is also possible to use a buildPlugin.recommendedConfigurations() method to get recommended configurations for testing. Note that the recommended configuration may change over time, and hence your CI may break if the new recommended configuration is not compatible_ (https://github.com/jenkins-infra/pipeline-library)

Previous attempt to update by @batmat did not come through, and caused a lot of push back from other contributors. I am comfortable to merge it even if I approve it. FTR https://groups.google.com/g/jenkinsci-dev/c/xGGPTI1zCfc/m/j7QtknIZCgAJ for the last discussion, https://groups.google.com/forum/#!topic/jenkinsci-dev/lmZ_e7eFh2k for the previous one

vars/buildPlugin.groovy
```diff
@@ -305,7 +305,7 @@ List<Map<String, String>> getConfigurations(Map params) {
  * Includes testing Java 8 and 11 on the newest LTS.
  */
 static List<Map<String, String>> recommendedConfigurations() {
-    def recentLTS = "2.164.1"
+    def recentLTS = "2.176.4"
```
timja, line 308: this is erm not recent at all =/

batmat:
I will not block this, I've expressed my views multiple times already and don't want to be a chicken.
I think doing this will create a sea of dissatisfaction from various Jenkins plugin developers that will see their PRs and existing branches suddenly explode for no clear reason.
I strongly believe this should be either parameterized*, or deprecated and left untouched. 

https://github.com/search?q=org%3Ajenkinsci+recommendedConfigurations&type=code

\* I've given the feedback that this function should be parameterized since day 1 (well, before it technically as the code was not merged yet :smile:): https://github.com/jenkins-infra/pipeline-library/pull/86#discussion_r263697425

oleg-nenashev:
Yes, everybody is tired of iterations for that topic. I regret that I
introduced this API as is (with extensibility assumption), but at the same
time I am not ready to push this topic to any resolution.

Making a final call is up to whomever has time to maintain the lib (not me
unfortunately)

timja:
Not against merging this but I think it should just be binned. Better to have it explicit in the jenkinsfile

slide:
I think centralizing something like this is great for plugin developers. It allows their plugin to be built against recommended versions. We should be more active in updating the recommended configurations rather than making plugin developers go modify their files when releases are made.

oleg-nenashev:
I think we could do a solution for both worlds and introduce lastStable and
lastWeekly macros for jenkinsVersion fields (similarly to how we pass
'null' now). It might give developers an opportunity to explicitly declare
that they want to have a moving target testing in their Pipelines.

batmat:
@slide I thoroughly agree `recommendedConfigurations` is great. 
IMO, we also as an industry have learned the hard way that reproducibility is key. This is what Maven provided on the Java side years ago.
Here, despite I understand your concern that users may need a version bump regularly, I think this is fine and aligned with the "GitOps" modern industry standards. 
And I do think this is exponentially better for developers to have to have a PR from time to time, but keep being green, than suddenly having CI explode for an unrelated reason.

batmat:
While I think this is a very bad idea for anyone to do this, I _can_ agree with Oleg's proposal that people can opt-in for a moving target.
Though in practice, my concern is that a maintainer may at a given point in time decide to do this. But in the long term, with changing maintainers, actually do a disservice to the overall maintenability of the Jenkins components/plugins.

oleg-nenashev:
Alternate way is to embed the actual list of recommended configurations
into Plugin POM so that changes get picked up in a predictable way (when
dependabot updates the Plugin POM in a PR). It would be a decent solution
for actively maintained plugins with automated dependency management

slide:
I think there is enough opposition to just close this and re-evaluate how this is done.

PR #169, opened 2020-10-20
fix: stage name cannot be a GString

garethjevans:


vars/buildDockerAndPublishImage.groovy
```diff
@@ -130,7 +130,7 @@ spec:
           }
         }
       }
-      stage("Deploy ${config.mainBranch} as latest") {
+      stage("Deploy Main Branch as latest") {
```
timja, line 133: i would lowercase
```suggestion
      stage("Deploy main branch as latest") {
```

or even
```suggestion
      stage("Deploy latest") {
```

worried about it getting too long

timja (approved):

PR #171, opened 2020-11-14
Update infra checks to match the start of the url, so ports can't mess things up

halkeye:


MarkEWaite (approved):

Zastai:
Devil's advocate: this will consider `https://trusted.ci.jenkins.io.notreally` trusted.
Perhaps a regex match of `^https://trusted.ci.jenkins.io[:/]` would be better? Or is there no real security implication for the is-trusted check?

halkeye:
There's no security implication. Its just deciding if it should attempt to use credentials. On CI those credentials are not installed.

https://github.com/jenkins-infra/crawler/blob/master/Jenkinsfile#L42 for example will pull and sign on trusted, but on CI.jenkins.io just builds.

I agree though, normally starts with is not secure. Personally I'd parse the url and just confirm hostname. But since the whole thing is based on an env variable that can change its not remotely secure

PR #173, opened 2021-01-05
Chore: Allow running tests with OpenJDK 11 and 15: bump Groovy and Junit dependencies

dduportal:
This PR allows executing the library's unit tests (`mvn clean test`) when you have a recent OpenJDK (11 or 15).

It involves bumping the Groovy dependency (and associated compiler sub-dependencies) to 3.x (instead of 2.x), as well as Junit to 5.x.

Please note that the `pom.xml` is still specifying a compile target as `1.8` as before of course.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,10 +32,10 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <!-- Dependency versions -->
-        <groovy.version>2.4.17</groovy.version>
+        <groovy.version>3.0.7</groovy.version>
         <junit.version>4.13.1</junit.version>
         <jenkins-pipeline-unit.version>1.8</jenkins-pipeline-unit.version>
-        <groovy-eclipse-compiler.version>3.4.0-01</groovy-eclipse-compiler.version>
+        <groovy-eclipse-compiler.version>3.6.0-03</groovy-eclipse-compiler.version>
     </properties>
 
     <dependencies>
@@ -43,6 +43,8 @@
             <groupId>org.codehaus.groovy</groupId>
             <artifactId>groovy-all</artifactId>
             <version>${groovy.version}</version>
+            <!-- Type pom replaces the über "groovy-all.jar", for version 3+ -->
+            <type>pom</type>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


markyjackson-taulia (approved):
/lgtm

pom.xml
```diff
@@ -32,22 +32,24 @@
...
+        <groovy-eclipse-compiler.version>3.6.0-03</groovy-eclipse-compiler.version>
     </properties>
 
     <dependencies>
         <dependency>
             <groupId>org.codehaus.groovy</groupId>
             <artifactId>groovy-all</artifactId>
             <version>${groovy.version}</version>
+            <!-- Type pom replaces the über "groovy-all.jar", for version 3+ -->
+            <type>pom</type>
```
timja, line 47: do we still need this groovy dependency? or should we be specifying just the actual groovy jar? seems weird to depend on a pom artifact?

dduportal: I have no idea if it is required or not. On 1 side it might be right as this dependency is not mentionned in https://github.com/jenkinsci/JenkinsPipelineUnit#add-to-your-project-as-test-dependency .
On the other, my knowledge is not sufficient enough how to make this work without this 😱 

It looks like that the `groovy-all` (e.g. the Eclipse Groovy Compiler) is not the only solution: https://docs.groovy-lang.org/latest/html/documentation/tools-groovyc.html#section-groovyeclipse . But it is recommended when there is a joint Java-Groovy compilation, which should be the case for shared library if I understand correctly the `src/**` usage is that right?
=> If this assumption is valid, then https://github.com/groovy/groovy-eclipse/wiki/Groovy-Eclipse-Maven-plugin is to be used, but there is no mention of `groovy-all` in this doc. So I don't really know 🤷🏾‍♂️

dduportal: => http://groovy-lang.org/releasenotes/groovy-2.5.html#Groovy2.5releasenotes-Packaging might help

timja: I guess if it works

timja (approved):

oleg-nenashev (approved):
Should do the job. Would be nice to add it to the Pipeline though

pom.xml
```diff
@@ -32,22 +32,24 @@
...
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
```
jglick, line 52: https://github.com/jenkins-infra/pipeline-library/pull/272#issuecomment-1004318020

PR #176, opened 2021-01-13
[INFRA-2860] Improve library buildDockerAndPublishImage function

dduportal:
This PR introduces the following changes:

- Enable a "Test" phase after building a Docker Image, using Google's Container Structure Test tool with the TAR driver (and avoiding running unsecured pods with a nested container engine within).
- Allow reproductibility outside the build process by relying on a `Makefile` (documentation is added on how to use the Makefile for local builds)
- Add a unit-test harness for the function `buildDockerAndPublishImage()`
- Update HTML's documentation of the function `buildDockerAndPublishImage()`

Please note that an full "end to end" testing had been done in the following PRs (on other consumer repositories):

- https://github.com/jenkins-infra/docker-builder/pull/10 to validate a nominal usage (with tests)
- https://github.com/jenkins-infra/docker-mirrorbits/pull/4 to validate non breaking usage (on an already existing repository)

Signed-off-by: dduportal <damien.duportal@gmail.com>


vars/buildDockerAndPublishImage.groovy
```diff
@@ -1,176 +1,86 @@
...
+  final DockerConfig dockerConfig = new DockerConfig(imageName, config, new InfraConfig(env))
 
-  if (!config.mainBranch) {
-    config.mainBranch = "master"
-  }
+  // Retrieve Library's Static File Resources
+  final String makefileContent = libraryResource 'io/jenkins/infra/docker/Makefile'
+  final String podTemplate = libraryResource 'io/jenkins/infra/docker/pod-template.yml'
 
   pipeline {
```
jglick, line 14: I do not recommend trying to use Declarative syntax inside a library.

dduportal: The reasons behing keeping the declarative syntax here are because the effort to switch to scripted are crazy, and contribution gate should be reachable for a non pipeline expert.

Reading your comment, what would be the rationale behing using scripted pipeline (so we can assess the requirements here and see if it a reachable goal)?

Thanks for the review and the time taken on this subject !

dduportal:
Hello @MarckK , could you review the `README.adoc` in this PR to see if I missed something (and if my English is not that terrible)?

resources/io/jenkins/infra/docker/pod-template.yml
```diff
@@ -0,0 +1,51 @@
+---
+apiVersion: "v1"
+kind: "Pod"
+metadata:
+  labels:
+    jenkins: "agent"
```
olblak, line 6: I would be interested for label which could easily identify this job  like `job: docker-builder`

olblak, line 6:
```suggestion
    jenkins: "agent"
    job: "docker-builder"
```

resources/io/jenkins/infra/docker/pod-template.yml
```diff
@@ -0,0 +1,51 @@
...
+  annotations:
+    container.apparmor.security.beta.kubernetes.io/builder: unconfined
+    container.seccomp.security.alpha.kubernetes.io/builder: unconfined
+spec:
+  tolerations:
```
olblak, line 11: This block is specific to the kubernetes cluster `publick8s` and IMHO it should be here. On release.ci we have a generic jnlp pod that include this section and then every Jenkinsfile inherit that specific job like [here](https://github.com/jenkins-infra/release/blob/master/Jenkinsfile.d/components/remoting#L14)

olblak: I mean the tolerations block

resources/io/jenkins/infra/docker/pod-template.yml
```diff
@@ -0,0 +1,51 @@
...
+            values:
+            - linux
+      preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: 1
+        preference:
+          matchExpressions:
+          - key: agentpool
+            operator: In
+            values:
+            - highmemlinux
```
olblak, line 36: Do we really need a highmemlinux instance? My guess is no

olblak:
That's already a very nice pull request

vars/buildDockerAndPublishImage.groovy
jglick: > what would be the rationale behing using scripted pipeline

Just trying to keep Declarative to its original goal—scriptless, fully structured project definitions suitable for round-tripping.

resources/io/jenkins/infra/docker/README.adoc
```diff
@@ -0,0 +1,130 @@
...
+** `GIT_SCM_URL` (Defaults to the URL of the current git's remote named "origin"): URL of the git SCM repository's remote of the Docker image's code.
+** `SCM_URI` (Defaults from `GIT_SCM_URL` value): HTTPS (enforced) URL of `GIT_SCM_URL`.
+** `BUILD_DATE` (Default to "now" when called): UTC timestamp formatted in `%Y-%m-%dT%H:%M:%S` when the command occurs.
+
+You can check the file `Makefile` if you want to get a better deep-dive understanding.
+
+
+== Contributing
+
+* Contributing to this documentation or tooling must follow the contribution rules of this repository (please check link:../../../../../README.adoc#Contributing[])
```
MarckK, line 119: The link here leads to the README for repo, which doesn't have much in the way of Contributing Guidelines.

MarckK (dismissed):

MarckK:
Great PR, @dduportal!

vars/buildDockerAndPublishImage.groovy
dduportal: Interesting, I wasn't really aware of this original goal. Thanks for sharing!
To be honest, I feel like there is an "intermediate" situation (very well represented by this particular function) 
where the initial repository was using a declarative, and the need to "reuse" it (whole) arised.

The goal of this PR (despite the branch name and some back and forth with scripted) is to add a stage for "testing",
which does not feel like to migrate the whole pipeline in scripted. (I admit that I tried initially, but in the end there was no real need to only call the "build" or the "test" step alone: either you call the whole pipeline, or not.

=> I've seen mostly this kind of need in GitLab' Library or in YAML-powered CI (using YAML anchors) which bias me in the "I don't want to deal with a full fledge scripted language as I already am dealing with shell,Make, gradle, etc.".

resources/io/jenkins/infra/docker/pod-template.yml
dduportal: I need a precision on this: should the value of the label be static (eg. always `docker-builder`)  or should it be the name of the image currently being built (e.g. `docker-datadog`, `docker-terraform` etc. ?)

resources/io/jenkins/infra/docker/pod-template.yml
dduportal: Good point ! It looks like a PR is incoming for https://github.com/jenkins-infra/charts/blob/master/config/default/jenkins-infra.yaml#L85 :)
I'm removing the tolerations block from the shared library at the same time. Good point.

resources/io/jenkins/infra/docker/pod-template.yml
dduportal: Hey, at least the PR's check was fast... :D

resources/io/jenkins/infra/docker/pod-template.yml
olblak: The goal is to easily and clearly identify logs for a specific  job, so using dynamic names would be even better

dduportal: OK, this will need a bit more work then

resources/io/jenkins/infra/docker/pod-template.yml
dduportal: Removed the `tolerationsù` block

resources/io/jenkins/infra/docker/pod-template.yml
dduportal: Removed the `affinity` block as I prefer delegating to the orchestrator as much as possible

resources/io/jenkins/infra/docker/README.adoc
dduportal: Really good point. Incoming PR to add a contribution guide to the repo (as it is.. lacking).

resources/io/jenkins/infra/docker/pod-template.yml
dduportal: It's done: here is the final pod from the build logs (https://infra.ci.jenkins.io/job/Docker%20Builds/job/docker-builder/job/PR-10/24/console):

```
---
apiVersion: "v1"
kind: "Pod"
metadata:
  annotations:
    container.apparmor.security.beta.kubernetes.io/builder: "unconfined"
    container.seccomp.security.alpha.kubernetes.io/builder: "unconfined"
    buildUrl: "http://jenkins-infra:8080/job/Docker%20Builds/job/docker-builder/job/PR-10/24/"
    runUrl: "job/Docker%20Builds/job/docker-builder/job/PR-10/24/"
  labels:
    jenkins: "agent"
    job/kind: "docker-image-build"
    docker-image/name: "builder"
    jenkins/jenkins-infra-agent: "true"
    jenkins/label-digest: "168581a5f7dd7bc78734253113adb8b4115e61ba"
    jenkins/label: "Docker_Builds_docker-builder_PR-10_24-tb7k6"
  name: "docker-builds-docker-builder-pr-10-24-tb7k6-zz4t4-tsjxp"
spec:
  containers:
  - command:
    - "cat"
    image: "jenkinsciinfra/builder:latest"
    name: "builder"
    tty: true
    volumeMounts:
    - mountPath: "/home/jenkins/agent"
      name: "workspace-volume"
      readOnly: false
  - args:
    - "********"
    - "docker-builds-docker-builder-pr-10-24-tb7k6-zz4t4-tsjxp"
    env:
    - name: "JENKINS_SECRET"
      value: "********"
    - name: "JENKINS_TUNNEL"
      value: "jenkins-infra-agent:50000"
    - name: "JENKINS_AGENT_NAME"
      value: "docker-builds-docker-builder-pr-10-24-tb7k6-zz4t4-tsjxp"
    - name: "JENKINS_NAME"
      value: "docker-builds-docker-builder-pr-10-24-tb7k6-zz4t4-tsjxp"
    - name: "JENKINS_AGENT_WORKDIR"
      value: "/home/jenkins/agent"
    - name: "JENKINS_URL"
      value: "http://jenkins-infra:8080/"
    image: "jenkins/inbound-agent:latest-jdk11"
    imagePullPolicy: "Always"
    name: "jnlp"
    resources:
      limits:
        memory: "512Mi"
        cpu: "500m"
      requests:
        memory: "512Mi"
        cpu: "500m"
    tty: false
    volumeMounts:
    - mountPath: "/home/jenkins/agent"
      name: "workspace-volume"
      readOnly: false
  nodeSelector:
    kubernetes.io/os: "linux"
  restartPolicy: "Never"
  volumes:
  - emptyDir:
      medium: ""
    name: "workspace-volume"
```

dduportal: It defines 2 labels:

- `job/kind` which is static and always the same value `docker-image-build`
- `docker-image/name` which is dynamic and holds the Docker Image name being built

@olblak is it ok for you?

src/io/jenkins/infra/DockerConfig.groovy
```diff
@@ -0,0 +1,61 @@
+package io.jenkins.infra
+
+
+import java.text.SimpleDateFormat
+import java.util.Date
+
+class DockerConfig implements Serializable {
```
jglick, line 7:
```suggestion
class DockerConfig {
```

src/io/jenkins/infra/DockerConfig.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+  // Custom getter to avoid declaring NonCPS method called from constructor
+  String getRegistry() {
+    if (registry == null) {
+      if (infraConfig != null) {
+        return infraConfig.getDockerRegistry()
+      }
+      return "noregistry"
+    }
+    return registry
```
jglick, line 59: If you like idiomatic Groovy:

```suggestion
    registry ?: infraConfig?.dockerRegistry ?: 'noregistry'
```

resources/io/jenkins/infra/docker/pod-template.yml
olblak: Yeaaa that's awesome, thank about that.

vars/buildDockerAndPublishImage.groovy
olblak: I think @jglick  is a very good recommendation and if you are not planning to switch to a full scripted pipeline, I guess it worthwhile to document it about the why? People get inspiration from this public shared-library.

dduportal: Fair point, I'm only wondering where should be document such thing? I was thinking about adding comment on the top of the groovy file defining the function. Is there another location for this?

src/io/jenkins/infra/DockerConfig.groovy
dduportal: Thanks for the suggestion!

If I understand correctly, the reson of NOT implementing Serializable for this object is because there is no pipeline step involved (while the object `InfraConfig` should implement as it uses the `env` object).

Is it correct?

src/io/jenkins/infra/DockerConfig.groovy
dduportal: Oh it's a nice tip, thanks! I did not know about the [Elvis operator](https://groovy-lang.org/operators.html#_elvis_operator).

src/io/jenkins/infra/DockerConfig.groovy
```diff
@@ -0,0 +1,55 @@
...
+
+    this.credentials = config.credentials
+    if (!config.credentials) {
+      this.credentials = "jenkins-dockerhub"
+    }
+
+    this.mainBranch = config.mainBranch
+    if (!config.mainBranch) {
+      this.mainBranch = "master"
+    }
```
v1v, line 42: Cosmetic suggestions to use `map.get('key', 'default')`

```suggestion
    this.dockerfile = config.get('dockerfile', 'Dockerfile')

    this.credentials = config.get('credentials', 'jenkins-dockerhub')

    this.mainBranch = config.get('mainBranch', 'master')
```

test/groovy/BaseTest.groovy
```diff
@@ -1,4 +1,4 @@
-import com.lesfurets.jenkins.unit.BasePipelineTest
+import com.lesfurets.jenkins.unit.declarative.*
```
v1v, line 1: NIT

```suggestion
import com.lesfurets.jenkins.unit.declarative.DeclarativePipelineTest
```

resources/io/jenkins/infra/docker/README.adoc
```diff
@@ -0,0 +1,130 @@
...
+* Command prompt with bash (or zsh/ksh)
+
+* GNU Make 3.81+ command line
+
+* A https://github.com/moby/buildkit[Buildkit] compliant container tool
+** The command line top use is defined by the value of the environment variable `CONTAINER_BIN` which defauls to `img`
+** Our pipelines are using (Jessie Frazelle's) https://github.com/genuinetools/img[`genuinetools/img`] command line in v0.5.11, for security purposes (container in containers)
+** You can use https://docs.docker.com/get-docker/[Docker v20.10+] and its command line instead, by setting `CONTAINER_BIN` to `docker`
+
+* https://github.com/GoogleContainerTools/container-structure-test[Google's Container Structure Test] command line in v1.9.1+
```
v1v, line 55: If this PR is only about https://github.com/jenkins-infra/docker-builder will it make sense to point to its README.md instead?

v1v (commented):
Question regarding decoupling the build scripts from this repo and move them to the project instead.

- IIUC, https://github.com/jenkins-infra/docker-builder is one of the consumers, but in order to use the same CI experience, it's required some manual actions as described in the README.md section

Would the gitsubmodules be a good approach to fix this? https://github.com/jenkinsci/docker/blob/master/Makefile#L48 already uses that approach for the BATS dependencies.

src/io/jenkins/infra/DockerConfig.groovy
dduportal: Thanks for pointing this language improvement! I was not aware of this method : it is way more readable with your suggestion!

resources/io/jenkins/infra/docker/README.adoc
dduportal: In fact this PR (and the associated shared library) are targeted at any repository defining a Docker image to be built, in the organization `jenkins-infra` (at least?).

So https://github.com/jenkins-infra/docker-builder is the first one to really utilize the test part, however it is expected that other repos/pipelines are going to (ref. https://github.com/jenkins-infra/charts/blob/master/config/default/jenkins-infra.yaml#L178 => This is the JCasc config of the instance infra.ci.jenkins.io, which defines a "Docker Builds" GH org  folder). For instance: https://github.com/jenkins-infra/docker-datadog/, https://github.com/jenkins-infra/docker-terraform/, etc.

The trick here is that `docker-builder` is used by all, including itself: we are in a egg-and-chicken situation which makes the initial README confusing.
=> I confirm that this README is aimed to be generic for each of this image.

dduportal:
Hello @v1v, thanks for the review, your suggestions had been applied! Much appreciated.

This PR is aimed to be a first "simple but functional" version, this is why the "git submodule" approach had not been considered yet.  But it would be really a nice addition if it adds another "choice" for the contributor instead of having to manually `curl` the Makefile (and even a better "default" way with the curl as a backup).

How would you see the git submodule to be configured? Should it be a `git clone --recursive https://github.com/jenkins-infra/docker-xxxx` that would references another git repository (which would contains the Makefile and the associated README), with the pipeline library here responsible to retrieve the Makefile from the submodule as well?

jglick:
Not sure if you can use submodules from a Groovy library, for the `resources/` subdirectory for example. Do not recall having ever tried. I suppose if the Git plugin supports a submodule checkout option, that could be included in the library definition in Jenkins config.

dduportal:
@jglick Good point. It means that the submodule should be on the "consumer" repo (e.g. the repo being built) and not on the "library" side. If no one disagree, I propose to switch the discussion about the git submodule on a subsequent issue + PR as it brings another level of challenges (but we can also decide here and update the PR accordingly).

dduportal:
@v1v  the "git submodule" improvement is tracked in https://issues.jenkins.io/browse/INFRA-2880 . Did I capture correctly the discussion? Are you willing to help me on this improvement (could be a great excuse to have a virtual coffee together while implementing this ;) )

resources/io/jenkins/infra/docker/README.adoc
dduportal: The missing "contribution" doc is now tracked in https://issues.jenkins.io/browse/INFRA-2881 , and tagged as a "newbie issue" as it is a topic that can be taken by contributors with no prior Jenkins expertise.

olblak:
> Not sure if you can use submodules from a Groovy library, for the `resources/` subdirectory for example. Do not recall having ever tried. I suppose if the Git plugin supports a submodule checkout option, that could be included in the library definition in Jenkins config.

I usually avoid git submodules in favor of cloning the repository that I need. I remember having that discussion with @MarkEWaite a while time ago, and if I remember correctly that was also his approach.

v1v (approved):

v1v:
>@v1v the "git submodule" improvement is tracked in https://issues.jenkins.io/browse/INFRA-2880 . Did I capture correctly the discussion? Are you willing to help me on this improvement (could be a great excuse to have a virtual coffee together while implementing this ;) )

I'm totally in!! 💯

vars/buildDockerAndPublishImage.txt
```diff
@@ -1,15 +1,50 @@
 <p>
```
timja, line 1: not really sure what the point of a .txt file is for documentation that includes html?

shouldn't this be ascii doc or markdown?

timja (approved):
lgtm

vars/buildDockerAndPublishImage.txt
dduportal: My understanding is that it is used in the UI, in the "Pipeline Syntax" section to document the available libraries.

jglick: Yes, it will be run through the configured `Jenkins.markdownFormatter`, but currently this file extension is expected by `workflow-cps-global-lib`.

timja: huh never knew about it.

Unfortunately it's quite buried and doesn't work so well with organisation folders, i.e. you have to actually go to a branch that has run, not very discoverable,

It's not in pipeline syntax but in Global Variables reference:
https://ci.jenkins.io/job/Plugins/job/abap-ci-plugin/job/master/pipeline-syntax/globals#buildPlugin

documented here:
https://www.jenkins.io/doc/book/pipeline/shared-libraries/#directory-structure

timja: and FYI infra.ci.jenkins.io doesn't seem to have a markup formatter configured, all the output just comes as plain text

jglick: > you have to actually go to a branch that has run

Yes because Jenkins has to look this up from an actual build record.

PR #177, opened 2021-01-19
fix: regression when determining docker registry name

garethjevans:
slight regression in the pipeline library when overridding the registry
name for docker builds.  Previously the user was expected to include a
trailing slash at the end of the registry name, this is no longer expected.

An example of the previous usage can be seen https://github.com/garethjevans/jenkins-image/blob/05a56a379a63c43955f28163e0be44662246ada8/Jenkinsfile#L1


timja (approved):

dduportal (approved):
LGTM 👍 Good catch, I'm sorry for the disturbance

PR #184, opened 2021-02-10
Update recentLTS

MRamonLeon:
Keep up to date the recentLTS on recommendedConfiguration.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -309,7 +309,7 @@ List<Map<String, String>> getConfigurations(Map params) {
  * Includes testing Java 8 and 11 on the newest LTS.
  */
 static List<Map<String, String>> recommendedConfigurations() {
-    def recentLTS = "2.164.1"
+    def recentLTS = "2.263.3"
     def configurations = [
         // Intentionally test configurations which have detected the most problems
         // Linux - Java 8 with plugin specified minimum Jenkins version


timja:
I think you will get shot down,

see history in:
https://github.com/jenkins-infra/pipeline-library/pull/166
and
https://github.com/jenkins-infra/pipeline-library/pull/146
https://github.com/jenkins-infra/pipeline-library/pull/92

https://github.com/jenkins-infra/pipeline-library/pull/92#commitcomment-38865433

The recommendation currently is to not use this functionality and to put the recommended configurations into your Jenkinsfile and adjust it as you wish.

Ideally not hardcoding against a jenkins version outside of the pom as it's a nightmare for a non maintainer to test against a newer version of Jenkins for bom / PCT / new core feature / whatever.
As they don't have write access...

MRamonLeon:
I didn't want to add more controversy to the topic, I know it's sensible. I just wanted to honor the code in `recommendedConfiguration` by bumping the `recentLTS` variable.

A question, don't know whether it was discussed before. Why not just? To avoid breaking any plugin using `recommendedConfigurations` but at the same time avoid the usage of an old version. We can also add a log entry to recommend an alternative to `recommendedConfiguration` before removing it.

```
    def configurations = [
        [ platform: "linux", jdk: "8", jenkins: null, javaLevel: "8" ],
        [ platform: "windows", jdk: "8", jenkins: null, javaLevel: "8" ],
        [ platform: "linux", jdk: "11", jenkins: null, javaLevel: "8" ],
        [ platform: "windows", jdk: "11", jenkins: null, javaLevel: "8" ]
    ]
    return configurations
}
```

Just thinking 🤔

timja:
Sounds like it could work, in order to optimise CI usage we don't normally recommend Windows Java 11, as:

1. windows CI is less reliable
2. windows java 11 failure where it didn't fail on 8 (or linux 11) is unheard of or at least incredibly rare

jglick:
Can we merge #145?

jglick:
Superseded.

PR #192, opened 2021-03-11
feat: allow build metadata to be calculated from a script

garethjevans:
This is used when `automaticSemanticVersioning` is configured, as an example it can be used to calculate the build metadata to be appended to the calculated  next version.

e.g. 

To determine a version from a parent dockerfile, the following could be used:

```
metadataFromSh: 'cat Dockerfile | grep FROM | sed "s|FROM jenkins/jenkins:|+|"'
```

this will calculate something like:

```
+2.280
```

which will be appended to the new version as follows:

```
1.1.0+2.280
```

### DIFF
--- a/src/io/jenkins/infra/DockerConfig.groovy
+++ b/src/io/jenkins/infra/DockerConfig.groovy
@@ -25,6 +25,8 @@ class DockerConfig {
 
   String gitCredentials
 
+  String metadataFromSh
+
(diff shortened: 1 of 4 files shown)


dduportal (approved):
LGTM 👍

markyjackson-taulia (approved):

oleg-nenashev (approved):
Looks good

PR #196, opened 2021-03-30
Allow `buildDockerAndPublish` library to be called multiple times on a single pipeline build

dduportal:
This Pull requests allow to define different Docker images to be built for a given repository, or to add custom pipeline code around the library call.

Usage example (also added to the library's txt internal doc):

```groovy
properties([
  buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '5')),
  pipelineTriggers([cron('@daily')]),
])

def registryName = 'dduportal'
def principalBranch = 'main'

parallel(
  failFast: true,
  'docker': {
     buildDockerAndPublishImage('docker', [
      mainBranch: principalBranch,
      dockerfile: 'docker/Dockerfile',
      registry: registryName,
    ])
  },
  'coreruntime-2.2': {
     buildDockerAndPublishImage('coreruntime-2.2', [
      mainBranch: principalBranch,
      dockerfile: 'coreruntime/22/Dockerfile',
      registry: registryName,
    ])
  },
)
```

In particular, this PR changed the following:

* Switch pipeline syntax to scripted for `buildDockerAndPublishImage.groovy`
* Introduce a new input parameter `imageDir` to specify the Docker build context when required (defaults to the directory containing the specified `Dockerfile`)
* Updating stage names (showing the image name) and the recordIssue (hadolint) step to handle multiple calls on a single build
* Introduce a new kind of (optional) test harness: "Common test", which expects a file `common-cst.yml` to be present at the repository's root. The existing `cst.yml` is renamed to "Image Specific test", and the file is expected to be found on the `imageDir` directory
* Support images where the tag is specified in the provided Docker Image Name (example: `jenkins-infra/coreruntime:2.2`). It's a simple and not smart support: the git tag is appended with an hyphen instead of a colon.

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -13,129 +13,90 @@ def call(String imageName, Map config=[:]) {
   // Customize Pod label to improve build analysis
   final String yamlPodDef = podYamlTemplate.replaceAll('\\$IMAGE_NAME', imageName).replaceAll('\\$?\\{IMAGE_NAME\\}', imageName)
 
-  // This function uses Declarative Syntax as we haven't met (yet) use case where only a subpart of the pipeline is needed: either we run it "whole" or not.
-  // It means that the function is acting more as a "template" of an initial declarative standalone pipeline, than a real life well-constructed library.
-  // A second reason is to lower the contribution bar by avoiding the "know very well Groovy" requirement (but this argument can be disputed as the test unit is pure Groovy...)
-  // If you feel like to rewrite this function in scripted, you can spend this effort of course :)
-  pipeline {
-    agent {
-      kubernetes {
-        inheritFrom 'jnlp-linux'
-        defaultContainer 'builder'
-        yaml yamlPodDef
-      } // kubernetes
-    } // agent
+  final boolean semVerEnabled = dockerConfig.automaticSemanticVersioning && env.BRANCH_NAME == dockerConfig.mainBranch
 
-    environment {
-      BUILD_DATE       = "${dockerConfig.buildDate}"
-      IMAGE_NAME       = "${dockerConfig.imageName}"
-      DOCKERFILE       = "${dockerConfig.dockerfile}"
-      PLATFORM         = "${dockerConfig.platform}"
-      SEMANTIC_RELEASE = "${dockerConfig.automaticSemanticVersioning}"
-    } // environment
+  podTemplate(
+    inheritFrom: 'jnlp-linux',
+    yaml: yamlPodDef,
+  ) {
+    node(POD_LABEL) {
+      container('builder') {
+        withEnv([
+          "BUILD_DATE=${dockerConfig.buildDate}",
+          "IMAGE_NAME=${dockerConfig.imageName}",
+          "DOCKERFILE=${dockerConfig.dockerfile}",
+          "PLATFORM=${dockerConfig.platform}",
+        ]) {
+          stage('Prepare') {
+            withCredentials([usernamePassword(credentialsId: dockerConfig.credentials, passwordVariable: 'DOCKER_REGISTRY_PSW', usernameVariable: 'DOCKER_REGISTRY_USR')]) {
+              checkout scm
 
-    stages {
-      stage('Next Version') {
-        when {
-          allOf {
-            environment name: 'SEMANTIC_RELEASE', value: 'true'
-            branch dockerConfig.mainBranch  
-          }
-        }
-        steps {
-          container('next-version') {
-            script {
-              nextVersion = sh(script:"${dockerConfig.nextVersionCommand}", returnStdout: true).trim()
-              if (dockerConfig.metadataFromSh != '') {
-                metadata = sh(script: "${dockerConfig.metadataFromSh}", returnStdout: true).trim()
-                nextVersion = nextVersion + metadata
-              }
+              // Logging in on the Dockerhub helps to avoid request limit from DockerHub
+              sh 'echo "${DOCKER_REGISTRY_PSW}" | img login -u "${DOCKER_REGISTRY_USR}" --password-stdin'
+
+              // The makefile to use must come from the pipeline to avoid a nasty user trying to exfiltrate data from the build
+              // Even though we have mitigation trhough the multibranch job config only allowed to build PRs from repo. contributors
+              writeFile file: 'Makefile', text: makefileContent
+            } // withCredentials
+          } // stage
+
+          // Automatic tagging on principal branch is not enabled by default
+          if (semVerEnabled) {
+            stage('Next Version') {
+              container('next-version') {
+                nextVersion = sh(script:"${dockerConfig.nextVersionCommand}", returnStdout: true).trim()
+                if (dockerConfig.metadataFromSh != '') {
+                  metadata = sh(script: "${dockerConfig.metadataFromSh}", returnStdout: true).trim()
+                  nextVersion = nextVersion + metadata
+                } // if
+              } // container
+              echo "Next Release Version = $nextVersion"
+            } // stage
+          } // if
+
+          stage("Lint") {
+            try {
+              sh 'make lint'
+            } finally {
+              recordIssues enabledForFailure: true, tool: hadoLint(pattern: 'hadolint.json')
             }
-          }
-          echo "Next Release Version = $nextVersion"
-        } // steps
-      } // stage
+          } // stage
 
-      stage('Prepare') {
-        environment {
-          DOCKER_REGISTRY = credentials("${dockerConfig.credentials}")
-        }
-        steps {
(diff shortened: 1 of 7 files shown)


vars/buildDockerAndPublishImage.groovy
```diff
@@ -13,96 +13,73 @@ def call(String imageName, Map config=[:]) {
...
-          allOf {
-            expression { env.SEMANTIC_RELEASE == 'true' }
-            branch dockerConfig.mainBranch  
+      // Test step is not mandatory as not all repositories
+      if (fileExists('cst.yml')) {
+        stage("Test") {
+          if (sh(returnStatus: true, script: 'make test') > 0) {
+            currentBuild.result = 'FAILURE'
+            error('Test Failed')
           }
```
jglick, line 76: First of all, avoid ever setting `currentBuild.result`. Throwing `error` will automatically mark the build as a failure (assuming it is not caught and suppressed). And why are you doing all this anyway?
```suggestion
          sh 'make test'
```
also fails the build if `make` failed. KISS

vars/buildDockerAndPublishImage.groovy
```diff
@@ -13,96 +13,73 @@ def call(String imageName, Map config=[:]) {
...
-        }
-        steps {
+        } //stage
+      } // if
+
+      // Automatic tagging on principal branch is not enabled by default
+      if (semVerEnabled) {
+        stage("Semantic Release") {
           echo "Configuring credential.helper"
           sh 'git config --local credential.helper "!f() { echo username=\\$GIT_USERNAME; echo password=\\$GIT_PASSWORD; }; f"'
```
jglick, line 84: Cannot comment on unmodified lines below (thx GitHub) but

```groovy
            echo "Tagging New Version: $nextVersion"
            sh "git tag $nextVersion"
            echo "Pushing Tag"
            sh "git push origin --tags"
```

can be simplified to just

```groovy
            sh 'git tag $nextVersion && git push origin --tags'
```

assuming `nextVersion` is bound as an environment variable (which it should be). Do not ever pass `"`-quoted args to `sh`, and you should never have any series of adjacent `sh` and/or `echo`.

vars/buildDockerAndPublishImage.groovy
```diff
@@ -13,96 +13,73 @@ def call(String imageName, Map config=[:]) {
...
-            recordIssues enabledForFailure: true, tool: hadoLint(pattern: 'hadolint.json')
-          }
-        } // post
-      } //stage
+        int lintStatus = sh(returnStatus: true, script: 'make lint')
+        recordIssues enabledForFailure: true, tool: hadoLint(pattern: 'hadolint.json')
+        if (lintStatus > 0) {
+          currentBuild.result = 'FAILURE'
+          error('Lint Failed')
+        }
```
jglick, line 61:
```suggestion
        sh 'make lint'
        recordIssues enabledForFailure: true, tool: hadoLint(pattern: 'hadolint.json')
```
(If `make lint` failed, no reason to also record a Jenkins publisher.)

vars/buildDockerAndPublishImage.groovy
```diff
@@ -13,96 +13,73 @@ def call(String imageName, Map config=[:]) {
...
-    stages {
-      stage('Next Version') {
-        when {
-          allOf {
-            environment name: 'SEMANTIC_RELEASE', value: 'true'
-            branch dockerConfig.mainBranch  
-          }
-        }
-        steps {
+  final Boolean semVerEnabled = dockerConfig.automaticSemanticVersioning && env.BRANCH_NAME == dockerConfig.mainBranch
```
jglick, line 16:
```suggestion
  final boolean semVerEnabled = dockerConfig.automaticSemanticVersioning && env.BRANCH_NAME == dockerConfig.mainBranch
```
?

vars/buildDockerAndPublishImage.groovy
```diff
@@ -13,96 +13,73 @@ def call(String imageName, Map config=[:]) {
...
-                metadata = sh(script: "${dockerConfig.metadataFromSh}", returnStdout: true).trim()
-                nextVersion = nextVersion + metadata
-              }
-            }
-          }
+            nextVersion = sh(script:'jx-release-version', returnStdout: true).trim()
+            if (dockerConfig.metadataFromSh != '') {
+              metadata = sh(script: "${dockerConfig.metadataFromSh}", returnStdout: true).trim()
+              nextVersion = nextVersion + metadata
+            } // if
```
jglick, line 37: Assuming these variables are only being used in subsequent `sh` steps, better to avoid ever round-tripping through Groovy.
```groovy
sh 'jx-release-version > version.txt'
```
etc. And then of course if you do not have a specific reason for having intervening Pipeline steps, you can coalesce with subsequent `sh` steps.

This whole script still looks way too complicated. You could probably reduce it to a few lines of `podTemplate`, `withCredentials`, `sh`, and `recordIssues`, with all of the logic in some Bash script (or Ruby, Makefile, etc.) stored _elsewhere_.

vars/buildDockerAndPublishImage.groovy
timja: you may want to record it so that it gets reported to github with the direct issues rather than having to look through the build

the idiom in scripted would normally be
```suggestion
        try {
          sh 'make lint'
        } finally {
          recordIssues enabledForFailure: true, tool: hadoLint(pattern: 'hadolint.json')
        }
```

jglick: You _can_ do that, if the `make lint` output is illegible. Just make sure `recordIssues` does not produce an error if no `hadolint.json` file exists.

dduportal: 👍

vars/buildDockerAndPublishImage.groovy
dduportal: The goal to have separate stages for different "build step" is important for letting human understand what is going on when it fails. You're correct that we **might** be able to move a lot within the `Makefile` (for instance the conditional for the `test` goal) but whatever language it's written on, we still have to test it.

The pipeline unit help works out of the bo (except potential bugs) for this level of abstraction, while if we have to build a fully fledged framework, not really error-prone, for testing a Makefile based on its environment, it's possible but the effort is not really something that valuable. 

Thanks for the initial tip  though, we might be able to simplify some parts here (taking in account that it's currently being migrated into scripted syntax so the idioms have to be changed)

vars/buildDockerAndPublishImage.groovy
dduportal: (for the record, I was mistakenly using the `createShMock` method which led me to these overcomplicated scripts. With your pointers and some tries/retries I understand where my error is, and your feedback are clear!)

vars/buildDockerAndPublishImage.groovy
dduportal: 👍

vars/buildDockerAndPublishImage.groovy
garethjevans: Sounds like there is need for a pipeline linter

dduportal: Thanks for the tips @jglick . Gareth and I will check what can be improved on a subsequent PR, as this code block is outside the scope here, but for sure we are willing to improve this to simplify the pipeline code 👍

vars/buildDockerAndPublishImage.groovy
```diff
@@ -9,133 +9,119 @@ def call(String imageName, Map config=[:]) {
...
-        }
-        steps {
-          echo "Configuring credential.helper"
-          sh 'git config --local credential.helper "!f() { echo username=\\$GIT_USERNAME; echo password=\\$GIT_PASSWORD; }; f"'
 
-          withCredentials([usernamePassword(credentialsId: "${dockerConfig.gitCredentials}", passwordVariable: 'GIT_PASSWORD', usernameVariable: 'GIT_USERNAME')]) {
-            echo "Tagging New Version: $nextVersion"
-            sh "git tag $nextVersion"
+          // Automatic tagging on principal branch is not enabled by default
+          if (semVerEnabled) {
```
garethjevans, line 93: this will fail if enabled, and more than one image is defined.

dduportal: Good point. I'm not sure on how to prevent this, as I don't see a way to now if the end user is calling other instances of the function. WDYT?

olblak (approved):

vars/buildDockerAndPublishImage.groovy
olblak: I would just update the comment with Gareth input and move forward

olblak: We don't have that need today and we can still improve the process once that need raise. Commenting the behavior will help in the futur

olblak:
@dduportal I let you merge this PR once you are ready

PR #197, opened 2021-03-30
feat: make next version command configurable

garethjevans:
Make the  command to be run to determine the next version configurable.  Whilst most scenarios will probably be fine with the default `jx-release-version`, it may be required to use something a little more complex like `jx-release-version -next-version=semantic:strip-prerelease`


olblak (approved):
it sounds good to me even though I didn't tested it

dduportal (approved):
LGTM 👍

dduportal:
Good idea! Thanks a lot!

PR #198, opened 2021-03-31
Validate readyness for release

alecharp:
With the current build process, for plugins using Maven, it can happen that the CI is not detecting that the code is not valid for a release. Which means that, maintainers cannot trust the CI when they want to perform a release (issues can happen).

With this changeset, the PR and main branch of the repositories would be validated to see if the content of the repository is good enough for to be released.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -86,6 +86,7 @@ def call(Map params = [:]) {
                                         '-Dspotbugs.failOnError=false',
                                         '-Dcheckstyle.failOnViolation=false',
                                         '-Dcheckstyle.failsOnError=false',
+                                        '-Pjenkins-release',
                                 ]
                                 // jacoco had file locking issues on Windows, so only running on linux
                                 if (isUnix()) {
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -108,7 +108,7 @@ def call(Map params = [:]) {
                                 if (skipTests) {
                                     mavenOptions += "-DskipTests"
                                 }
-                                mavenOptions += "clean install"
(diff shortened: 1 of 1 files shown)


jglick (changes requested):
Not necessary, and perhaps harmful, for incrementalified plugins: https://github.com/jenkinsci/plugin-pom/blob/986acb190dc2f3cd112b31451c8b4dea6c133449/pom.xml#L1362-L1393

alecharp:
I knew I didn't understood how bad it could break things, the problem is that we can have a main branch which build correctly in CI and fail miserably on release because the Javadoc is never built by CI. All I want is to fix that.

Is the `mvn verify` or the usage of the profile that is bothering you @jglick?

jglick:
> we can have a main branch which build correctly in CI and fail miserably on release because the Javadoc is never built by CI

Which, again, does not happen for an incrementalified plugin, because the CI pipeline passes `-Dset.changelist` which activates all the mechanics of an actual release, including sources and Javadoc. So if you want to do this, it should happen for example as an `else` block after https://github.com/jenkins-infra/pipeline-library/blob/46bc2cb9096ae06cd3a0ee6542ffd817b893ec8d/vars/buildPlugin.groovy#L95-L101

(I would also encourage plugins to move to JEP-229 which is immune to such risks by construction: every default branch build is a true release candidate.)

The switch from `verify` to `install` is also suspect. In theory it should not affect properly written POMs, but due to the vagaries of `maven-hpi-plugin` this is impractical and such a change might in fact break the build for multimodule (reactor) repositories.

alecharp:
Ok I see. Let's forget this then.

jglick:
Well, adding `-Pjenkins-release` could be useful for non-incrementalified plugins.

PR #200, opened 2021-04-12
Feat: add a new 'updatecli()' shared library

dduportal:
This PR adds a new shared library to allow executing `updatecli` in your pipelines.

Example:

```groovy
@Library('pipeline-library@pull/200/head') _

parallel(
  failFast: true,
  'docker-image': {
    // Do something...
  },
  'updatecli': {
    withCredentials([string(credentialsId: 'updatecli-github-token', variable: 'UPDATECLI_GITHUB_TOKEN')]) {
      updatecli(action: 'diff')
      if (env.BRANCH_NAME == 'main') {
        updatecli(action: 'apply')
      }
    }
  },
)
```

### DIFF
--- a/vars/updatecli.groovy
+++ b/vars/updatecli.groovy
@@ -7,8 +7,8 @@ def call(userConfig = [:]) {
     updatecliDockerImage: 'ghcr.io/updatecli/updatecli:latest',
   ]
 
-  // No Deep Merge - https://e.printstacktrace.blog/how-to-merge-two-maps-in-groovy/
-  final Map finalConfig = defaultConfig + userConfig
+  // Merging the 2 maps - https://blog.mrhaki.com/2010/04/groovy-goodness-adding-maps-to-map_21.html
+  final Map finalConfig = defaultConfig << userConfig
 
   // final String podYamlTemplate = libraryResource 'io/jenkins/infra/docker/pod-template.yml'
   // Customize Pod label to improve build analysis
--- a/vars/updatecli.groovy
+++ b/vars/updatecli.groovy
@@ -1,0 +1,34 @@
+
+def call(userConfig = [:]) {
+  def defaultConfig = [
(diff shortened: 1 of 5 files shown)


vars/updatecli.groovy
```diff
@@ -0,0 +1,34 @@
...
+def call(userConfig = [:]) {
+  def defaultConfig = [
+    action: 'diff',
+    config: './updatecli/updatecli.d',
+    values: './updatecli/values.yaml',
+    updatecliDockerImage: 'ghcr.io/updatecli/updatecli:latest',
+  ]
+
+  // No Deep Merge - https://e.printstacktrace.blog/how-to-merge-two-maps-in-groovy/
+  final Map finalConfig = defaultConfig + userConfig
```
timja, line 11: not sure if any advantage, but we normally use this shorter form: https://github.com/hmcts/cnp-jenkins-library/blob/master/vars/notifyBuildEvent.groovy#L19-L22

dduportal: Thanks @timja ! I have to admit that, even though I understand with your comment that this code is somehow merging the 2 maps, I'm not sure to really understand the operator `<<` . The Groovy documentation at https://groovy-lang.org/operators.html#_bit_shift_operators tells me that it's a shift operator, but how does it work with a Map ? Is there an override in the Map class or something?

olblak (dismissed):
I am quite excited to easily it \o/ from our infra

vars/updatecli.groovy
timja: it's called `leftShift()` https://blog.mrhaki.com/2010/04/groovy-goodness-adding-maps-to-map_21.html

dduportal: @timja Switched to your proposal, with a link to the page you mentioned. Thanks!

timja (approved):

PR #201, opened 2021-04-14
feat: perform a github release when using semantic versioning

garethjevans:


vars/buildDockerAndPublishImage.groovy
```diff
@@ -120,6 +120,23 @@ def call(String imageName, Map config=[:]) {
...
+                String origin = sh(returnStdout: true, script: 'git remote -v | grep origin | grep push | sed \'s/^origin\\s//\' | sed \'s/\\s(push)//\'').trim() - '.git'
+                String org = origin.split('/')[3]
+                String repository = origin.split('/')[4]
+
+                try {
+                    String release = sh(returnStdout: true, script: "gh api /repos/${org}/${repository}/releases | jq -e -r '.[] | select(.draft == true and .name == \"next\") | .id'").trim()
+                    sh "gh api -X PATCH -F draft=false -F name=${env.TAG_NAME} -F tag_name=${env.TAG_NAME} /repos/${org}/${repository}/releases/$release"
+                } catch (err) {
+                    echo "Release named 'draft' does not exist"
+                }
```
jglick, line 136: This could all be one `sh` step.

(And avoid Groovy interpolation on `sh` args.)

garethjevans: I've split it up to aid testing - i want to handle the scenario where release-drafter is not configured on the repo.

When you say avoid "interpolation" - is concatenation ok?

dduportal (approved):
LGTM 👍

vars/buildDockerAndPublishImage.groovy
jglick: > When you say avoid "interpolation" - is concatenation ok?

Not sure what you mean by concatenation in this context. I was referring to use of double-quoted `GString`s as `sh` args, which is considered a security risk (you will see warnings in the build page accordingly). Use a literal single-quoted `String` (`'` or `'''` for multiline) and evaluate environment variables using Bash syntax. If you _must_ pass a Groovy variable in, use `withEnv`, though usually this can be avoided and the Pipeline script minimized by doing calculations inside a single `sh` step.

PR #204, opened 2021-04-27
Feat: Allow to customize docker images for BuildDockerAndPublish and UpdateCli libs

dduportal:
This PRs allows to specify a custom docker image to be used for the pod templates of the 2 libraries BuildDockerAndPublish and UpdateCli.

The goal is to avoid breaking builds when the upstream images have been updated (such as https://github.com/jenkins-infra/charts/pull/1106) with a breaking change.
=> But keeping the "latest" image as a default is still the rule as the balance is still good when not requiring human validation for most of the updates. But now there is an escape hatch for "sensitive pipelines".

This PR also adds a fix to a bug introduced with #196 , where the hadolint report's pattern was the absolute path while it should only be a pattern relative to the workspace or current dir (was blocking @olblak for the image docker-mirrorbit: a great end to test use case for this PR :) )


timja (dismissed):

dduportal:
Sorry @timja , I've slipped an hotfix in this PR while you reviewed it

PR #205, opened 2021-04-29
Deprecating `javaLevel`

jglick:
For years now `java.level` has been mandatory in the POM (https://github.com/jenkinsci/plugin-pom/pull/88) and this is something intrinsic to the source code, not something you would ever want to override for a particular build; it will almost always need to stay fixed at `8`. (Not to be confused with the JDK version, which you may switch to `11` to test against Java 11.) Thus it is silly to redundantly specify `javaLevel: 8` in your `Jenkinsfile`.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -27,7 +27,9 @@ def call(Map params = [:]) {
         String label = config.platform
         String jdk = config.jdk
         String jenkinsVersion = config.jenkins
-        String javaLevel = config.javaLevel
+        if (config.javaLevel != null) {
+            echo 'WARNING: javaLevel is deprecated and should not be used'
+        }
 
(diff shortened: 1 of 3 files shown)


vars/buildPlugin.groovy
```diff
@@ -27,7 +27,9 @@ def call(Map params = [:]) {
         String label = config.platform
         String jdk = config.jdk
         String jenkinsVersion = config.jenkins
-        String javaLevel = config.javaLevel
+        if (config.javaLevel != null) {
+            echo 'WARNING: javaLevel is deprecated and should not be used'
```
basil, line 31: Perhaps something like

```
WARNING: Ignoring deprecated "javaLevel" parameter. This parameter should be removed from your "Jenkinsfile".
```

might make it more clear to someone reading the message that the parameter is in fact dead code while also informing them what specific action they can take to (trivially) fix the warning.

basil (approved):

basil (approved):

timja (approved):

PR #207, opened 2021-05-04
buildDockerAndPublishImage: update resources used by the pod template

dduportal:
Following #204 , an unplanned change was applied to the docker builds: the removed pod template had an indentation issue resulting in the pod resources/limits not applied.
The result is that the "docker" builds using this library are now being applied limits and killed when requesting more.
Exemple: https://github.com/jenkins-infra/docker-jenkins-weekly/pull/177

This PR increase the resources for the "builder" container: 2 CPUs to acknoledge the fact that a JVM process (like a Jenkins controller) or a compilation could happen during the "docker build" step, and 2 Gb for the same reasons.

The resources for the `next-version` had been decreased as it's a lightweight process.


timja (approved):

dduportal:
Tested successfully on https://infra.ci.jenkins.io/blue/organizations/jenkins/Docker%20Builds%2Fdocker-jenkins-weekly/detail/PR-177/7/pipeline . Let's go

PR #208, opened 2021-05-11
JEP-229 Wait longer than 15 minutes to get an executor

timja:
Please don't fail my release because there was no executors for 15 minutes -.-

https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fazure-vm-agents-plugin/detail/master/177/pipeline

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -272,17 +272,19 @@ void prepareToPublishIncrementals() {
 void maybePublishIncrementals() {
     if (new InfraConfig(env).isRunningOnJenkinsInfra() && currentBuild.currentResult == 'SUCCESS') {
         stage('Deploy') {
-          timeout(15) {
+          timeout(300) {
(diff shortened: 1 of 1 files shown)


vars/infra.groovy
```diff
@@ -272,17 +272,19 @@ void prepareToPublishIncrementals() {
 void maybePublishIncrementals() {
     if (new InfraConfig(env).isRunningOnJenkinsInfra() && currentBuild.currentResult == 'SUCCESS') {
         stage('Deploy') {
-          timeout(15) {
+          timeout(300) {
```
timja, line 275: not sure whether this is needed, I added it as a safety net in case the connection get's broken to try prevent zombie builds

jglick:
I have also thought about just writing a tiny bespoke plugin with a `RunListener` that would detect builds whose artifacts look like incrementals, and automatically upload asynchronously.

PR #210, opened 2021-05-22
Simplify incrementals publishing

timja:
I've had a number of failures on incrementals publishing from remoting errors, causing long core builds to fail and CD releases not to happen

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -277,7 +277,7 @@ void maybePublishIncrementals() {
                         contentType: 'APPLICATION_JSON',
                         validResponseCodes: '100:599',
                         timeout: 300,
-                        requestBody: '{"build_url":"'$BUILD_URL'"}',
+                        requestBody: /{"build_url":"$BUILD_URL"}/,
                         authentication: 'incrementals-publisher-token'
         }
     } else {
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -272,13 +272,16 @@ void prepareToPublishIncrementals() {
 void maybePublishIncrementals() {
     if (new InfraConfig(env).isRunningOnJenkinsInfra() && currentBuild.currentResult == 'SUCCESS') {
         stage('Deploy') {
-            httpRequest url: 'https://incrementals.jenkins.io/',
-                        httpMode: 'POST',
-                        contentType: 'APPLICATION_JSON',
-                        validResponseCodes: '100:599',
-                        timeout: 300,
-                        requestBody: /{"build_url":"$BUILD_URL"}/,
-                        authentication: 'incrementals-publisher-token'
+            withCredentials([string(credentialsId: 'incrementals-publisher-token', variable: 'FUNCTION_TOKEN')]) {
+                httpRequest 
+                    url: 'https://incrementals.jenkins.io/',
+                    httpMode: 'POST',
+                    contentType: 'APPLICATION_JSON',
+                    validResponseCodes: '100:599',
+                    timeout: 300,
+                    requestBody: /{"build_url":"$BUILD_URL"}/,
+                    customHeaders: [[maskValue: true, name: 'Authorization', value: 'Bearer ${FUNCTION_TOKEN}']]
+            }
         }
     } else {
         echo 'Skipping deployment to Incrementals'
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -279,7 +279,7 @@ void maybePublishIncrementals() {
                     validResponseCodes: '100:599',
                     timeout: 300,
                     requestBody: /{"build_url":"$BUILD_URL"}/,
-                    customHeaders: [[name: 'Authorization', value: 'Bearer ${FUNCTION_TOKEN}']],
+                    customHeaders: [[name: 'Authorization', value: 'Bearer $FUNCTION_TOKEN']],
                     consoleLogResponseBody: true
             }
         }
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -273,8 +273,7 @@ void maybePublishIncrementals() {
     if (new InfraConfig(env).isRunningOnJenkinsInfra() && currentBuild.currentResult == 'SUCCESS') {
         stage('Deploy') {
             withCredentials([string(credentialsId: 'incrementals-publisher-token', variable: 'FUNCTION_TOKEN')]) {
-                httpRequest 
-                    url: 'https://incrementals.jenkins.io/',
+                httpRequest url: 'https://incrementals.jenkins.io/',
                     httpMode: 'POST',
                     contentType: 'APPLICATION_JSON',
                     validResponseCodes: '100:599',
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -273,7 +273,7 @@ void maybePublishIncrementals() {
     if (new InfraConfig(env).isRunningOnJenkinsInfra() && currentBuild.currentResult == 'SUCCESS') {
         stage('Deploy') {
             httpRequest url: 'https://incrementals.jenkins.io/',
-                        httpMode: 'POST'
+                        httpMode: 'POST',
                         contentType: 'APPLICATION_JSON',
                         validResponseCodes: '100:599',
                         timeout: 300,
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -279,7 +279,8 @@ void maybePublishIncrementals() {
                     validResponseCodes: '100:599',
                     timeout: 300,
                     requestBody: /{"build_url":"$BUILD_URL"}/,
-                    customHeaders: [[maskValue: true, name: 'Authorization', value: 'Bearer ${FUNCTION_TOKEN}']]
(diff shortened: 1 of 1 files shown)


vars/infra.groovy
```diff
@@ -272,23 +272,13 @@ void prepareToPublishIncrementals() {
...
-                }
-            }
-          }
+            httpRequest url: 'https://incrementals.jenkins.io/',
+                        httpMode: 'POST'
+                        contentType: 'APPLICATION_JSON',
+                        validResponseCodes: '100:599',
+                        timeout: 300,
+                        requestBody: '{"build_url":"'$BUILD_URL'"}',
+                        authentication: 'incrementals-publisher-token'
```
jglick, line 281: Does this actually work? This will be a `Bearer` token, not username/password. https://www.jenkins.io/doc/pipeline/steps/http_request/#httprequest-perform-an-http-request-and-return-a-response-object is not specific.

vars/infra.groovy
```diff
@@ -272,23 +272,13 @@ void prepareToPublishIncrementals() {
...
-                    }
-                }
-            }
-          }
+            httpRequest url: 'https://incrementals.jenkins.io/',
+                        httpMode: 'POST'
+                        contentType: 'APPLICATION_JSON',
+                        validResponseCodes: '100:599',
+                        timeout: 300,
+                        requestBody: '{"build_url":"'$BUILD_URL'"}',
```
jglick, line 280: Does not look like valid Groovy syntax. Maybe try
```suggestion
                        requestBody: /{"build_url":"$BUILD_URL"}/,
```

vars/infra.groovy
```diff
@@ -272,23 +272,13 @@ void prepareToPublishIncrementals() {
...
-                            '''
-                        }
-                    }
-                }
-            }
-          }
+            httpRequest url: 'https://incrementals.jenkins.io/',
+                        httpMode: 'POST'
+                        contentType: 'APPLICATION_JSON',
+                        validResponseCodes: '100:599',
```
jglick, line 278: Will this at least print a warning if it gets an error?

vars/infra.groovy
```diff
@@ -272,23 +272,13 @@ void prepareToPublishIncrementals() {
...
-    curl.exe --retry 10 --retry-delay 10 -i -H "Authorization: Bearer %FUNCTION_TOKEN%" -H "Content-Type: application/json" -d "{""build_url"":""%BUILD_URL%""}" "https://incrementals.jenkins.io/" || echo Problem calling Incrementals deployment function
-                            '''
-                        }
-                    }
-                }
-            }
-          }
+            httpRequest url: 'https://incrementals.jenkins.io/',
+                        httpMode: 'POST'
+                        contentType: 'APPLICATION_JSON',
```
jglick, line 277: Hmm, weird that this is an enum rather than the actual string value. :shrug:

jglick (commented):
Let us see if it works…

vars/infra.groovy
timja: should do

vars/infra.groovy
```diff
@@ -272,23 +272,13 @@ void prepareToPublishIncrementals() {
...
-                            bat '''
-    curl.exe --retry 10 --retry-delay 10 -i -H "Authorization: Bearer %FUNCTION_TOKEN%" -H "Content-Type: application/json" -d "{""build_url"":""%BUILD_URL%""}" "https://incrementals.jenkins.io/" || echo Problem calling Incrementals deployment function
-                            '''
-                        }
-                    }
-                }
-            }
-          }
+            httpRequest url: 'https://incrementals.jenkins.io/',
+                        httpMode: 'POST'
```
timja, line 276:
```suggestion
                        httpMode: 'POST',
```

vars/infra.groovy
timja: > 13:57:21  Response Code: HTTP/1.1 403 Forbidden

vars/infra.groovy
```diff
@@ -272,23 +272,16 @@ void prepareToPublishIncrementals() {
...
-                    }
-                }
+            withCredentials([string(credentialsId: 'incrementals-publisher-token', variable: 'FUNCTION_TOKEN')]) {
+                httpRequest url: 'https://incrementals.jenkins.io/',
+                    httpMode: 'POST',
+                    contentType: 'APPLICATION_JSON',
+                    validResponseCodes: '100:599',
+                    timeout: 300,
+                    requestBody: /{"build_url":"$BUILD_URL"}/,
+                    customHeaders: [[name: 'Authorization', value: 'Bearer $FUNCTION_TOKEN']],
```
jglick, line 282:
```suggestion
                    customHeaders: [[name: 'Authorization', value: "Bearer $FUNCTION_TOKEN"]],
```

If the plugin interpolates env vars in supplied arguments, that is a [bug in the plugin](https://www.jenkins.io/doc/developer/plugin-development/pipeline-integration/#variable-substitutions) and we should not rely on the bad behavior.

vars/infra.groovy
```diff
@@ -272,23 +272,16 @@ void prepareToPublishIncrementals() {
...
-    curl.exe --retry 10 --retry-delay 10 -i -H "Authorization: Bearer %FUNCTION_TOKEN%" -H "Content-Type: application/json" -d "{""build_url"":""%BUILD_URL%""}" "https://incrementals.jenkins.io/" || echo Problem calling Incrementals deployment function
-                            '''
-                        }
-                    }
-                }
+            withCredentials([string(credentialsId: 'incrementals-publisher-token', variable: 'FUNCTION_TOKEN')]) {
+                httpRequest url: 'https://incrementals.jenkins.io/',
+                    httpMode: 'POST',
+                    contentType: 'APPLICATION_JSON',
+                    validResponseCodes: '100:599',
```
jglick, line 279: An awkward way of saying “do not fail on error” but :shrug:

vars/infra.groovy
timja: no it wasn't working thanks for this

timja:
Tested here:
https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/9

Works 🎉

vars/infra.groovy
timja: no it doesn't work, it requires username / password

basil:
> Tested here:
> [jenkinsci/jenkins-infra-test-plugin#9](https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/9)
> 
> Works tada

Albeit with [`InterpolatedSecretsAction` taped onto the result](https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/PR-9/15/). I had to move away from HTTP Request Plugin at my company for the same reason.

basil:
Wow, I guess that chased `InterpolatedSecretsAction` away. Though I'm not sure how it compares from a security perspective. Seems to me that either the original `InterpolatedSecretsAction` was a false positive or else `workflow-cps` ought to warn in the case of `+` as well. Probably a better topic for a `workflow-cps` bug.

timja:
> Wow, I guess that chased `InterpolatedSecretsAction` away. Though I'm not sure how it compares from a security perspective. Seems to me that either the original `InterpolatedSecretsAction` was a false positive or else `workflow-cps` ought to warn in the case of `+` as well. Probably a better topic for a `workflow-cps` bug.

This runs on the controller in java not in a process on an agent like the `Action` is warning against so I don't see any security concern here.

It would be good if the plugin handled credential binding better / it's `Authenticate` option was more flexible than just username/password auth, but I don't believe it should be an issue for this

jglick:
https://github.com/jenkins-infra/pipeline-library/pull/210#discussion_r645617802 hmm yeah, means that **Arguments** in https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/PR-9/15/flowGraphTable/ is deliberately suppressing information. Would be better if the request plugin had support for passing a `credentialsId` with `Bearer` authorization. I think the warning may be overzealous; what we really care about are secrets interpolated into arguments for `sh` and other durable tasks, but there are lots of miscellaneous steps for which this is harmless enough (@car-roll @dwnusbaum).

basil:
> I don't believe it should be an issue for this

I concur. This PR looks good to me.

> Would be better if the request plugin had support for passing a `credentialsId` with Bearer authorization

Or for my use case, the ability to pass a credential as a query string parameter when making API calls to arbitrary GET endpoints without requiring the user to construct the string with the credential in it.

> I think the warning may be overzealous

That's been my experience in practice with our Jenkins jobs. It caught a few legitimate issues but also flagged a few false positives like this one.

jglick:
> false positives like this one

Specifically I would say that if the step in question is consuming a contextual `Launcher` & `EnvVars` then the warning is legitimate: secrets should be passed via the environment and interpreted by the launched process on the agent. But if the step is doing stuff inside the controller, as this one is (the whole point of this PR), then there is no reasonable alternative to using `GString` interpolation unless the step happens to directly support a `credentialsId`.

timja:
@jglick can I get a +1?

jglick (approved):
If it works, :shipit:

jetersen:
Belated LGTM! 👏 Thanks for fix this, I remember complaining about this 😅

PR #212, opened 2021-06-01
feat: add a new global library 'withRetryAndTimeout' to allow circuit breaking pipeline steps

dduportal:
This PR provide a circuit breaker of type "retry with timeout per retry" library.

Its initial case it to ensure that some of the jenkins-infra's docker builds involving network operations were not blocking (e.g. 10 min timeout before failing the step), but retrying a few times before really failing the job.

Ref. https://issues.jenkins.io/browse/JENKINS-51454

Please note that there are incoming PRs on the Jenkins Pipeline-related plugins that would allow using pure `retry` and `timeout` methods instead of try/catch: the library would have to be reconsidered or updated and better tested at this moment.

Another added value of this library: an environment variable `PIPELINE_RETRY_COUNTER` is injected to hold the retry counter's values (e.g. "how many retries?"). This value can be used by consumer to identify uniquely some resources such as warning-ng reports.

### DIFF
--- a/vars/withRetryAndTimeout.groovy
+++ b/vars/withRetryAndTimeout.groovy
@@ -9,11 +9,10 @@ def call(Map userConfig = [:], Closure body) {
   // Merging the 2 maps - https://blog.mrhaki.com/2010/04/groovy-goodness-adding-maps-to-map_21.html
   final Map finalConfig = defaultConfig << userConfig
 
-  def counter = 0
+  def counter = 1
 
   // From https://issues.jenkins.io/browse/JENKINS-51454?focusedCommentId=389893&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-389893
   retry(finalConfig.retries) {
-    counter += 1
     try {
       timeout(time: finalConfig.bodyTimeout, unit: finalConfig.bodyTimeoutUnit) {
         withEnv(["PIPELINE_RETRY_COUNTER=${counter}"]) {
@@ -26,6 +25,7 @@ def call(Map userConfig = [:], Closure body) {
         throw ex
       }
       throw new RuntimeException(ex)
+      counter += 1
     }
   }
 }
--- a/vars/withRetryAndTimeout.groovy
+++ b/vars/withRetryAndTimeout.groovy
@@ -20,7 +20,7 @@ def call(Map userConfig = [:], Closure body) {
         }
       }
     } catch(org.jenkinsci.plugins.workflow.steps.FlowInterruptedException ex){
-      def causes = ex.getCauses()
+      def causes = ex.causes
       if(causes.find { ! (it instanceof org.jenkinsci.plugins.workflow.steps.TimeoutStepExecution$ExceededTimeout) }) {
         throw ex
       }
--- a/vars/withRetryAndTimeout.groovy
+++ b/vars/withRetryAndTimeout.groovy
@@ -1,0 +1,31 @@
+
+def call(Map userConfig = [:], Closure body) {
+  def defaultConfig = [
+    retries: 3,
+    bodyTimeout: 10,
+    bodyTimeoutUnit: 'MINUTES',
+  ]
+
+  // Merging the 2 maps - https://blog.mrhaki.com/2010/04/groovy-goodness-adding-maps-to-map_21.html
+  final Map finalConfig = defaultConfig << userConfig
+
+  def counter = 0
+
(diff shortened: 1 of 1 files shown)


vars/withRetryAndTimeout.groovy
```diff
@@ -0,0 +1,31 @@
...
+  ]
+
+  // Merging the 2 maps - https://blog.mrhaki.com/2010/04/groovy-goodness-adding-maps-to-map_21.html
+  final Map finalConfig = defaultConfig << userConfig
+
+  def counter = 0
+
+  // From https://issues.jenkins.io/browse/JENKINS-51454?focusedCommentId=389893&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-389893
+  retry(finalConfig.retries) {
+    counter += 1
```
timja, line 16: shouldn't this be incremented at the end? it hasn't retried yet?

timja (commented):
seems sensible? any other thoughts?

vars/withRetryAndTimeout.groovy
dduportal: Depends on if you like 0-indexed or 1-indexed arrays :) The intent is to have the value `1` during the first pass, `2` on the next pass, etc. And the increment at the beginning is to make it easier to "know" but I realize it's not really idiomatic, can lead to errors and the `withEnv` solves this intent.

dduportal: Fixed: let me know if the last change fit what you wanted (If not I need guidance).

dduportal:
@timja I'm confused by the compiler error messages (I got the same locally).
I understand that the exceptions types are not known by the Eclipse Groovy compiler: in real life it works as expected (see. https://github.com/jenkins-infra/docker-jenkins-weekly/pull/208) but not in the build/test environment.

Should we add the hpi plugins of the workflow steps (which provides timeout and the exception types) as part of the build/test dependencies here? Or did I miss something?

vars/withRetryAndTimeout.groovy
```diff
@@ -0,0 +1,31 @@
...
+          body()
+        }
+      }
+    } catch(org.jenkinsci.plugins.workflow.steps.FlowInterruptedException ex){
+      def causes = ex.getCauses()
+      if(causes.find { ! (it instanceof org.jenkinsci.plugins.workflow.steps.TimeoutStepExecution$ExceededTimeout) }) {
+        throw ex
+      }
+      throw new RuntimeException(ex)
+      counter += 1
```
jtnord, line 28: is this not purely dead code - never to be reached as the previous line thows a new Exception?

vars/withRetryAndTimeout.groovy
```diff
@@ -0,0 +1,31 @@
...
+  // From https://issues.jenkins.io/browse/JENKINS-51454?focusedCommentId=389893&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-389893
+  retry(finalConfig.retries) {
+    try {
+      timeout(time: finalConfig.bodyTimeout, unit: finalConfig.bodyTimeoutUnit) {
+        withEnv(["PIPELINE_RETRY_COUNTER=${counter}"]) {
+          body()
+        }
+      }
+    } catch(org.jenkinsci.plugins.workflow.steps.FlowInterruptedException ex){
+      def causes = ex.getCauses()
```
jglick, line 23: or
```suggestion
      def causes = ex.causes
```

jglick:
see https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/144

vars/withRetryAndTimeout.groovy
dduportal: Thanks, I should have paid more attention to this. Instruction moved with a comment: WDYT?

dduportal:
> see jenkinsci/workflow-basic-steps-plugin#144
Should I look at this change as a FYI, or is it related to the dependency question (or something else)?

timja:
> > see [jenkinsci/workflow-basic-steps-plugin#144](https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/144)
> 
> @jglick Should I look at this change as a FYI, or is it related to the dependency question (or something else)?

it means we wouldn't need this PR at all

dduportal:
Fair, given that the linked PR is "almost there" (modulo the GitHub incident/ end of approval window / merge / release), I'll go back to the road of "try catch on pipeline with approved call to causes" on the few repositories (https://github.com/jenkins-infra/docker-jenkins-weekly/ and https://github.com/jenkins-infra/docker-jenkins-lts) which requires the change right now. Thanks a lot for taking time to review and give advises on this!

PR #216, opened 2021-06-15
Delete `jenkins-repos` profile

jglick:
I believe we should make this file empty and perhaps even deprecate the library method to use it.

* The name `settings-azure.xml` is now misleading and obsolete. Made sense when its primary purpose was to activate the mirror in Azure that no longer exists.
* There is no need to explicitly add a repository definition for `https://repo.jenkins-ci.org/public/`. All Jenkins components should already be specifying this in their POM. (If you forget, things break for developers and also for JEP-229: https://github.com/jenkinsci/mock-slave-plugin/pull/103)
* Nor should `https://repo.jenkins-ci.org/incrementals/` be added unconditionally. This is activated by the `consume-incrementals` profile in the parent POMs for projects which have opted in to that.

Reminded by https://github.com/jenkinsci/plugin-pom/pull/419#issuecomment-861377441.


timja (approved):

dduportal (approved):
Good point, thanks! 👍

jglick:
> All Jenkins components should already be specifying this in their POM.

To be clear, if merged, CI builds of (Maven-based) components which neglect that will begin to fail. Example error message (from [here](https://github.com/jenkinsci/mock-slave-plugin/runs/2823915771?check_suite_focus=true#step:4:22)):

```
[FATAL] Non-resolvable parent POM for org.jenkins-ci.plugins:mock-slave:${changelist}: Could not find artifact org.jenkins-ci.plugins:plugin:pom:4.19 in central (https://repo.maven.apache.org/maven2) and 'parent.relativePath' points at no local POM @ line 3, column 13
```

This could affect some old plugins I suppose, depending on how the POM was originally authored. New components from [archetype](https://github.com/jenkinsci/archetypes/blob/dbf82ba12aa702dde9bd0dfdf2b810047a0b1a3d/empty-plugin/src/main/resources/archetype-resources/pom.xml#L53-L64) should be fine.

dduportal:
@jglick Could you write an email to the dev list before merging?

jglick:
Sure, or I thought we were switching to Discourse for this sort of thing?

dduportal:
Good point, I would vote a Discourse subject an I'll take the task of sending an email with the link to your discourse post (for people note using community). Does it sounds good to you?

jglick:
Hmm, now not sure. I see https://community.jenkins.io/c/community/announcement/13 but this is under https://community.jenkins.io/c/community/11 rather than https://community.jenkins.io/c/contributing/10. I will just go with the dev list for now.

timja (approved):

jglick:
Ah I see https://groups.google.com/g/jenkinsci-dev/c/2Ou4kTITPR0/m/tKsqPdXAAQAJ the contributor announcement category is still just a proposal.

timja (approved):

PR #217, opened 2021-06-30
[JENKINS-43353] Use new `disableConcurrentBuilds/abortPrevious` project option

jglick:
If https://github.com/jenkinsci/workflow-job-plugin/pull/200 is merged and released and ci.jenkins.io updated, simplify the cancellation of older running builds.


dduportal:
PR updated with conflict solved. As soon as the plugin is released, we'll install it on ci (and everywhere \o/)

PR #220, opened 2021-08-19
feat: introduces the attribute 'useContainerAgent' to replace 'useAci'

dduportal:
As per https://issues.jenkins.io/browse/INFRA-2918, the ACI agents are
replaced by Kubernetes agents.

Following this change, the library `buildPlugin` has been updated:
attribute `useAci` is deprecated in favor of the new `useContainerAgent`.

To avoid breaking pipelines, a "soft depreciation" policy is implemented for now:
when attribute `useAci` is defined, then a deprecation warning is printed and
`useContainerAgent` is set the value of `useAci`.


vars/buildPlugin.groovy
```diff
@@ -14,7 +14,12 @@ def call(Map params = [:]) {
     def repo = params.containsKey('repo') ? params.repo : null
     def failFast = params.containsKey('failFast') ? params.failFast : true
     def timeoutValue = params.containsKey('timeout') ? params.timeout : 60
-    def useAci = params.containsKey('useAci') ? params.useAci : false
+
+    def useContainerAgent = params.containsKey('useContainerAgent') ? params.useContainerAgent : false
+    if (params.containsKey('useAci')) {
```
timja, line 19: looks to be ~100 usages of useAci, https://github.com/search?q=org%3Ajenkinsci+useAci&type=code

including some forks of buildPlugin that won't be updated by this.

Any plans to migrate them or provide a more visible warning, e.g. PR check?

dduportal: Not really thought about this, as I never faced a change of this magnitude.

I'm all hears to any pointers if there would be a method to mass-open PRs to do this?

For the check, do you think using a pipeline instruction next to the warning echoed to add a warning check could help? I'm not sure if the concept of "warning" exists on the GH checks? Is there something you had in mind?

timja: There’s no concept of warning but either neutral or failure could work.

last time we did this was when we removed findbugs / checkstyle for 30-40 repoes I think I manually filed PRs for them all it didn’t take very long although this is more repoes,


I know that Spotify has leveraged probot to do fleet wide migrations across micro services.

internally we add a slack alert to our pipelines for similar things and DM devs who raise PRs on every commit

dduportal: As we discussed during the last infra team meeting, this PR need to be updated to:
- Raise a github check message to developers (non blocking, non failing)

After this PR, we might want to do the following:
- Communicate to developers through the mailing list
- Script the 100-ish PRs to be opened using the `gh` command line or anything else

In ~1/2 months, we'll have to:
- Update the github check to fail the PRs (assuming most of the 100-ish PRs have been merged)

README.adoc
```diff
@@ -28,7 +28,9 @@ buildPlugin()
 ==== Optional arguments
 
 * `repo` (default: `null`  inherit from Multibranch) - custom Git repository to check out
-* `useAci` (default: `false`) - uses link:https://azure.microsoft.com/en-us/services/container-instances/[Azure Container Instances] for build instead of link:https://azure.microsoft.com/en-us/services/virtual-machines/[Azure Virtual Machines].
+* `useContainerAgent` (default: `false`) - uses a link:https://github.com/jenkins-infra/documentation/blob/main/ci.adoc#container-agents[Container agent] instead of a Virtual Machine: usually faster to start and generates less costs for the project
```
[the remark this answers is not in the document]
dduportal: that is a really good point! Do we agree that changing this default would be a breaking change though?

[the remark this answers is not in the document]
jglick: It would definitely be breaking for plugins which assumed that `linux` implied Docker. For other cases, maybe, unclear.

PR #237, opened 2021-11-04
Work around issue with symlinks in security release with ATH

timja:
https://github.com/jenkinsci/jenkins/pull/5880#issuecomment-961469352


jglick:
Maybe OK as a hotfix to get ATH running again.

dduportal (approved):

PR #247, opened 2021-11-15
configurable git default branch

lifeofguenter:
Something that might soon be relevant: https://github.blog/2020-07-27-highlights-from-git-2-28/ and git-client-plugin recently added support for this: https://github.com/jenkinsci/git-client-plugin/pull/733

Making this configurable might help plugins that have "master" hardcoded still work if the default should ever change to something other than "master" (I am surprised this is not already the case on the build agents 🤔), but also help those who want to already change (https://github.com/jenkinsci/bitbucket-branch-source-plugin/pull/513) to make the build work.

This should hopefully be the correct place to do the change - I am _assuming_ that executors are not long running, so doing a change here won't affect any other builds?


timja (commented):
Seems fine, any other thoughts?

Other option would be to set it explicitly on the agents

lifeofguenter:
Yah, I think the cleanest would be for the tests to be written while not assuming "master" as the default branch - but until then, I think setting it on the agents directly could break stuff 😅

jglick:
What is the purpose of this change? `buildPlugin` runs `checkout scm` which checks out a specific commit and builds it. It makes no difference which branch (if any) that commit is on. What is broken that this PR would fix?

lifeofguenter:
Correct, this change is not for the pipeline scm itself, but for any git commands that run inside the pipeline. That would most likely happen as part of tests.

jglick:
> for any git commands that run inside the pipeline. That would most likely happen as part of tests.

And what specific test cases did you have in mind? Tests should never be written to assume environmental configuration like this.

lifeofguenter:
you can have a look at https://github.com/jenkinsci/git-client-plugin/pull/733 for more context on the issue and https://github.com/jenkinsci/bitbucket-branch-source-plugin/pull/513 for a branch with the renaming issue (which works locally since I switched over to main as default branch).

I agree, it should not be hardcoded if you have a suggestion? :)

jglick:
https://github.com/jenkinsci/git-client-plugin/pull/733 looks like a bug fix to me; now that it is merged, no workarounds ought to be required in CI systems running it. I do not have any context about what https://github.com/jenkinsci/bitbucket-branch-source-plugin/pull/513 is doing or why.

lifeofguenter:
jenkinsci/bitbucket-branch-source-plugin#513 is renaming mentions of master to main as described here: https://sfconservancy.org/news/2020/jun/23/gitbranchname/

jglick:
> will find a different way to fix the tests

You mean some tests in that plugin are broken when using a different Git global configuration? If so, then definitely yes those tests should be fixed.

Can this then be closed?

dduportal:
Thanks for proposing this change. 

Agreed with @jglick , it is not clear what is the problem you are trying to solve in this PR. The associated issues/PRs on some Jenkins plugins are not helping to underline what is it you want to achieve.

I'm closing this PR, but feel free to open an issue describing the problem you face and the behavior not honored by the pipeline library if you have one in mind.

timja:
The issue is that git changes its default branch to main at some point and throws up a big warning message as well if you don't set a default branch.

This can cause impact on SCM based tests.

It broke a test in ATH.

I added a workaround there, and had to set it back to master.

It probably makes sense to explicitly set it, but I would say on the agents.

lifeofguenter:
Yup I'd say setting it in the agent would definitely be the way to go (in terms of inclusivity) - it might break some plugins tests today, but then they/we can fix those :)

dduportal:
Thanks @timja @lifeofguenter to explain it!

Sorry for asking but it was not clear at once (and IRC is not always easy to catch on subject).

I'll let you drive from here @timja (I'm reopening the PR) for the pipeline / tests part.

@lifeofguenter if you are ok, do you want to contribute on the agents? For the VMs:
- For Linux agents: https://github.com/jenkins-infra/packer-images/blob/main/scripts/ubuntu-20-provision.sh
- For Windows agents: https://github.com/jenkins-infra/packer-images/blob/main/scripts/windows-2019-provision.ps1

Ping me on IRC for any help!

MarkEWaite:
The Linux agents on ci.jenkins.io do not generally run a new enough version of command line git to be affected by the value of `init.defaultBranch`.  They are usually configured with the command line git version packaged by the operating system release.

Command line git versions did not begin honoring `init.defaultBranch` until [git 2.28](https://github.blog/2020-07-27-highlights-from-git-2-28/#introducing-init-defaultbranch) in July 2020.

jglick:
> throws up a big warning message as well if you don't set a default branch

So…a warning?

> It broke a test in ATH.
> 
> I added a workaround there

Can you link to details please? This still smells to me like a bad test that we should not be working around by configuring a CI environment in a special way.

dduportal:
> The Linux agents on ci.jenkins.io do not generally run a new enough version of command line git to be affected by the value of `init.defaultBranch`. They are usually configured with the command line git version packaged by the operating system release.
> 
> Windows agents generally run a new enough version of command line git to be affected by the value of `init.defaultBranch`.
> 
> Command line git versions did not begin honoring `init.defaultBranch` until [git 2.28](https://github.blog/2020-07-27-highlights-from-git-2-28/#introducing-init-defaultbranch) in July 2020.

@MarkEWaite this isn't true anymore for VM agents: since https://github.com/jenkins-infra/packer-images/pull/70, the git version is the same on both Windows and Linux, and defined in a centralized properties file at https://github.com/jenkins-infra/packer-images/blob/main/.auto.pkrvars.hcl#L2 (and it's automatically kept up to date as per https://github.com/jenkins-infra/packer-images/pull/109)

timja:
> Can you link to details please? This still smells to me like a bad test that we should not be working around by configuring a CI environment in a special way.

This test here:
https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/plugins/GitPluginTest.java#L190

Locally it failed because of master/main issues

I worked around it with https://github.com/jenkinsci/acceptance-test-harness/blob/3e2cc97f21e2bb22a5d5a110c7a43689442d7cae/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java#L60

I couldn't track down why it failed on CI even after fixing that

jglick:
Just looks like a bad test. https://github.com/jenkinsci/acceptance-test-harness/pull/705#discussion_r750652698 also. Better to leave it `@Ignore`d until someone can fix it properly.

dduportal (approved):
Doesn't bite so I'll proceed and merge it. Worst case, we revert it.

vars/buildPlugin.groovy
```diff
@@ -68,11 +69,23 @@ def call(Map params = [:]) {
...
                                 if (isUnix()) {
                                     sh(script: 'git clean -xffd > /dev/null 2>&1',
-                                       label:'Clean for incrementals',
+                                       label: 'Clean for incrementals',
                                        returnStatus: true) // Ignore failure if CLI git is not available
                                 } else {
                                     bat(script: 'git clean -xffd 1> nul 2>&1',
-                                        label:'Clean for incrementals',
+                                        label: 'Clean for incrementals',
+                                        returnStatus: true) // Ignore failure if CLI git is not available
```
jglick, line 77: BTW this is a poor idiom to copy. The Unix version should have been simply

```groovy
sh 'git clean -xffd > /dev/null 2>&1 || :'
```

or

```groovy
sh 'if which git; then git clean -xffd > /dev/null 2>&1; fi'
```

etc.

dduportal: @jglick Not sure about the idiom for the bat command line though, do you have some experience with it?

dduportal: (for info: https://github.com/jenkins-infra/pipeline-library/pull/344)

dduportal: self answer for bat: https://stackoverflow.com/questions/22046780/whats-the-windows-command-shell-equivalent-of-bashs-true-command

PR #255, opened 2021-11-29
feat: add a parameter to updatecli to specify the container memory amount

lemeurherve:


vars/updatecli.groovy
```diff
@@ -4,7 +4,8 @@ def call(userConfig = [:]) {
     action: 'diff',
     config: './updatecli/updatecli.d',
     values: './updatecli/values.yaml',
-    updatecliDockerImage: 'ghcr.io/updatecli/updatecli:v0.16.0',
+    updatecliDockerImage: 'ghcr.io/updatecli/updatecli:latest',
+    containerMemory: '128Mi'
```
timja, line 8: should we just increase it to what you need?

lemeurherve: Most of the time the default 128Mi value is enought, but for [querying nginx releases for example](https://github.com/jenkins-infra/docker-confluence-data/pull/13), we have to increase it. To avoid spending too much memory everywhere, I prefer to be able to change this value only when needed.

vars/updatecli.groovy
```diff
@@ -4,7 +4,8 @@ def call(userConfig = [:]) {
     action: 'diff',
     config: './updatecli/updatecli.d',
     values: './updatecli/values.yaml',
-    updatecliDockerImage: 'ghcr.io/updatecli/updatecli:v0.16.0',
+    updatecliDockerImage: 'ghcr.io/updatecli/updatecli:latest',
```
dduportal, line 7: @lemeurherve can you make sure that your branch is rebased against the master branch?

dduportal (changes requested):
Should be rebased

dduportal (approved):
Looks good to me !

PR #260, opened 2021-12-07
Variable was used before it was defined

basil:
Amends #232. CC @dduportal @timja


MarkEWaite (approved):

dduportal (approved):
Good catch 🦅 👁️ !

Thanks!

basil:
> Good catch eagle eye !

They don't call me "Crow Eyes" for nothing. :smile:

PR #263, opened 2021-12-10
Show color during CI build

basil:
I realize this is purely subjective/superficial, but I think this would be a welcome change for those looking at CI build failures.


timja (approved):
sure 🤷

basil:
Tested in jenkinsci/jenkins#6045

jglick (approved):
Great if it works.

basil:
Back in draft state since apparently [`ansicolor` is broken](https://github.com/jenkinsci/jenkins/pull/6045#issuecomment-991219860).

dduportal:
Hello @basil , do you feel that we should update and try again this PR?

Your initial request feels legit (and useful) but I'm not sure how to evaluate if ansicolor is still broken without disturbing contributors.

basil:
The `ansicolor` plugin is not being actively developed and unfortunately I do not believe the regression has been fixed.

dduportal:
> The `ansicolor` plugin is not being actively developed and unfortunately I do not believe the regression has been fixed.

Many thanks for the feedback

PR #299, opened 2022-02-07
Revert "Bump docker-builder version on shared library resources to 2.0.9"

dduportal:
Reverts jenkins-infra/pipeline-library#295 because https://github.com/jenkins-infra/helpdesk/issues/2190

It also stops tracking (with updatecli) the version of `docker-builder` to avoid repeating the mistake.
Not an issue to stop updating this dependency as we need to switch from `img` to `docker` for the library in the future?


dduportal:
Force merging as admin to unblock the failing Docker builds on infra.ci.

PR #302, opened 2022-02-10
fix(buildPlugin) handle container nodes with JDK > 11

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/2775

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -41,7 +41,10 @@ def call(Map params = [:]) {
         boolean addToolEnv = !useContainerAgent
 
         if(useContainerAgent && (label == 'linux' || label == 'windows')) {
-            String agentContainerLabel = jdk == '8' ? 'maven' : 'maven-11'
+            String agentContainerLabel = 'maven'
+            if (jdk != '8') {
+              agentContainerLabel += "-${jdk}"
+            }
             if(label == 'windows') {
                 agentContainerLabel += "-windows"
             }


basil (approved):
cf. https://github.com/jenkinsci/jenkins/blob/fb6f28bc023a3f1ab582cea355f9350893723498/Jenkinsfile#L35

basil:
There was nothing wrong with the first revision, but the second revision is wrong to do `jdk == 8`, since `jdk` here is a string and not an integer. It should be `jdk == '8'` (unlike the version in core, where it is an integer).

basil (approved):

lemeurherve (approved):

vars/buildPlugin.groovy
```diff
@@ -41,7 +41,7 @@ def call(Map params = [:]) {
         boolean addToolEnv = !useContainerAgent
 
         if(useContainerAgent && (label == 'linux' || label == 'windows')) {
-            String agentContainerLabel = jdk == '8' ? 'maven' : 'maven-11'
+            def agentContainerLabel = jdk == '8' ? 'maven' : 'maven-' + jdk
```
jglick, line 44: or

```groovy
            def agentContainerLabel = jdk == '8' ? 'maven' : "maven-$jdk"
```

jglick (commented):
Maybe we should retire `maven` and make it `maven-8`?

vars/buildPlugin.groovy
dduportal: 🤔 what are the pro and cons of these 2 concatenation methods (in this context, e.g. pipeline library)?

jglick: None, purely stylistic.

basil: In pure Groovy there is no difference, but in Pipeline the latter is subject to `InterpolatedSecretsAction` while the former is not (see #210).

jglick: Oh I forgot about https://github.com/jenkins-infra/pipeline-library/pull/210#issuecomment-854852071

PR #307, opened 2022-02-24
avoid deep pattern for maven-invoker-plugin generated junit files

olamy:


dduportal:
Hello 👋 Could you explain the why/what of this change? It's not clear what problem it is solving?

vars/buildPlugin.groovy
```diff
@@ -115,7 +115,7 @@ def call(Map params = [:]) {
                                     infra.runMaven(mavenOptions, jdk, null, null, addToolEnv)
                                 } finally {
                                     if (!skipTests) {
-                                        junit('**/target/surefire-reports/**/*.xml,**/target/failsafe-reports/**/*.xml,**/target/invoker-reports/**/*.xml')
```
dduportal, line 118: If I understand correctly this change, the junit reports that would be in nested directories (for instance `target/surefire-reports/foo/test-1.xml`) would not be selected anymore.

Am I correct?

dduportal:
`java.lang.AssertionError
	at BuildPluginStepTests.test_buildPlugin_with_defaults(BuildPluginStepTests.groovy:130)`

do not forget to update the unit test ;)

timja:
> Hello 👋 Could you explain the why/what of this change? It's not clear what problem it is solving?

Context: https://github.com/jenkinsci/plugin-pom/pull/508

vars/buildPlugin.groovy
timja: I don't think they are actually nested

jglick (commented):
While I think this change is correct, I doubt it will solve the problem in https://github.com/jenkinsci/plugin-pom/pull/508 which IIUC was not about `target/surefire-reports/some/deep/subdir/TEST-Stuff.xml` but about `some/deep/subdir/target/surefire-reports/TEST-Stuff.xml`. And switching to for example `target/surefire-reports/*.xml` would cause us to lose legitimate test results from multimodule repos.

vars/buildPlugin.groovy
olamy: surefire xml files are only generated at the top level directory.

olamy:
> While I think this change is correct, I doubt it will solve the problem in [jenkinsci/plugin-pom#508](https://github.com/jenkinsci/plugin-pom/pull/508) which IIUC was not about `target/surefire-reports/some/deep/subdir/TEST-Stuff.xml` but about `some/deep/subdir/target/surefire-reports/TEST-Stuff.xml`. And switching to for example `target/surefire-reports/*.xml` would cause us to lose legitimate test results from multimodule repos.

the new pattern `**/target/surefire-reports/*.xml`  still starts with `**` so that should not be a problem with multimodule repos.

jglick:
Right, so (AFAICT) this change should not _break_ anything, but neither is it likely to _fix_ the problem in https://github.com/jenkinsci/plugin-pom/pull/508.

olamy:
Even if those deep scanning could be avoided, the PR is not needed anymore.

jglick:
https://github.com/jenkinsci/plugin-pom/pull/508#issuecomment-1051313886: maybe would be appropriate to look only one level deep?

PR #311, opened 2022-03-02
feat(buildDockerAndPublishImage) allow building on VM agent with Docker Engine

dduportal:
This PR allows executing the library `buildDockerAndPublishImage()` with Docker Engine inside a VM (instead of the container-less tool `img` inside Kuberneets Pod).

The default behavior is kept to use `img` to avoid breaking everything: let's use this on a set of predefined docker images for 1-2 weeks, and if no major issue arise, then let's switch default to Docker.

Tested on a real life example in https://github.com/jenkins-infra/docker-helmfile/pull/66 .

It allows to build images based on Alpine 3.14+ that require using a recent container engine with the correct setups.

Please note that the tests run in CST are faster when using the Docker driver instead of the tar driver, and are able to run "docker commands" (improved test harnesses).

PR #313, opened 2022-03-04
fix(buildDockerAndPublishImage) add missing tools for releases

dduportal:
This PR introduces a set of fixes to the `buildDockerAndPublishImage()` library, following #311 

- Fix a bug at the `lint` phase which was using (unexpectedly) `docker run` to execute hadolint when running in docker-less environment (was blocking the last build of `docker-openvpn`)
- Fix a bug where automatic tagging and automatic releases were failing in VM environments by installing the missing tools  (`gh` and `jx-release-version`)
- Fix a bug where release builds weren't failing when the GH release wasn't found (or tooling not installed)


lemeurherve (approved):
🎉

PR #318, opened 2022-03-08
[WIP] feat: add getGithubAppAccessToken

lemeurherve:


timja (commented):
why aren't you just using a Jenkins credentials?

just need to use a `withCredentials(...)` and you'll get an access token

lemeurherve:
> just need to use a withCredentials(...) and you'll get an access token

A Personal Acces Token for sure, but can I get an Installation Access Token linked to a Github App?

dduportal:
It is the same idea as https://github.com/tibdex/github-app-token

timja:
> > just need to use a withCredentials(...) and you'll get an access token
> 
> A Personal Acces Token for sure, but can I get an Installation Access Token linked to a Github App?

yes it does that

dduportal:
> > > just need to use a withCredentials(...) and you'll get an access token
> > 
> > 
> > A Personal Acces Token for sure, but can I get an Installation Access Token linked to a Github App?
> 
> yes it does that
> 
> https://www.jenkins.io/blog/2020/04/16/github-app-authentication/#how-do-i-get-an-api-token-in-my-pipeline

ah ah you are cheating: there is this big "curl" :D (but the idea here is to abstract this curl request into a shared library)

PR #320, opened 2022-03-10
feat: use a githubApp credentialsId for updatecli shared pipeline

lemeurherve:
Depends on https://github.com/jenkins-infra/kubernetes-management/pull/2083
Part of https://github.com/jenkins-infra/helpdesk/issues/2817

### DIFF
--- a/test/groovy/ParallelDockerUpdatecliStepTests.groovy
+++ b/test/groovy/ParallelDockerUpdatecliStepTests.groovy
@@ -11,8 +11,8 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
   static final String testImageName = 'myImage'
   static final String anotherCronTriggerExpression = '@daily'
   static final String anotherContainerMemory = '345Mi' // different than the default value specified in ${scriptName}
-  static final String defaultCredentialsId = 'github-app-updatecli-on-jenkins-infra'
-  static final String defaultDockerGitCredentialsId = 'github-docker-github-app-infra'
+  static final String defaultUpdatecliCredentialsId = 'github-app-updatecli-on-jenkins-infra'
+  static final String defaultDockerGitCredentialsId = 'github-app-infra'
   static final String anotherCredentialsId = 'another-github-token'
 
   @Override
@@ -153,7 +153,6 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
     assertFalse(assertMethodCallContainsPattern('echo', 'WARNING:'))
   }
 
-
   @Test
   void itRunsSuccessfullyWithCustomParameters() throws Exception {
     def script = loadScript(scriptName)
@@ -170,7 +169,7 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
       buildDockerConfig: [
         useContainer: false,
       ],
-      credentialsId: anotherCredentialsId
+      updatecliCredentialsId: anotherCredentialsId
     )
     printCallStack()
 
@@ -196,7 +195,7 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
 
     // And the custom parameters are taken in account for docker image build
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', "gitCredentials=${defaultDockerGitCredentialsId}"))
-    assertTrue(assertMethodCallContainsPattern('string', anotherCredentialsId))
+    assertTrue(assertMethodCallContainsPattern('updatecli', "credentialsId=${anotherCredentialsId}"))
 
     // And the method "updatecli()" is called for "diff" and "apply" actions (both with the same custom parameters)
     assertTrue(assertMethodCallOccurrences('updatecli', 2))
--- a/test/groovy/ParallelDockerUpdatecliStepTests.groovy
+++ b/test/groovy/ParallelDockerUpdatecliStepTests.groovy
@@ -195,6 +195,7 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
     assertFalse(assertMethodCallContainsPattern('echo', 'WARNING:'))
 
     // And the custom parameters are taken in account for docker image build
+    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', "gitCredentials=${defaultDockerGitCredentialsId}"))
     assertTrue(assertMethodCallContainsPattern('string', anotherCredentialsId))
 
     // And the method "updatecli()" is called for "diff" and "apply" actions (both with the same custom parameters)
--- a/test/groovy/ParallelDockerUpdatecliStepTests.groovy
+++ b/test/groovy/ParallelDockerUpdatecliStepTests.groovy
@@ -183,7 +183,7 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', testImageName))
     // And the correct fixed settings for buildDockerAndPublishImage
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', 'automaticSemanticVersioning=true'))
-    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', "gitCredentials=${defaultCredentialsId}"))
+    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', "gitCredentials=${defaultDockerGitCredentialsId}"))
     // And the custom settings for buildDockerAndPublishImage
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', 'useContainer=false'))
 
--- a/test/groovy/ParallelDockerUpdatecliStepTests.groovy
+++ b/test/groovy/ParallelDockerUpdatecliStepTests.groovy
@@ -11,6 +11,7 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
   static final String testImageName = 'myImage'
   static final String anotherCronTriggerExpression = '@daily'
   static final String anotherContainerMemory = '345Mi' // different than the default value specified in ${scriptName}
+  static final String defaultCredentialsId = 'github-app-updatecli-on-jenkins-infra'
   static final String anotherCredentialsId = 'another-github-token'
 
   @Override
@@ -63,7 +64,7 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', testImageName))
     // And the correct settings
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', 'automaticSemanticVersioning=true'))
-    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', 'gitCredentials=github-app-infra'))
+    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', "gitCredentials=${defaultCredentialsId}"))
 
     // And updatecli(action: 'diff') is called
     assertTrue(assertMethodCallContainsPattern('updatecli', 'action=diff'))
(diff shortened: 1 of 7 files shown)


vars/updatecli.groovy
```diff
@@ -53,10 +54,16 @@ def call(userConfig = [:]) {
       }
       stage(updatecliRunStage) {
         if (runUpdatecli) {
-          sh 'updatecli version'
-          sh updatecliCommand
-        }
-      }// stage
+          withCredentials([usernamePassword(
+            credentialsId: finalConfig.credentialsId,
+            usernameVariable: 'USERNAME_NOT_USED', // Setting this variable is mandatory, even if of no use
```
dduportal, line 59:
```suggestion
            usernameVariable: 'USERNAME_NOT_USED', // Setting this variable is mandatory, even if of not used
```

vars/updatecli.txt
```diff
@@ -11,6 +11,7 @@
       <li>String updatecliDockerImage: (Optional - Default: "jenkinsciinfra/helmfile:2.3.0") Docker Image of updatecli to be used in the process.</li>
       <li>String cronTriggerExpression: (Optional - Default: "") Enable periodic execution by providing a cron-like expression.</li>
       <li>String containerMemory: (Optional - Default: "512Mi") specify the amount of memory dedicated to the updatecli container.</li>
+      <li>String credentialsId: (Optional - Default: "github-app-updatecli-on-jenkins-infra") specify the githubApp credentials id to use to get an Access Token.</li>
```
dduportal, line 14:
```suggestion
      <li>String credentialsId: (Optional - Default: "github-app-updatecli-on-jenkins-infra") specify the githubApp credentials id to use to get an access token. The environment variable `UPDATECLI_GITHUB_TOKEN` holds the token value to be used within updatecli.</li>
```

vars/updatecli.txt
```diff
@@ -32,8 +33,16 @@
...
 
+    // Note: to get an access token from a Github App instead of a Github user, you'll have to use this as credentials:
+    withCredentials([usernamePassword(
+      credentialsId: 'github-app-updatecli-on-jenkins-infra',
+      usernameVariable: 'USERNAME_NOT_USED', // Setting this variable is mandatory, even if of no use
+      passwordVariable: 'UPDATECLI_GITHUB_TOKEN'
+    )]) {
+      // ...
+    }
+    </code></pre>
```
dduportal, line 45: Note sure what is the intent of this block? Since the default behavior is to use a github app, I don't udnerstand the comment?

vars/parallelDockerUpdatecli.groovy
```diff
@@ -35,16 +35,14 @@ def call(userConfig = [:]) {
...
-      withCredentials([string(credentialsId: finalConfig.credentialsId,variable: 'UPDATECLI_GITHUB_TOKEN')]) {
-        // Merging the 2 maps - https://blog.mrhaki.com/2010/04/groovy-goodness-adding-maps-to-map_21.html
-        Map updatecliConfig = finalConfig.updatecliConfig << [action: 'diff']
+      // Merging the 2 maps - https://blog.mrhaki.com/2010/04/groovy-goodness-adding-maps-to-map_21.html
+      Map updatecliConfig = finalConfig.updatecliConfig << [action: 'diff', credentialsId: finalConfig.credentialsId]
 
+      updatecli(updatecliConfig)
+      if (env.BRANCH_IS_PRIMARY) {
+        // Merging the 2 maps - https://blog.mrhaki.com/2010/04/groovy-goodness-adding-maps-to-map_21.html
+        updatecliConfig = finalConfig.updatecliConfig << [action: 'apply', cronTriggerExpression: finalConfig.updatecliApplyCronTriggerExpression]
```
dduportal, line 44:
```suggestion
        updatecliConfig = finalConfig.updatecliConfig << [action: 'apply', cronTriggerExpression: finalConfig.updatecliApplyCronTriggerExpression, credentialsId: finalConfig.credentialsId]
```

test/groovy/UpdatecliStepTests.groovy
```diff
@@ -129,4 +130,19 @@ class UpdatecliStepTests extends BaseTest {
...
+
+    // when calling the "updatecli" function with a custom credentialsId
+    script.call(credentialsId: anotherCredentialsId)
+    printCallStack()
+
+    // Then we expect a successful build
+    assertJobStatusSuccess()
+
+    // And the custom credentialsId is taken in account
+    assertTrue(assertMethodCallContainsPattern('string', anotherCredentialsId))
```
dduportal, line 146:
```suggestion
    assertTrue(assertMethodCallContainsPattern('withCredentials', anotherCredentialsId))
```

dduportal (changes requested):
Great job! A few elements (some are nitpicks, other are blocking :) )

vars/updatecli.txt
lemeurherve: I wouldn't add this as this env var is declared only in the code below, while this pipeline parameter doesn't say anything about it.

vars/updatecli.txt
lemeurherve: You might want to use the withCredential above with a Personal Access Token, while this withCredentials syntax is for using a Github App Installation Access Token (not the same parameters used)

vars/parallelDockerUpdatecli.groovy
lemeurherve: This config part is already included in finalConfig L39, no need to include it again.

lemeurherve: Or do I miss something?

vars/updatecli.txt
dduportal: Since the library code is now specifying the name of the env. var (https://github.com/jenkins-infra/pipeline-library/pull/320/files#diff-01eacbe5caf24c385fc8ab8a7684a5580faedb7d729aa4af4d9c3640cf9a4841R60) as part of your change, as a end user, I expect the variable to be documented (however we might have forgotten to document before in parallelDockerUpdatecli). But it's how the end user should expect the token to be used when writing their values.yaml

vars/updatecli.txt
dduportal: Then maybe inverting the comment then? Because it tells the opposite of what you are saying 😅

vars/updatecli.txt
lemeurherve, line 14:
```suggestion
      <li>String credentialsId: (Optional - Default: "github-app-updatecli-on-jenkins-infra") specify the githubApp credentials id to use to get an access token. The environment variable `UPDATECLI_GITHUB_TOKEN` holds the token value to be used within updatecli, be sure it's the name you defined as `token` value in your updatecli/values.yaml</li>
```

lemeurherve: @dduportal WDYT?

lemeurherve: Good point! See https://github.com/jenkins-infra/pipeline-library/pull/320#discussion_r823708232 for a more complete suggestion then :)

vars/updatecli.txt
lemeurherve: > You might want to use the withCredential above
... above this one, which is for a Github App

More clear here?

vars/parallelDockerUpdatecli.groovy
dduportal: I confirm that it is not part of final config. 

- [L39](https://github.com/jenkins-infra/pipeline-library/pull/320/files#diff-fd1e84d84679aaa1bc63579323628681c111e48b2aa5a90c765a7f153cdcff79R39) defines a new variable named `updatecliConfig` of type `Map` which is the result of merging finalConfig with the map `[action: 'diff', credentialsId: finalConfig.credentialsId]`.
- The current line pointed by this thread assign a new value to the variable `updatecliConfig`, which is the merge of `finalConfig` with  `[action: 'apply', cronTriggerExpression: finalConfig.updatecliApplyCronTriggerExpression, credentialsId: finalConfig.credentialsId]`. But it's a different one.

vars/updatecli.txt
lemeurherve: Oh.. @dduportal can you tell me if we have access to udpatecli/values.yaml data here? If it's the case, I would use the user value instead of hardcoding `UPDATECLI_GITHUB_TOKEN` in the shared pipeline.

dduportal: you have access through the workspace (you can run an `sh` pipeline command and get its output to a variable, or through a groovy method that would read the file from workspace on the agent).

However it's really hard to do: you don't know what keys were used by the end user. It could be directly in the manifest also. That's why I find your current code more elegant as soon as it it documented to end user:
- "If you specify a githubapp credential id, then the variable `UPDATECLI_GITHUB_TOKEN` is set to the token.
- "If you want another variable, then wrap your call to the library `updatecli(...)` inside your own `withEnvironment` and voila"

=> But there is a corner case: if the end user does not have the credential named `github-app-updatecli-on-jenkins-infra` AND does not need to specify another credential. Not probable in our case but that was the reason to delegate credential out of the library itself.

vars/parallelDockerUpdatecli.groovy
lemeurherve: My bad 😞

vars/updatecli.txt
lemeurherve: > => But there is a corner case: if the end user does not have the credential named github-app-updatecli-on-jenkins-infra AND does not need to specify another credential. 

There is at least another shared pipeline with a default credentialsId value, buildDockerAndPublishImage (which should be checked/updated too in another PR)

test/groovy/ParallelDockerUpdatecliStepTests.groovy
```diff
@@ -63,7 +64,7 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', testImageName))
     // And the correct settings
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', 'automaticSemanticVersioning=true'))
-    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', 'gitCredentials=github-app-infra'))
+    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', "gitCredentials=${defaultCredentialsId}"))
```
dduportal, line 67:
```suggestion
    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', "gitCredentials=${defaultDockerGitCredentialsId}"))
```

test/groovy/ParallelDockerUpdatecliStepTests.groovy
```diff
@@ -182,7 +183,7 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', testImageName))
     // And the correct fixed settings for buildDockerAndPublishImage
     assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', 'automaticSemanticVersioning=true'))
-    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', 'gitCredentials=github-app-infra'))
+    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', "gitCredentials=${defaultCredentialsId}"))
```
dduportal, line 186:
```suggestion
    assertTrue(assertMethodCallContainsPattern('buildDockerAndPublishImage', "gitCredentials=${defaultDockerGitCredentialsId}"))
```

test/groovy/ParallelDockerUpdatecliStepTests.groovy
```diff
@@ -11,6 +11,7 @@ class ParallelDockerUpdatecliStepTests extends BaseTest {
   static final String testImageName = 'myImage'
   static final String anotherCronTriggerExpression = '@daily'
   static final String anotherContainerMemory = '345Mi' // different than the default value specified in ${scriptName}
+  static final String defaultCredentialsId = 'github-app-updatecli-on-jenkins-infra'
```
dduportal, line 14:
```suggestion
  static final String defaultCredentialsId = 'github-app-updatecli-on-jenkins-infra'
  static final String defaultDockerGitCredentialsId='github-docker-github-app-infra'
```

dduportal (changes requested):
Sounds like the Dockerbuild and updatecli have 2 different GitHub credentials.

That looks good to me (different github apps with different rights) but you need to define both default values for these credentials in the test

dduportal (approved):

PR #328, opened 2022-03-15
fix: prioritize updatecliConfig block values over default values

lemeurherve:
Follow-up of https://github.com/jenkins-infra/pipeline-library/pull/320


dduportal (approved):
LGTM

PR #334, opened 2022-03-24
fix: enable long paths for git to avoid problems with tests on Windows agents

lemeurherve:
To solve https://github.com/jenkins-infra/helpdesk/issues/2847 in the short term, here is a proposition to add the git config command to the pipeline library, called [here](https://github.com/jenkins-infra/pipeline-library/blob/216eee69ae367cf11871013fb52b0dfae61b73ee/vars/buildPlugin.groovy#L64) when using the `buildPlugins` shared pipeline.

I'll add this command to the agents in following PRs.

cc @jtnord

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -35,7 +35,7 @@ Object checkout(String repo = null) {
 
 Object checkoutSCM(String repo = null) {
     // Enable long paths to avoid problems with tests on Windows agents 
-    git config core.longpaths true
+    sh 'git config core.longpaths true'
 
     if (env.BRANCH_NAME) {
         checkout scm
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -34,6 +34,9 @@ Object checkout(String repo = null) {
 }
 
 Object checkoutSCM(String repo = null) {
+    // Enable long paths to avoid problems with tests on Windows agents 
+    git config core.longpaths true
+
     if (env.BRANCH_NAME) {
         checkout scm
     } else if ((env.BRANCH_NAME == null) && (repo)) {
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -36,7 +36,7 @@ Object checkout(String repo = null) {
 Object checkoutSCM(String repo = null) {
     // Enable long paths to avoid problems with tests on Windows agents 
     if (!isUnix()) {
-        bat 'git config core.longpaths true'
+        bat 'git config --global core.longpaths true'
     }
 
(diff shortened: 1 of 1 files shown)


vars/infra.groovy
```diff
@@ -34,6 +34,9 @@ Object checkout(String repo = null) {
 }
 
 Object checkoutSCM(String repo = null) {
+    // Enable long paths to avoid problems with tests on Windows agents 
+    git config core.longpaths true
```
timja, line 38:
```suggestion
    sh 'git config core.longpaths true'
```

vars/infra.groovy
```diff
@@ -34,6 +34,9 @@ Object checkout(String repo = null) {
 }
 
 Object checkoutSCM(String repo = null) {
+    // Enable long paths to avoid problems with tests on Windows agents 
+    sh 'git config core.longpaths true'
```
lemeurherve, line 38:
```suggestion
    if (!isUnix()) {
        bat 'git config core.longpaths true'
    }
```

vars/infra.groovy
```diff
@@ -34,6 +34,11 @@ Object checkout(String repo = null) {
 }
 
 Object checkoutSCM(String repo = null) {
+    // Enable long paths to avoid problems with tests on Windows agents 
+    if (!isUnix()) {
+        bat 'git config core.longpaths true'
```
jtnord, line 39:
```suggestion
        bat 'git config --global core.longpaths true'
```

because we want this to apply to all git operations not just the checkout of this source (e.g. many tests amy use the git plugins tests that create new repos for tests)

lemeurherve: From what I've understood, setting it globally requires system permissions, I'm not sure the pipeline have them here.

lemeurherve: In the long term, I want to set this setting directly in the agents definitions.

jtnord: setting `system` requires admin permissions, `global` is per user.
Whilst you can lock down user access from their own homes (I have done this in the past for Jeknins installs) ci.jenkins.io does not do this as demonstrated by [this change](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/502/files#diff-9db55821d407770df4fc37f8dcbb12632e46316fbb4dcba2863e54b87533860dR165) which passes

lemeurherve: Perfect, thanks for the explanation and example!

jtnord (approved):

PR #335, opened 2022-03-25
fix: stack overflow

smerle33:
correct a parameter for the unit test to go through when checkout from a branch.

### DIFF
--- a/test/groovy/InfraStepTests.groovy
+++ b/test/groovy/InfraStepTests.groovy
@@ -58,7 +58,7 @@ class InfraStepTests extends BaseTest {
   }
 
   @Test
-  @Ignore("Some stackoverflow issues")
+  //@Ignore("Some stackoverflow issues")
   void testCheckoutWithEnvVariable() throws Exception {
     def script = loadScript(scriptName)
     env.BRANCH_NAME = 'BRANCH'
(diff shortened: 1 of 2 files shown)


lemeurherve (approved):

vars/infra.groovy
```diff
@@ -40,7 +40,7 @@ Object checkoutSCM(String repo = null) {
     }
 
     if (env.BRANCH_NAME) {
-        checkout scm
+        checkout repo
```
timja, line 43: `checkout scm` is the way of doing this, have you got a link for `checkout repo` I've never heard of it

dduportal, line 43:
```suggestion
        checkout scm
```

dduportal (changes requested):
+ 1 with Tim remark.

smerle33:
@timja what do you think of removing the checkout method that is generating the stackoverflow from within the unit test ? I think that it should not have the same name as "checkout" ....

test/groovy/InfraStepTests.groovy
```diff
@@ -58,7 +58,7 @@ class InfraStepTests extends BaseTest {
   }
 
   @Test
-  @Ignore("Some stackoverflow issues")
+  //@Ignore("Some stackoverflow issues")
```
dduportal, line 61:
```suggestion
```

timja:
> @timja what do you think of removing the checkout method that is generating the stackoverflow from within the unit test ? I think that it should not have the same name as "checkout" ....

I think okay, seems to just be a legacy method delegating to checkoutSCM

dduportal (approved):

timja (approved):

PR #343, opened 2022-03-31
fix(buildDockerAndPublishImage) always fetch the latest git tags to avoid dependency on job configuration

dduportal:
While working on https://github.com/jenkins-infra/helpdesk/issues/2840, a new job configuration was applied on infra.ci.jenkins.io.

It made clear that the Jenkins job behavior of discovering git tags is not always set up (or subject to bug?).

This PR fixes the docker shared library by always fetching the tags from the origin git repository.

The overhead of this operation is ~1s (even with a lot of tags) so the impact should be minimum. 

Tested with success on 2 builds:
- https://infra.ci.jenkins.io/job/docker-jobs/job/docker-jenkins-lts/job/main/3/console
- https://infra.ci.jenkins.io/job/docker-jobs/job/docker-jenkins-lts/job/main/4/console

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -73,8 +73,8 @@ def call(String imageName, Map userConfig=[:]) {
               '''
             }
           }
-
-          nextVersion = sh(script:"${finalConfig.nextVersionCommand}", returnStdout: true).trim()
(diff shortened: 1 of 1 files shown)


pilere (approved):
:rocket

smerle33 (approved):
🚀

PR #344, opened 2022-03-31
chore(buildPlugin) improve idiomatic way to call git command in 'never-fail' mode

dduportal:
Ref.  https://github.com/jenkins-infra/pipeline-library/pull/247#pullrequestreview-927497269

For bat commands: https://stackoverflow.com/questions/22046780/whats-the-windows-command-shell-equivalent-of-bashs-true-command
(TL;DR; add `|| VER >NUL` that keeps printing the error message but set the exit code to 0)

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -67,19 +67,23 @@ def call(Map params = [:]) {
                             incrementals = fileExists('.mvn/extensions.xml') &&
                                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
                             final String gitUnavailableMessage = "[buildPlugin] Git CLI may not be available"
-                            if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
-                                if (isUnix()) {
-                                    sh 'git clean -xffd || echo "${gitUnavailableMessage}"'
-                                } else {
-                                    bat 'git clean -xffd || echo "${gitUnavailableMessage}"'
+                            withEnv(["GITUNAVAILABLEMESSAGE=${gitUnavailableMessage}"]) {
+                                if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
+                                    if (isUnix()) {
+                                        sh 'git clean -xffd || echo "$GITUNAVAILABLEMESSAGE"'
+                                    } else {
+                                        bat 'git clean -xffd || echo %GITUNAVAILABLEMESSAGE%'
+                                    }
                                 }
-                            }
 
-                            if (gitDefaultBranch) {
-                                if (isUnix()) {
-                                    sh "git config --global init.defaultBranch '${gitDefaultBranch}' || echo '${gitUnavailableMessage}'"
-                                } else {
-                                    bat "git config --global init.defaultBranch '${gitDefaultBranch}' || echo '${gitUnavailableMessage}'"
+                                if (gitDefaultBranch) {
+                                    withEnv(["GITDEFAULTBRANCH=${gitDefaultBranch}"]) {
+                                        if (isUnix()) {
+                                            sh 'git config --global init.defaultBranch "$GITDEFAULTBRANCH" || echo "$GITDEFAULTBRANCH"'
+                                        } else {
+                                            bat 'git config --global init.defaultBranch %GITDEFAULTBRANCH% || echo %GITDEFAULTBRANCH%'
+                                        }
+                                    }
                                 }
                             }
                         }
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -66,27 +66,20 @@ def call(Map params = [:]) {
                             isMaven = fileExists('pom.xml')
                             incrementals = fileExists('.mvn/extensions.xml') &&
                                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
+                            final String gitUnavailableMessage = "[buildPlugin] Git CLI may not be available"
                             if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
                                 if (isUnix()) {
-                                    sh(script: 'git clean -xffd > /dev/null 2>&1',
-                                       label: 'Clean for incrementals',
-                                       returnStatus: true) // Ignore failure if CLI git is not available
+                                    sh 'git clean -xffd || echo "${gitUnavailableMessage}"'
                                 } else {
-                                    bat(script: 'git clean -xffd 1> nul 2>&1',
-                                        label: 'Clean for incrementals',
-                                        returnStatus: true) // Ignore failure if CLI git is not available
+                                    bat 'git clean -xffd || echo "${gitUnavailableMessage}"'
                                 }
                             }
 
                             if (gitDefaultBranch) {
                                 if (isUnix()) {
-                                    sh(script: "git config --global init.defaultBranch '${gitDefaultBranch}'",
-                                       label: 'Set default branch for git',
-                                       returnStatus: true) // Ignore failure if CLI git is not available
+                                    sh "git config --global init.defaultBranch '${gitDefaultBranch}' || echo '${gitUnavailableMessage}'"
                                 } else {
-                                    bat(script: "git config --global init.defaultBranch '${gitDefaultBranch}'",
-                                        label: 'Set default branch for git',
-                                        returnStatus: true) // Ignore failure if CLI git is not available
+                                    bat "git config --global init.defaultBranch '${gitDefaultBranch}' || echo '${gitUnavailableMessage}'"
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


vars/buildPlugin.groovy
```diff
@@ -68,7 +68,7 @@ def call(Map params = [:]) {
                                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
                             if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
                                 if (isUnix()) {
-                                    sh 'git clean -xffd > /dev/null 2>&1 || :', // Ignore failure if CLI git is not available
+                                      sh 'git clean -xffd > /dev/null 2>&1 || :' // Ignore failure if CLI git is not available
```
lemeurherve, line 71:
```suggestion
                                    sh 'git clean -xffd > /dev/null 2>&1 || :' // Ignore failure if CLI git is not available
```

lemeurherve (approved):

vars/buildPlugin.groovy
```diff
@@ -68,25 +68,17 @@ def call(Map params = [:]) {
                                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
                             if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
                                 if (isUnix()) {
-                                    sh(script: 'git clean -xffd > /dev/null 2>&1',
-                                       label: 'Clean for incrementals',
-                                       returnStatus: true) // Ignore failure if CLI git is not available
+                                    sh 'git clean -xffd > /dev/null 2>&1 || :' // Ignore failure if CLI git is not available
```
MarkEWaite, line 71: If the git command is not available, this will report the 'git not found' message in the log file without failing the job.  I think that is better than not reporting the 'git not found' message and is consistent with the later use in this pull request.
```suggestion
                                    sh 'git clean -xffd || :' // Ignore failure if CLI git is not available
```

vars/buildPlugin.groovy
```diff
@@ -68,25 +68,17 @@ def call(Map params = [:]) {
...
                                 } else {
-                                    bat(script: 'git clean -xffd 1> nul 2>&1',
-                                        label: 'Clean for incrementals',
-                                        returnStatus: true) // Ignore failure if CLI git is not available
+                                    bat 'git clean -xffd 1> nul 2>&1 || VER>NUL' // Ignore failure if CLI git is not available
```
MarkEWaite, line 73: If the git command is not available, this will report the 'git not found' message in the log file without failing the job. I think that is better than not reporting the 'git not found' message and is consistent with the later use in this pull request.
```suggestion
                                    bat 'git clean -xffd || VER>NUL' // Ignore failure if CLI git is not available
```

MarkEWaite (approved):
Accepted as is, since the goal will be achieved with the code as proposed.

Would like you to consider changing the `git clean` command invocation so that it uses the same syntax as the `git config` command because I prefer the behavior of the second form in the case of a missing git executable.

dduportal:
> Accepted as is, since the goal will be achieved with the code as proposed.
> 
> Would like you to consider changing the `git clean` command invocation so that it uses the same syntax as the `git config` command because I prefer the behavior of the second form in the case of a missing git executable.

Your suggestion totally make sense. Since the scope of this PR is to apply good idiomatics, it's really a nice suggestion!

vars/buildPlugin.groovy
```diff
@@ -68,25 +68,17 @@ def call(Map params = [:]) {
                                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
                             if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
                                 if (isUnix()) {
-                                    sh(script: 'git clean -xffd > /dev/null 2>&1',
-                                       label: 'Clean for incrementals',
-                                       returnStatus: true) // Ignore failure if CLI git is not available
+                                    sh 'git clean -xffd || :' // Ignore failure if CLI git is not available
```
jglick, line 71: or self-documenting
```suggestion
                                    sh 'git clean -xffd || echo "Git CLI may not be available"'
```

jglick (approved):

vars/buildPlugin.groovy
```diff
@@ -68,25 +68,17 @@ def call(Map params = [:]) {
...
+                                    bat 'git clean -xffd || VER>NUL' // Ignore failure if CLI git is not available
                                 }
                             }
 
                             if (gitDefaultBranch) {
                                 if (isUnix()) {
-                                    sh(script: "git config --global init.defaultBranch '${gitDefaultBranch}'",
-                                       label: 'Set default branch for git',
-                                       returnStatus: true) // Ignore failure if CLI git is not available
+                                    sh "git config --global init.defaultBranch '${gitDefaultBranch}' || :" // Ignore failure if CLI git is not available
```
MarkEWaite, line 79: Seems like you'll want the same message if this command fails as if the `git clean` fails.
```suggestion
                                    sh "git config --global init.defaultBranch '${gitDefaultBranch}' || echo Git CLI may not be available" // Ignore failure if CLI git is not available
```

dduportal: Thanks! I was planning to find « right way » to do it in windows as well, and got carried away with the CVE so put it back to draft. The suggestions helps and come handy!

dduportal: Updated WDYT?

vars/buildPlugin.groovy
```diff
@@ -66,27 +66,20 @@ def call(Map params = [:]) {
                             isMaven = fileExists('pom.xml')
                             incrementals = fileExists('.mvn/extensions.xml') &&
                                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
+                            final String gitUnavailableMessage = "[buildPlugin] Git CLI may not be available"
                             if (incrementals) { // Incrementals needs 'git status -s' to be empty at start of job
                                 if (isUnix()) {
-                                    sh(script: 'git clean -xffd > /dev/null 2>&1',
-                                       label: 'Clean for incrementals',
-                                       returnStatus: true) // Ignore failure if CLI git is not available
+                                    sh 'git clean -xffd || echo "${gitUnavailableMessage}"'
```
jglick, line 72: Will not work: `'` will not interpolate Groovy expressions. And you should not use `"`. Just inline.

vars/buildPlugin.groovy
```diff
@@ -66,27 +66,20 @@ def call(Map params = [:]) {
...
+                                    bat 'git clean -xffd || echo "${gitUnavailableMessage}"'
                                 }
                             }
 
                             if (gitDefaultBranch) {
                                 if (isUnix()) {
-                                    sh(script: "git config --global init.defaultBranch '${gitDefaultBranch}'",
-                                       label: 'Set default branch for git',
-                                       returnStatus: true) // Ignore failure if CLI git is not available
+                                    sh "git config --global init.defaultBranch '${gitDefaultBranch}' || echo '${gitUnavailableMessage}'"
```
jglick, line 80: Also best to use `"` here (and later for `bat`). To pass in `gitDefaultBranch` use

```groovy
withEnv(["GIT_DEFAULT_BRANCH=$gitDefaultBranch"]) {
  sh 'whatever $GIT_DEFAULT_BRANCH'
  bat 'whatever %GIT_DEFAULT_BRANCH%'
}
```

vars/buildPlugin.groovy
```diff
@@ -67,19 +67,23 @@ def call(Map params = [:]) {
...
-                                if (isUnix()) {
-                                    sh "git config --global init.defaultBranch '${gitDefaultBranch}' || echo '${gitUnavailableMessage}'"
-                                } else {
-                                    bat "git config --global init.defaultBranch '${gitDefaultBranch}' || echo '${gitUnavailableMessage}'"
+                                if (gitDefaultBranch) {
+                                    withEnv(["GITDEFAULTBRANCH=${gitDefaultBranch}"]) {
+                                        if (isUnix()) {
+                                            sh 'git config --global init.defaultBranch "$GITDEFAULTBRANCH" || echo "$GITDEFAULTBRANCH"'
+                                        } else {
+                                            bat 'git config --global init.defaultBranch %GITDEFAULTBRANCH% || echo %GITDEFAULTBRANCH%'
```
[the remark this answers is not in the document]
dduportal: Damn, I should restrain from comitting stuff end of day for me.
Many thanks for reviewing carefully @jglick

dduportal:
I've tested the `xx || yy` with the docker image for powershell (https://hub.docker.com/_/microsoft-powershell) which can be run on a Docker Linux Engine, even though it's powershell (and not the windows CMD).

My reference is https://ss64.com/nt/syntax-conditional.html but I cannot be completely sure that it does not break something.

Currently trying on https://github.com/jenkinsci/jenkins-infra-test-plugin

dduportal:
Looks good:

- Windows: https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fjenkins-infra-test-plugin/detail/jenkins-infra%2Fpipeline-library%2F344/1/pipeline/21#step-31-log-2

Capture d’écran 2022-04-07 à 16 28 08


- Linux: https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fjenkins-infra-test-plugin/detail/jenkins-infra%2Fpipeline-library%2F344/1/pipeline/14#step-50-log-1

Capture d’écran 2022-04-07 à 16 28 19

PR #356, opened 2022-04-20
Format repository

basil:
I've been finding it very difficult to edit this repository because of the wildly inconsistent formatting in each source file, which drives me nuts. Therefore in this PR I am adding Spotless formatting for the `pom.xml` file and all Groovy source code. Core developers should be used to Spotless by now: it automatically verifies that the code is formatted when running `mvn clean verify`, and if there is an error it fails the verification and prints the error along with the command you need to run to automatically correct the error. This is friendly to developers and ensures the code stays neatly formatted in each future PR going forward.

It is essential that this PR be squash merged. After it is squash merged I will add another commit to remove it from the blame view as in [this post](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view=).

The settings are derived from [this style guide](https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml). Reviewers are encouraged to hide whitespace changes while reviewing this.

CC @timja


timja (approved):

dduportal:
Hello @basil @timja , do you have a documentation or tutorial to point to guide on this "formatting stuff"? I personnaly have no idea how is that working and I would like to avoid you this kind of pain in the future: is there a maven step that could format the code ala `go fmt` ? Because it seems that it highly depends on the IDE and its setup?

timja:
Generally this: https://github.com/diffplug/spotless/tree/master/plugin-maven

basil:
As I wrote in the review description:

> Core developers should be used to Spotless by now: it automatically verifies that the code is formatted when running `mvn clean verify`, and if there is an error it fails the verification and prints the error along with the command you need to run to automatically correct the error.

basil:
> Generally this: https://github.com/diffplug/spotless/tree/master/plugin-maven

Yes but s/master/main to get the current copy of the docs rather than a one-year-old version (a mistake I've made many times, not sure why didn't set up a redirect):

https://github.com/diffplug/spotless/tree/main/plugin-maven

timja:
Filed https://github.com/diffplug/spotless/issues/1174

dduportal:
> As I wrote in the review description:
> 
> > Core developers should be used to Spotless by now: it automatically verifies that the code is formatted when running `mvn clean verify`, and if there is an error it fails the verification and prints the error along with the command you need to run to automatically correct the error.

Thanks for the work and the links.

I think there is a gap in the target audience (for now) but I'm willing to help filing it: the contributors of this repo are only half "Core developers". The other half are one-time contributor or infrastructure people (which are not "maven" users generally speaking). I'll try to add a PR in the contributing guide explaining it: I'll need your help as reviewer to validate that I understood it correctly

One last question: is there a specific reason to bind the spotless goals to the `verify` step of maven?
I understand that it is the usual way (as per https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#usual-command-line-calls for instance) but it feels weird to me (maybe because non english native) because I would expect such task (checking formatting) to happen really soon. The `validate` goal would make more sense for me: I guess it might be related to the semantic of `validate` / `verify` (where I expect maven to fail fast if the formatting is incorrect, or better to format the code as soon as possible).


This PR is a HUGE improvement as it makes the style rules written and auditable. Many thanks!

Jenkinsfile
```diff
@@ -1,21 +1,21 @@
...
+  agent {
+    label 'maven-11'
+  }
+  options {
+    timestamps()
+  }
+  stages {
+    stage('Test') {
+      steps {
+        sh 'mvn --no-transfer-progress -B clean verify'
```
jglick, line 12: Not a formatting change, FTR (just browsing the whitespace-ignoring diff).

jglick (commented):
Thanks!

dduportal:
> I suppose this could be done during the `validate` phase rather than the `verify` phase; I have no strong preference. There are some benefits to doing it during the `verify` phase, though: while iterating on tests, you don't get bothered with formatting problems as you re-run tests several times. The main concern is not whether formatting checks are done during the `verify` or `validate` phase but rather that they only check the code and not automatically fix violations (which would violate the principle of least surprise). Spotless meets this criteria by having a separate `spotless:check` goal that can be used during the build, whereas the `spotless:apply` goal can be used interactively, giving the developer the chance to inspect and commit the results.

Fair enough, sounds logical (I'm not a fan of this because it leads to surprise when the full job is running and failing in CI while the develoepr's mind is already on another topic, but it is a personal preference and your arguments also make sense to me). Thanks for the explanation.

> I didn't write any developer documentation for this when I introduced the change in core because I thought Spotless was self-explanatory enough: when it fails, it shows you exactly what is formatted incorrectly and prints a single command that you can run to fix the formatting error automatically. So I didn't think there was more I could explain beyond that. If I am wrong about that and there is more to explain, then feel free to let me know what needs more documentation.

I'm advocating, for the specific case of this repository (which contain pipeline library code used by the Jenkins infrastructure), for helping a normal user to know what they should do to format their code, and how to react to a build failure caused by this new behavior.
As you said, it is self explanatory for core and their developers.
But jenkins infra team is not part of this (I'm merely a former maven user i don't know if others have used Maven in their previous life).
But worry not buddy, I'll propose something to fill this gap :)

dduportal:
Here you are: https://github.com/jenkins-infra/pipeline-library/pull/371

PR #359, opened 2022-04-20
Publish checks when invoking deprecated functionality

basil:
Publishing a check when deprecated functionality is invoked is a useful way to draw attention to the usage and ensure that it gets updated, so let us abstract this into a common method and use it in more places.


dduportal:
That is wonderful! Many thanks!

PR #362, opened 2022-04-20
Remove unused argument in `retrieveMavenSettingsFile`

basil:
This argument was unused and also made no sense because it was declared of type string but its default value was of type integer. Technically an API change, but I verified all existing callers only use the single-argument version:

- https://github.com/search?ref=simplesearch&type=Code&q=user%3Ajenkinsci+retrieveMavenSettingsFile
- https://github.com/search?ref=simplesearch&type=Code&q=user%3Ajenkins-infra+retrieveMavenSettingsFile

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -52,7 +52,7 @@ Object checkoutSCM(String repo = null) {
  * @param jdk Version of JDK to be used (no longer used)
  * @return {@code true} if the file has been defined
  */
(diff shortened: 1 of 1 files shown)


timja (approved):

dduportal (approved):
Good catch! Are you using a code analyzer to catch such issues, or is it while reading/browsing/studying the code?

lemeurherve (approved):

jglick:
Unused as of #216 I think. See also #236.

PR #366, opened 2022-04-21
fix(buildDockerAndPublishImage) avoid sh interpolation calls

dduportal:
This PR works out the `sh` steps in `buildDockerAndPublishImage` which are using interpolated strings.

It removes them in favor of using environment variables.

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -137,16 +137,17 @@ def call(String imageName, Map userConfig=[:]) {
           withCredentials([
             usernamePassword(credentialsId: "${finalConfig.gitCredentials}", passwordVariable: 'GIT_PASSWORD', usernameVariable: 'GIT_USERNAME')
           ]) {
-            echo "Configuring git user and email"
-            sh 'git config user.name "${GIT_USERNAME}"'
-            sh 'git config user.email "jenkins-infra@googlegroups.com"'
+            withEnv(["NEXT_VERSION=${nextVersion}"]) {
+              echo "Tagging and pushing the new version: $nextVersion"
+              sh '''
+              git config user.name "${GIT_USERNAME}"
+              git config user.email "jenkins-infra@googlegroups.com"
 
-            echo "Tagging New Version: $nextVersion"
-            sh "git tag -a $nextVersion -m ${imageName}"
-
-            echo 'Pushing Tag'
-            sh 'git push origin --tags'
-          }
+              git tag -a "${NEXT_VERSION}" -m "${IMAGE_NAME}"
+              git push origin --tags
+              '''
+            } // withEnv
+          } // withCredentials
         } // stage
       } // if
 
@@ -163,7 +164,10 @@ def call(String imageName, Map userConfig=[:]) {
               imageDeployName += ":${env.TAG_NAME}"
             }
           }
-          sh "IMAGE_DEPLOY_NAME=${imageDeployName} make deploy"
+          withEnv(["IMAGE_DEPLOY_NAME=${imageDeployName}"]) {
+            // Please note that "make deploy" uses the environment variable "IMAGE_DEPLOY_NAME"
+            sh 'make deploy'
+          } // withEnv
         } //stage
       } // if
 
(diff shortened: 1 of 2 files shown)


dduportal:
Draft status for now, time to get initial code review + unit test on ci.jenkins.io.

Gotta check the library change on a custom end to end test before setting it "ready to merge".

vars/buildDockerAndPublishImage.groovy
```diff
@@ -178,24 +182,29 @@ def call(String imageName, Map userConfig=[:]) {
...
+              "GH_URL=${ghUrl}",
+              "GH_RELEASES_API_URI=${ghReleasesApiUri}",
+            ]) {
               sh '''
               if ! command -v gh 2>/dev/null >/dev/null
               then
                 echo "INFO: No gh binary found: Installing it from ${ghUrl}."
                 temp_dir="$(mktemp -d)"
-                curl --silent --show-error --location --output "${temp_dir}/gh.tgz" "${gh_url}"
+                curl --silent --show-error --location --output "${temp_dir}/gh.tgz" "${GH_URL}"
```
daniel-beck, line 195: Not the point of the PR, but ❤️ long opts :) Very easily readable

dduportal: that is a motto. I'm not doing it enough, but long flags is so much help for people reading scripts \o/

dduportal:
Test in progress:
- [ ] PR (to test half of this library): https://github.com/jenkins-infra/docker-confluence-data/pull/28
- [ ] principal branch build (that should create a new tag + push latest)
- [ ] tag build (that should push the tagged image)

dduportal:
THis PR is ready for review + merge. Tests are fine for me.

timja (approved):

vars/buildDockerAndPublishImage.groovy
```diff
@@ -178,24 +182,29 @@ def call(String imageName, Map userConfig=[:]) {
...
                 gh --version
               fi
               '''
-            }
 
-            final String release = sh(returnStdout: true, script: "gh api /repos/${org}/${repository}/releases | jq -e -r '.[] | select(.draft == true and .name == \"next\") | .id'").trim()
-            sh "gh api -X PATCH -F draft=false -F name=${env.TAG_NAME} -F tag_name=${env.TAG_NAME} /repos/${org}/${repository}/releases/$release"
+              final String release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'.[] | select(.draft == true and .name == "next") | .id\'').trim()
+              withEnv(["GH_NEXT_RELEASE_URI=${ghReleasesApiUri}/${release}"]) {
+                sh 'gh api -X PATCH -F draft=false -F name="${TAG_NAME}" -F tag_name="${TAG_NAME}" "${GH_NEXT_RELEASE_URI}"'
```
jglick, line 205: I think you are doing too much work here. Whenever you see a `sh` step with `returnStdout`, be on the alert. If the result is being fed into a general Pipeline command (`archiveArtifacts`, etc.), then OK. But if it is being piped back into `withEnv` or `sh` then you are doing too much in Pipeline. The `release` here could have been a shell variable, or just a `$(…)` inline evaluation (``` `…` ``` if using true Bourne shell rather than Bash). This whole block could have been one `sh` step with a literal argument. (Also consider using external `*.sh` files in `resources/` for maintainability. See `libraryResource` step.)

jglick: In fact most of the 250+ lines of Pipeline code in this file could be eliminated and converted to plain old Bash (Make, Ruby, whatever) scripting.

dduportal: I agree with you @jglick . Thanks for the insights. I'm not the initial author of this code, and this more a matter of taking the time to improve such behavior (instead of another infra task) to be quite transparent.

One of the reason for doing code in the pipeline here was initially to ensure that no `makefile` injection could be done (as Jenkins is the actor in charge of checking if a build author has the permission to execute stuff or not).
I guess that, with the pipeline library being responsible to load the correct `Makefile`, we are able to do what you point at, without security issue, getting the best of both worlds.

jglick: > ensure that no `makefile` injection could be done

If you think that by writing stuff in Pipeline script you are preventing that, I would be very nervous. There are numerous Shellshock-like ways to trick a library author into running unwanted code in a privileged context. The Jenkins security model generally treats a job as the granularity of access control. To have individual stages (or whatever) run at a different privilege requires great care. Really I would only trust https://perldoc.perl.org/perlsec#Taint-mode for that, or using containers for isolation (presuming you are using a fully patched Linux kernel).

dduportal: oh no worries on the granularity there: we treat the whole build (and the whole pipeline) as the minimum atom.
Either you have the permission to run it, or not. (eventually, you are allowed to execute a build on your PR but it will pick the Jenkinsfile from the principal branch). No more.

jglick: > you are allowed to execute a build on your PR but it will pick the Jenkinsfile from the principal branch

Remember that this still allows other files from the repo to be modified under your control by default. You can use the `readTrusted` step to tune that.

dduportal: @jglick yep, that is what are trying to do: https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildDockerAndPublishImage.groovy#L24 (`libraryResource`) to load the common `Makefile`.

Terraform pipeline library uses another strategy (we are comapring them): it provides the `Makefile` for developers through a git submodule, but it clones the sub-repo (overriding the submodule) so it has controls over the "what file is used": https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L209-L216

jglick: (edited to use permalinks via **Y** for inline display)

PR #374, opened 2022-05-05
feat: switch mirror to HTTPS on get.jenkins.io (Jenkins WAR retrieval)

dduportal:
This change request is motivated by https://github.com/jenkins-infra/helpdesk/issues/2888.

It switches the WAR retrieval from HTTP-only on the legacy (and on the verge of depreciation) mirror system `mirror.jenkins.io`.

The new URL uses the actual mirroring service `get.jenkins.io` which has the following benefits:

- Uses HTTPS
  - former service `mirror.jenkins.io` is HTTP only
- Highly available (scalable horizontally, on the Kubernetes cluster `prodpublick8s` on Azure)
  - former service `mirror.jenkins.io` is hosted on a standalone VM, which is also used for update center and packaging
- Better performances (dsitributed requests + a real life table of mirrors)
  - former service `mirror.jenkins.io` have an irregular response time, often peaks above 10s as per https://github.com/jenkins-infra/helpdesk/issues/2888
  - former service `mirror.jenkins.io` always redirect the WAR download to `archives.jenkins.io` (which acts as a fallback), while get.jenkins.io choose a mirror near the HTTP client (and only fall backs to `archives.jenkins.io` for older WAR version not mirrored)

This change request also cleans up the README and the `runAth()` from an unused variable mentionning `mirror.jenkins.io`.

The following project could be impacted by this change:

- Any call to the method `stashJenkinsWar()` of the global library `infra`
- Transitively:
  - Any project using the `runAth()` method because it uses the method `stashJenkinsWar()`
    - jenkinsci/jenkins main build: https://github.com/jenkinsci/jenkins/blob/master/Jenkinsfile#L158
  - Any project using the `runPct()` method because it uses the method `stashJenkinsWar()`
  - Any project using the method `essentialsTest()` because it uses both `runAth()` and `runPct()`

About the impacts:
- The 4 URLs  (for WAR retrieval) that are changed have been tested successfully: they all redirects to  working WAR file with the same checksum as the old URLs when writing these lines

ETA: this PR will be merged the 09th of May 2022 if approved, unless there is a strong reason to cancel it.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -145,7 +145,7 @@ infra.ensureInNode('docker,java') {
 
 === infra.stashJenkinsWar(jenkins, stashName)
 
-Given a version of jenkins downloads it if neccesary and stashes it under the given name (which defaults to "jenkinsWar",
+Given a version of jenkins downloads it if necessary and stashes it under the given name (which defaults to "jenkinsWar",
 see the step doc for more documentation about he allowed versions
 
 .Jenkinsfile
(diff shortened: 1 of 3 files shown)


MarkEWaite (approved):
Thanks for doing this!

timja (approved):

Issue #379, opened 2022-05-14
`buildDockerAndPublishImage` fails at `GitHub Release` stage if there are more than one "next" draft releases


lemeurherve:
https://github.com/jenkins-infra/pipeline-library/blob/8b9941bd72130ea732cc804726218d64da7da22b/vars/buildDockerAndPublishImage.groovy#L199

Example of [failing build](https://infra.ci.jenkins.io/blue/organizations/jenkins/docker-jobs%2Fdocker-openvpn/detail/1.0.75/1/pipeline/113) :
<img width="1125" alt="image" src="https://user-images.githubusercontent.com/91831478/168426157-944416d1-062b-4e7e-9308-fd6ef412308f.png">

<details>
<summary>Corresponding releases</summary>

![image](https://user-images.githubusercontent.com/91831478/168426201-64185b6b-a4d6-4ce8-94b5-17dff10235d4.png)

</details>

https://github.com/jenkins-infra/docker-openvpn/releases

dduportal:
Side note : there is also the case of no « next » release. Usually, adding trigger event « release » on the gh workflow for release drafter fix it. But maybe we could trigger release-drafter here and manage the draft release for ourselves ?

PR #380, opened 2022-05-14
fix: retrive only the last "next" draft release id

lemeurherve:
Fixes https://github.com/jenkins-infra/pipeline-library/issues/379

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -196,7 +196,7 @@ def call(String imageName, Map userConfig=[:]) {
               fi
               '''
 
-              final String release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'.[] | select(.draft == true and .name == "next") | .id\'').trim()
+              final String release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'.[] | select(.draft == true and .name == "next") | .id | sort_by(id) | reverse | first\'').trim()
               withEnv(["GH_NEXT_RELEASE_URI=${ghReleasesApiUri}/${release}"]) {
                 sh 'gh api -X PATCH -F draft=false -F name="${TAG_NAME}" -F tag_name="${TAG_NAME}" "${GH_NEXT_RELEASE_URI}"'
               } // withEnv
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -196,7 +196,7 @@ def call(String imageName, Map userConfig=[:]) {
               fi
               '''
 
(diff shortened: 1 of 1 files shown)


dduportal (dismissed):

vars/buildDockerAndPublishImage.groovy
```diff
@@ -196,7 +196,7 @@ def call(String imageName, Map userConfig=[:]) {
               fi
               '''
 
-              final String release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'.[] | select(.draft == true and .name == "next") | .id\'').trim()
+              final String release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'.[] | select(.draft == true and .name == "next") | .id | sort_by(id) | reverse | first\'').trim()
```
lemeurherve, line 199:
```suggestion
              final String release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r '. | max_by(.id) | select(.draft == true and .name == "next").id\'').trim()
```

vars/buildDockerAndPublishImage.groovy
```diff
@@ -196,7 +196,7 @@ def call(String imageName, Map userConfig=[:]) {
               fi
               '''
 
-              final String release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'.[] | select(.draft == true and .name == "next") | .id\'').trim()
+              final String release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r '. | max_by(.id) | select(.draft == true and .name == "next").id\'').trim()
```
lemeurherve, line 199:
```suggestion
              final String release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'. | max_by(.id) | select(.draft == true and .name == "next").id\'').trim()
```

dduportal (approved):

jglick:
I wonder if this also affects `jenkins-maven-cd-action`.

dduportal:
> I wonder if this also affects `jenkins-maven-cd-action`.

Interesting: is your comment about the fact that `jenkins-maven-cd-action` **could** use this library, or is it about the root problem (handling presence of multiple `next` releases)?

jglick:
I am guessing that the code here was copied from https://github.com/jenkins-infra/jenkins-maven-cd-action/blob/a0a2c865bf13c0937b93d3190ed128af1bab2947/run.sh#L11 so I was wondering whether an analogous problem affected CD workflows. Do not recall offhand. CC @timja

dduportal:
As far as I remember, this code was written by @garethjevans , so yep, smells like same source 🤗

PR #388, opened 2022-05-23
fix: correct IMAGE_DIR for Windows

lemeurherve:
Fix "process_begin: CreateProcess(NULL, @echo %cd%, ...) failed."


dduportal (approved):
👍

PR #398, opened 2022-05-31
chore(buildDockerAndPublishImage): use `docker` by default instead of `img`

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/2952


lemeurherve:
I've fixed tests but didn't added new ones testing specifically `img` since it's deprecated and will be removed later.

dduportal (approved):
LGTM

PR #399, opened 2022-05-31
feat(buildDockerAndPublishImage): add multipleSemverImagesBuild param

lemeurherve:
Close https://github.com/jenkins-infra/pipeline-library/issues/396
Follow-up of https://github.com/jenkins-infra/pipeline-library/pull/395

PR #400, opened 2022-06-01
fix(buildDockerAndPublishImage): semver tagging of multiple images in a parallel build

lemeurherve:
Follow-up of https://github.com/jenkins-infra/pipeline-library/pull/399

Test in https://github.com/jenkins-infra/docker-inbound-agents/pull/24


dduportal:
> Rename 'multipleSemverImagesBuild' by 'includeStageNameInTag' as we may want to include the stage name in tag in other context than a multiple semver images build
> 
> Follow-up of #399

Not sure to understand (at all) what do you refer as "stage name". Is it "build", "test", "get next version", "deploy" or something like that?

lemeurherve:
> Build fails because of the checkspot not applied:

Currently adding a pre-commit git hook to avoid forgetting it.

dduportal (dismissed):

dduportal (approved):

PR #412, opened 2022-06-09
fix(buildDockerAndPublishImage): use the complete tag as previous-version arg

lemeurherve:
Failed test in this replay: https://infra.ci.jenkins.io/blue/organizations/jenkins/docker-jobs%2Fdocker-inbound-agents/detail/main/21/pipeline, `jx-release-version` see the commit associated with the tag hasn't changed, doesn't take in account the commit message(s) and bump only a patch:

image

Another test in this replay: https://infra.ci.jenkins.io/blue/organizations/jenkins/docker-jobs%2Fdocker-inbound-agents/detail/PR-26/2/pipeline ✅ 

image

Follow-up of https://github.com/jenkins-infra/pipeline-library/pull/408

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -75,11 +75,11 @@ def call(String imageName, Map userConfig=[:]) {
                 echo "Including the image name '${imageName}' in the next version"
                 // Retrieving the semver part from the last tag including the image name
                 String currentTagScript = 'git tag --list \"*' + imageInTag + '\" --sort=-v:refname | head -1'
-                String currentSemVerVersionPart = sh(script: currentTagScript, returnStdout: true).trim().replace(imageInTag, '')
-                echo "Current semver version part is '${currentSemVerVersionPart}'"
+                String currentSemVerVersion = sh(script: currentTagScript, returnStdout: true).trim()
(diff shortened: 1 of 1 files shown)


dduportal (commented):
WDYT about using unit test to map different cases?

lemeurherve:
> WDYT about using unit test to map different cases?

That would be great, but I admit I just want to ship a functional version right now 😛

lemeurherve:
> > WDYT about using unit test to map different cases?
> 
> That would be great, but I admit I just want to ship a functional version right now 😛

Also, I presume simulating these different git cases won't be that easy.

dduportal:
> > > WDYT about using unit test to map different cases?
> > 
> > 
> > That would be great, but I admit I just want to ship a functional version right now 😛
> 
> Also, I presume simulating these different git cases won't be that easy.

That would allow to:
- List the identified cases (not a problem if not exhaustive at the beginning, that we'll grow when new ones are met)
- Validate the logic for these cases, excluding the behavior of `jx-release-version`

lemeurherve:
> approving to unblock you, but I'm not sure to follow what's going on

I was passing only the semver part of the existing tag to `jx-release-version` as `previous-version` arg for every image instead of the complete existing tag (including the image name)

PR #413, opened 2022-06-10
fix: don't fail if there is no draft release

lemeurherve:


dduportal (approved):

test/groovy/BuildDockerAndPublishImageStepTests.groovy
```diff
@@ -355,6 +355,32 @@ class BuildDockerAndPublishImageStepTests extends BaseTest {
     verifyMocks()
   }
 
+  @Test
+  void itBuildsAndDeploysButNoReleaseWhenTriggeredByTagAndSemVerEnabledButNoReleaseDrafter() throws Exception {
+    def script = loadScript(scriptName)
```
lemeurherve, line 360: Need to make [shellMocks respond with an error for the cases](https://github.com/jenkins-infra/pipeline-library/blob/20661392972a865809e5487f2db5fcfaabf1c4fa/test/groovy/BuildDockerAndPublishImageStepTests.groovy#L52-L53) `case {command.contains('gh api ${GH_RELEASES_API_URI}')}:` and `case {command.contains('gh api $env:GH_RELEASES_API_URI')}:` but only for this test. Any idea?

dduportal: You might be interested in checking the last example of https://github.com/jenkinsci/JenkinsPipelineUnit#mocking-shell-steps. let me prepare a suggestion to give hints

dduportal:
```suggestion
    def script = loadScript(scriptName)
    
    helper.addShMock('gh api ${GH_RELEASES_API_URI} | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\'', '[Warn] No release found in GitHub.', 1)
    
    
```

dduportal: Alternative:

```suggestion
    def script = loadScript(scriptName)
    
    helper.registerAllowedMethod('sh', [Map.class], { m -> throw new Exception('[Warn] No release found in GitHub') })
```

dduportal: The whole idea is to override, only for this test, the sh behavior. I assume that you might want to integrate this into the "generic" mock, so you might want to improve the mock structure to handle custom cases such this one: maybe a parameter and you rewrite it?


Don't forget : KISS is not always a good thing.

lemeurherve: I'm not sure helper.addShMock would have any effect, as the previous ones were overriden with the registerAllowedMethod.

Adding the alternative to the test I also fail to see how other shell scripts would be mocked.
I can copy/paste the complete shellMocks method in this test to return an exception but it doesn't seems clean at all.

vars/buildDockerAndPublishImage.groovy
```diff
@@ -227,24 +227,34 @@ def call(String imageName, Map userConfig=[:]) {
...
-                  sh 'gh api -X PATCH -F draft=false -F name="${TAG_NAME}" -F tag_name="${TAG_NAME}" "${GH_NEXT_RELEASE_URI}"'
+                  release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\'').trim()
                 } else {
-                  powershell 'gh api -X PATCH -F draft=false -F name="$env:TAG_NAME" -F tag_name="$env:TAG_NAME" "$env:GH_NEXT_RELEASE_URI"'
+                  release = powershell(returnStdout: true, script: 'gh api $env:GH_RELEASES_API_URI | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\'').trim()
                 }
-              } // withEnv
+                withEnv(["GH_NEXT_RELEASE_URI=${ghReleasesApiUri}/${release}"]) {
+                  if (isUnix()) {
+                    sh 'gh api -X PATCH -F draft=false -F name="${TAG_NAME}" -F tag_name="${TAG_NAME}" "${GH_NEXT_RELEASE_URI}"'
```
jglick, line 249: Better to just use e.g.

```groovy
sh 'gh api … || echo "Warning: …"'
```

lemeurherve: Indeed, thanks!

dduportal (approved):

jglick:
FYI the idiom in https://github.com/jenkins-infra/interesting-category-action/pull/5 may be helpful.

PR #417, opened 2022-06-13
fix: integer instead of string for release id

lemeurherve:
Follow-up of https://github.com/jenkins-infra/pipeline-library/pull/413

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -236,14 +236,14 @@ def call(String imageName, Map userConfig=[:]) {
             final String repository = origin.split('/')[4]
             final String ghReleasesApiUri = "/repos/${org}/${repository}/releases"
             withEnv(["GH_RELEASES_API_URI=${ghReleasesApiUri}"]) {
-              String release = ''
+              int releaseId = 0
               if (isUnix()) {
-                release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\' || echo \'\'').trim()
+                releaseId = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\' || echo 0').trim()
               } else {
-                release = powershell(returnStdout: true, script: 'gh api $env:GH_RELEASES_API_URI | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\' || echo \'\'').trim()
+                releaseId = powershell(returnStdout: true, script: 'gh api $env:GH_RELEASES_API_URI | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\' || echo 0').trim()
               }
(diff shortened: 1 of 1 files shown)


dduportal (commented):
What problem does it solve?

lemeurherve:
> What problem does it solve?

The pipeline executed the gh api release call even if release was null:
image

I've got another other problem to fix too, the `||` not interpreted by `powershell`
image

It has been introduced [in Powershell 7 though](https://stackoverflow.com/a/59039112/4074148). Will try with a call with `pwsh` instead of `powershell` to confirm.
(Related: https://github.com/jenkins-infra/helpdesk/issues/2974)

vars/buildDockerAndPublishImage.groovy
```diff
@@ -236,14 +236,14 @@ def call(String imageName, Map userConfig=[:]) {
...
             final String ghReleasesApiUri = "/repos/${org}/${repository}/releases"
             withEnv(["GH_RELEASES_API_URI=${ghReleasesApiUri}"]) {
-              String release = ''
+              int releaseId = 0
               if (isUnix()) {
-                release = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\' || echo \'\'').trim()
+                releaseId = sh(returnStdout: true, script: 'gh api ${GH_RELEASES_API_URI} | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\' || echo 0').trim()
               } else {
-                release = powershell(returnStdout: true, script: 'gh api $env:GH_RELEASES_API_URI | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\' || echo \'\'').trim()
+                releaseId = powershell(returnStdout: true, script: 'gh api $env:GH_RELEASES_API_URI | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\' || echo 0').trim()
```
lemeurherve, line 243:
```suggestion
                releaseId = pwsh(returnStdout: true, script: 'gh api $env:GH_RELEASES_API_URI | jq -e -r \'[ .[] | select(.draft == true and .name == "next").id] | max\' || echo 0').trim()
```
To check if `||` is valid with `pwsh` (it isn't for `powershell`, cf https://github.com/jenkins-infra/pipeline-library/pull/417#issuecomment-1153805816)

dduportal:
About the `||` on windows, if that might help:
- https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildPlugin.groovy#L73
- https://github.com/jenkins-infra/pipeline-library/pull/344#issue-1188027952

jglick (commented):
This whole block should be made into a single `sh` or `powershell` step.

lemeurherve:
> This whole block should be made into a single `sh` or `powershell` step.

I concur, that what I will do, idem for the "Get Next Version" step.

I'd like to finish this existing function update before though, as it seems I'm really close to a successful docker-inbound-agents multi-OS build now.

jglick:
> a separate script

Well, ideally, but as a first step I would suggest just coalescing to a single step with a multiline argument.

dduportal (approved):

PR #433, opened 2022-07-08
Discover the reference build before recording code coverage

uhafner:
In [release 3.0.0](https://github.com/jenkinsci/code-coverage-api-plugin/releases/tag/v3.0.0) the code coverage plugin  introduces a new feature to compute the delta coverage of pull requests. In order to compute this coverage the reference build must be computed before the coverage recording step. Otherwise we cannot compute the delta coverage of a pull request.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -127,6 +127,7 @@ def call(Map params = [:]) {
                     if (!skipTests) {
                       junit('**/target/surefire-reports/**/*.xml,**/target/failsafe-reports/**/*.xml,**/target/invoker-reports/**/*.xml')
                       if (first) {
+                        discoverReferenceBuild()
                         publishCoverage calculateDiffForChangeRequests: true, adapters: [jacocoAdapter('**/target/site/jacoco/jacoco.xml')]
                       }
                     }
@@ -162,11 +163,9 @@ def call(Map params = [:]) {
                 }
 
                 if (first) {
-                  folders = env.JOB_NAME.split('/')
-                  if (folders.length > 1) {
-                    discoverGitReferenceBuild(scm: folders[1])
+                  if (skipTests) { // otherwise the reference build has been computed already
+                    discoverReferenceBuild()
                   }
-
                   echo "Recording static analysis results on '${stageIdentifier}'"
 
                   recordIssues(
@@ -279,6 +278,13 @@ def call(Map params = [:]) {
   parallel(tasks)
   if (publishingIncrementals) {
     infra.maybePublishIncrementals()
+  }
+}
+
+private void discoverReferenceBuild() {
+  folders = env.JOB_NAME.split('/')
+  if (folders.length > 1) {
+    discoverGitReferenceBuild(scm: folders[1])
   }
 }
 


basil (approved):
I suppose [the `Jenkinsfile` in core](https://github.com/jenkinsci/jenkins/blob/7aa91f8158624dba15b890431440d8b0f5860e32/Jenkinsfile) needs to be adapted in the same way?

timja (approved):

vars/buildPlugin.groovy
```diff
@@ -282,6 +281,13 @@ def call(Map params = [:]) {
   }
 }
 
+private void discoverReferenceBuild() {
+  folders = env.JOB_NAME.split('/')
```
jglick, line 285: Note that just `JOB_NAME` suffices

vars/buildPlugin.groovy
```diff
@@ -282,6 +281,13 @@ def call(Map params = [:]) {
...
+  if (folders.length > 1) {
+    discoverGitReferenceBuild(scm: folders[1])
```
jglick, line 287: What is this doing? If `JOB_NAME` were e.g. `some/path/to/some/folder/and/then/your/job-name/PR-123`, you are computing `path`??

Perhaps you meant `folders[-2]` (`job-name` in this example)? Or `../master`? Or 

[Docs for the step](https://www.jenkins.io/doc/pipeline/steps/git-forensics/#discovergitreferencebuild-discover-git-reference-build) list the `scm` parameter but it has no documentation.

jglick (commented):
Commenting on weird stuff. Yes I can see you are just moving code around, it is not new, but this is the first time I have seen it.

vars/buildPlugin.groovy
uhafner: In the `discoverGitReferenceBuild` step the reference build is discovered from the associated SCM. In order to find the correct SCM (there are multiple ones available and Jenkins sadly does not define a *main* SCM) the ID `git https://github.com/jenkinsci/analysis-model.git'` is matched against the `SCM` parameter (it is documented in the JavaDoc and UI). Seems that we need a third place were we would copy and paste that documentation right now (`help-scm.html`). 

I am using the job name to identify the correct SCM. So for plugins we do not have a generic `some/path/to/some/folder/and/then/your/job-name/PR-123`, we always have the fixed name `Plugins/job-name/PR-738`. So I am using folders[1] to extract the `job-name`. But `[-2]` as index should work as well and is more stable when we might refactor the jobs structure in the future.

vars/buildPlugin.groovy
uhafner: Ok, I copied it from the `env.NODE_LABELS` part a couple of lines down.

vars/buildPlugin.groovy
jglick: > there are multiple ones available and Jenkins sadly does not define a _main_ SCM

For branch project it does. https://javadoc.jenkins.io/plugin/scm-api/jenkins/scm/api/SCMRevisionAction.html

> more stable when we might refactor the jobs structure

Exactly.

PR #436, opened 2022-07-13
Use `kubernetesAgent(handleNonKubernetes: true)`

jglick:
https://github.com/jenkins-infra/pipeline-library/pull/405#discussion_r916032309
https://github.com/jenkinsci/kubernetes-plugin/pull/1190/commits/c38fe0fabf27d0583c9b546aa701dc075245e72d
https://github.com/jenkins-infra/helpdesk/issues/2984#issuecomment-1183305821

### DIFF
--- a/test/groovy/BaseTest.groovy
+++ b/test/groovy/BaseTest.groovy
@@ -43,6 +43,8 @@ class BaseTest extends DeclarativePipelineTest {
     helper.registerAllowedMethod('node', [String.class, Closure.class], { s, body -> body() })
     helper.registerAllowedMethod('retry', [Map.class, Closure.class], { m, body ->body() })
     helper.registerAllowedMethod('agent', [], { 'agent' })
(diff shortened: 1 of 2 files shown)


jglick:
CC @vlatombe

dduportal:
If I udnerstand correctly, with this change:

- That will manage node retry for kube in a "native" way (ref. pod killing reasons different than VMs)
- If the infra-team changes the labels to use VM agents instead of kube agent , that should keep handling node retries: am I correct?

jglick:
> That will manage node retry for kube in a "native" way

Correct.

> If the infra-team changes the labels to use VM agents instead of kube agent , that should keep handling node retries

Yes, or for that matter various plugins already specifically request the `docker` label, or decline to specify `useContainer: true`. So for those builds (or branches of builds), this PR causes no change in behavior.

timja (approved):

PR #453, opened 2022-08-12
Add placeholders for LATEST and LATEST_LTS jenkins versions

strangelookingnerd:
IMHO it should be common practice to test plugins against the minimal Jenkins version by setting the corresponding buildPlugin() configuration to `null`. Further one should be testing against the latest stable LTS release and the cutting edge latest releases as well. It is very cumbersome to always update the corresponding configurations manually in my Jenkinsfile and trying to update it via Dependabot or similar does not seem very feasible.

This is why I propose to introduce placeholders `LATEST` and `LATEST_LTS` that will automatically be replaced with the information from the Jenkins Update Sites (https://github.com/jenkins-infra/update-center2/blob/master/site/LAYOUT.md) by the pipeline-library.
This way it could be easily encouraged to always include these versions in a plugin build in order to identify compatibility issues early.

If the general idea is accepted I'd of course add the corresponding documentation for it in this PR!

### DIFF
--- a/test/groovy/BuildPluginStepTests.groovy
+++ b/test/groovy/BuildPluginStepTests.groovy
@@ -98,6 +98,19 @@ class BuildPluginStepTests extends BaseTest {
   }
 
   @Test
+  void test_getConfigurations_implicit_with_latest_jenkinsVersions() throws Exception {
+    def script = loadScript(scriptName)
+    def configurations = script.getConfigurations(jenkinsVersions: ['LATEST', 'LATEST_LTS'])
+    assertEquals(configurations.size, 4)
+    URL latest = new URL('https://updates.jenkins.io/current/latestCore.txt')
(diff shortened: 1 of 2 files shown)


jglick:
I think this is a bad idea as it means that CI builds could suddenly start failing for reasons unrelated to any code changes in the plugin repository.

If you want in-repo CI to consistently run against recent core versions, it is possible to set up updatecli to offer PRs updating that version. (Dependabot is regularly used to keep tests up to date vs. plugin dependencies, via a `bom-2.*.x` import, but for technical reasons it is hard to do the same for Jenkins core.)

If you just want to detect regressions affecting your plugin’s tests as soon as possible as new core and plugin releases appear, you can add it to the managed set in `jenkinsci/bom`.

strangelookingnerd:
@jglick Thanks for the clarifications! I created https://github.com/jenkinsci/bom/pull/1406 to add my plugin to the managed set.

PR #465, opened 2022-09-05
feat(buildDockerAndPublishImage): add 'unstash' parameter

lemeurherve:
Allow to unstash artifacts generated previously, like the .jar of a Java app we want to make a Docker image for. (Ex: https://github.com/jenkins-infra/plugin-health-scoring/pull/36)

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -14,7 +14,7 @@ def call(String imageShortName, Map userConfig=[:]) {
     gitCredentials: 'github-app-infra', // Credential ID for tagging and creating release
     imageDir: '.', // Relative path to the context directory for the Docker build
     registryNamespace: '', // Empty by default (means "autodiscover based on the current controller")
-    skipCheckout: false, // Allow to skip the `checkout scm` instruction in case it's already been done previously and we want to keep generated artifacts
+    unstash: '', // Allow to unstash files if not empty
   ]
 
   // Merging the 2 maps - https://blog.mrhaki.com/2010/04/groovy-goodness-adding-maps-to-map_21.html
@@ -63,7 +63,9 @@ def call(String imageShortName, Map userConfig=[:]) {
       infra.withDockerPullCredentials{
         String nextVersion = ''
         stage("Prepare ${imageName}") {
-          if (!finalConfig.skipCheckout) {
+          if (finalConfig.unstash != '') {
+            unstash finalConfig.unstash
+          } else {
             checkout scm
           }
 
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -14,6 +14,7 @@ def call(String imageShortName, Map userConfig=[:]) {
     gitCredentials: 'github-app-infra', // Credential ID for tagging and creating release
     imageDir: '.', // Relative path to the context directory for the Docker build
     registryNamespace: '', // Empty by default (means "autodiscover based on the current controller")
+    skipCheckout: false, // Allow to skip the `checkout scm` instruction in case it's already been done previously and we want to keep generated artifacts
   ]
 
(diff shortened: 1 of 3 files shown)


vars/buildDockerAndPublishImage.groovy
```diff
@@ -62,7 +63,11 @@ def call(String imageShortName, Map userConfig=[:]) {
       infra.withDockerPullCredentials{
         String nextVersion = ''
         stage("Prepare ${imageName}") {
-          checkout scm
+          if (finalConfig.unstash != '') {
+            unstash finalConfig.unstash
+          } else {
+            checkout scm
+          }
```
lemeurherve, line 70:
```suggestion
          checkout scm
          if (finalConfig.unstash != '') {
            unstash finalConfig.unstash
          }
```
Keeping the checkout so no one would be bitten by this.

vars/buildDockerAndPublishImage.groovy
```diff
@@ -14,6 +14,7 @@ def call(String imageShortName, Map userConfig=[:]) {
     gitCredentials: 'github-app-infra', // Credential ID for tagging and creating release
     imageDir: '.', // Relative path to the context directory for the Docker build
     registryNamespace: '', // Empty by default (means "autodiscover based on the current controller")
+    unstash: '', // Allow to unstash files if not empty
```
dduportal, line 17: Warning here: I'll approve the PR assuming you've tested it on real life and that it unblocks https://github.com/jenkins-infra/plugin-health-scoring/pull/36.

BUT: be aware that the `unstash` pipeline instruction can be specified with a map of arguments (instead of the short call with a string as sole argument).

Maybe worth setting the default value to empty and use Groovy magic (`if (finalConfig.unstash) { ...}`) so it would be transparent?

Better adding unit test for both cases

dduportal (dismissed):

vars/buildDockerAndPublishImage.groovy
dduportal: yes, my apologies I mis-wrote. `null` looks good (assuming you can write it with Groovy). don't forget to test in both "unit test execution engine" environement, and also inside a fully-fledged Jenkins instance.

lemeurherve: > be aware that the unstash pipeline instruction can be specified with a map of arguments (instead of the short call with a string as sole argument).

Do you have a reference to this map of args? I've only found this one saying the `unstash` command accepts a string: https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#unstash-restore-files-previously-stashed

dduportal: Ooooh you are correct, I mixed up with `unstash`.

Under the hood, the `unstash` keywrod can style accept a map with the key `name` of type String (as per the documentation). But it's unlikely that another would be added so you should go with the initial proposal with an empty String 👍

vars/buildDockerAndPublishImage.txt
```diff
@@ -15,6 +15,7 @@ The following arguments are available for this function:
   * String **gitCredentials**: (Optional, defaults to "") If "automaticSemanticVersioning" is set, name of the credential to use when tagging the git repository. Support user/password or GitHub App credential types.
   * String **imageDir**: (Optional, defaults to ".", the parent directory of the Dockerfile) Path to a directory to use as build context (Example: "docker/", "python/2.7").
   * String **registry**: (Optional, defaults to "" which defines the registry based on the current controller setup) Container registry to deploy this image to (Example: "ghcr.io", "python/2.7").
+  * Boolean **unstash**: (Optional, default to "") Allow to restore files from a previously saved stash if not empty.
```
smerle33, line 18: If I understand correctly, this is not a boolean but a string representing the name of the previously saved stash : https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#unstash-restore-files-previously-stashed

PR #478, opened 2022-09-22
chore(buildDockerAndPublishImage): add log info when unstashing file(s)

lemeurherve:


lemeurherve:
`unstash` already produces logs.

PR #489, opened 2022-10-12
feat(terraform) keep the agent when terraform apply fails

dduportal:

### DIFF
--- a/vars/terraform.groovy
+++ b/vars/terraform.groovy
@@ -1,6 +1,6 @@
 /**
-terraform.groovy: Please look at https://github.com/jenkins-infra/shared-tools/tree/main/terraform#jenkins-pipeline for documentation and usage.
-Note: this shared pipeline is tailored to Jenkins infrastructure usage and thus set some default values which might not be desirable for yours.
+ terraform.groovy: Please look at https://github.com/jenkins-infra/shared-tools/tree/main/terraform#jenkins-pipeline for documentation and usage.
+ Note: this shared pipeline is tailored to Jenkins infrastructure usage and thus set some default values which might not be desirable for yours.
  **/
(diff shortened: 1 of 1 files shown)


dduportal:
Tested with success

vars/terraform.groovy
```diff
@@ -163,7 +163,13 @@ def call(userConfig = [:]) {
             }
             if (!isBuildCauseTimer && isBuildOnProductionBranch) {
               stage('🚢 Shipping Changes') {
-                sh makeCliCmd + ' deploy'
+                try {
+                  sh makeCliCmd + ' deploy'
+                } catch(Exception e) {
+                  // If the deploy failed, keep the pod until a user catch the problem (cloud be an errored state, or many reason to keep the workspace)
+                  input message: 'An error happened while applying the terraform plan. Keeping the agent up and running. Delete the agent?'
```
hervelemeur, line 170: I would save the error.tfstate as artifact here too, wdyt?

hervelemeur: And block subsequent builds (by not allowing builds if there is another running?)

dduportal: >  I would save the error.tfstate as artifact here too, wdyt?

Nope: this files contains sensitive datas so it should be kept in the ephemeral agents to avoid any accidental exposure

dduportal: > And block subsequent builds (by not allowing builds if there is another running?)

Good idea 🤔 I wonder if we should not use a `lock()` in this case (that would be released when the user interacts). WDYT?

lemeurherve: Idk the details of the lock, yes maybe? What would be locked?

PR #499, opened 2022-10-19
feat: implement `artifact-caching-proxy` in `buildPlugin`

lemeurherve:
This PR add an `artifactCachingProxyEnabled` parameter to the `buildPlugin` function to allow using one of [the artifact caching proxies](https://github.com/jenkins-infra/helm-charts/tree/main/charts/artifact-caching-proxy) deployed on each provider ([AWS](https://github.com/jenkins-infra/kubernetes-management/blob/1b11606e66822bcb60d31e3fcd23fafdc7b92d13/clusters/eks-public.yaml#L91-L101), [Azure](https://github.com/jenkins-infra/kubernetes-management/blob/1b11606e66822bcb60d31e3fcd23fafdc7b92d13/clusters/prodpublick8s.yaml) and [DigitalOcean](https://github.com/jenkins-infra/kubernetes-management/blob/1b11606e66822bcb60d31e3fcd23fafdc7b92d13/clusters/doks-public.yaml#L59-L69)) by specifying a Maven settings file with the basic auth required to access a artifact caching proxy (they all have the same credentials).

The choice of the proxy provider depends on the `ARTIFACT_CACHING_PROXY` environment variable [defined on every agent](https://github.com/jenkins-infra/jenkins-infra/pull/2351), except the Azure VM ones as the corresponding plugin doesn't allow setting env var. (This is also why the default artifact caching proxy provider is [set to `azure`](https://github.com/jenkins-infra/pipeline-library/pull/499/files#diff-db2794773b738dbb5566313f115cdf3ff1f45e13e57cbc37aa1bd181819ed5fbR113))

Supersedes #458 & #498 as [no Maven settings encryption is needed](https://github.com/jenkins-infra/pipeline-library/pull/498#discussion_r999413158).

For now I'm putting this functionality as opt-in for testing.
When this testing will be done, I intend to put it as opt-out by modifying the default value of the `artifactCachingProxyEnabled` parameter from `false` to `true`.

Successfully tested on Linux and Windows: https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/44 & https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/PR-44/126/

Ref: https://github.com/jenkins-infra/helpdesk/issues/2752

PR #502, opened 2022-10-20
feat: implement `artifact-caching-proxy` in `buildPlugin` with config-file-provider

lemeurherve:
This PR add an `artifactCachingProxyEnabled` parameter to the `buildPlugin` function to allow using one of [the artifact caching proxies](https://github.com/jenkins-infra/helm-charts/tree/main/charts/artifact-caching-proxy) deployed on each provider ([AWS](https://github.com/jenkins-infra/kubernetes-management/blob/1b11606e66822bcb60d31e3fcd23fafdc7b92d13/clusters/eks-public.yaml#L91-L101), [Azure](https://github.com/jenkins-infra/kubernetes-management/blob/1b11606e66822bcb60d31e3fcd23fafdc7b92d13/clusters/prodpublick8s.yaml) and [DigitalOcean](https://github.com/jenkins-infra/kubernetes-management/blob/1b11606e66822bcb60d31e3fcd23fafdc7b92d13/clusters/doks-public.yaml#L59-L69)) by specifying a [Maven settings file](https://github.com/jenkins-infra/kubernetes-management/pull/3078/).

Unlike my previous attempts as generating myself the Maven settings file, I'm now using the [Config File Provider](https://plugins.jenkins.io/config-file-provider/) to patch the settings with the basic auth required to access an artifact caching proxy (they all have the same credentials). As a result, the logic here is simpler.

The choice of the proxy provider depends on the `ARTIFACT_CACHING_PROXY` environment variable [defined on every agent](https://github.com/jenkins-infra/jenkins-infra/pull/2351), except the Azure VM ones as the corresponding plugin doesn't allow setting env var. (This is also why the default artifact caching proxy provider is [set to `azure`](https://github.com/jenkins-infra/pipeline-library/pull/499/files#diff-db2794773b738dbb5566313f115cdf3ff1f45e13e57cbc37aa1bd181819ed5fbR113)).

Needs https://github.com/jenkins-infra/kubernetes-management/pull/3078 and the equivalent on https://github.com/jenkins-infra/jenkins-infra (incoming PR)

Supersedes https://github.com/jenkins-infra/pipeline-library/pull/499 (thanks @jglick for the advises and inspiration!)

Tested on [jenkinsci/jenkins-infra-test-plugin](https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/44) with [`useContainerAgent: false`](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fjenkins-infra-test-plugin/detail/PR-44/131/pipeline) and [`useContainerAgent: true`](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fjenkins-infra-test-plugin/detail/PR-44/132/pipeline)

Ref: https://github.com/jenkins-infra/helpdesk/issues/2752

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -12,6 +12,8 @@ def call(Map params = [:]) {
   def failFast = params.containsKey('failFast') ? params.failFast : true
   def timeoutValue = params.containsKey('timeout') ? params.timeout : 60
   def gitDefaultBranch = params.containsKey('gitDefaultBranch') ? params.gitDefaultBranch : null
+  // This functionnality is set as opt-in for now
+  def artifactCachingProxyEnabled = params.containsKey('artifactCachingProxyEnabled') ? params.artifactCachingProxyEnabled : false
 
   def useContainerAgent = params.containsKey('useContainerAgent') ? params.useContainerAgent : false
   if (params.containsKey('useAci')) {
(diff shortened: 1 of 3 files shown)


vars/buildPlugin.groovy
```diff
@@ -121,7 +123,28 @@ def call(Map params = [:]) {
...
+                      String requestedProvider = env.ARTIFACT_CACHING_PROXY_PROVIDER
+                      if (requestedProvider == null || requestedProvider == '' || !availableProxyProviders.contains(requestedProvider)) {
+                        requestedProvider = defaultProxyProvider
+                        echo "INFO: no valid artifact caching proxy provider specified, set to '$defaultProxyProvider' by default."
+                      } else {
+                        echo "INFO: using artifact caching proxy from '${requestedProvider}' provider."
+                      }
+
+                      configFileProvider(
+                          [configFile(fileId: "artifact-caching-proxy-${requestedProvider}", variable: 'MAVEN_SETTINGS')]) {
```
lemeurherve, line 139: These files and the associated credentials to use are defined here: https://github.com/jenkins-infra/kubernetes-management/blob/3cf7d326badfce9deb6946e7cfc2956287b833d1/config/ext_jenkins-infra.yaml#L383-L464

jglick: I.e., https://github.com/jenkins-infra/kubernetes-management/pull/3078. Makes sense to define those things in infra rather than this library I guess.

jglick (dismissed):
Seems straightforward.

dduportal (dismissed):
nice job 👏

vars/buildPlugin.groovy
```diff
@@ -121,7 +123,28 @@ def call(Map params = [:]) {
...
+                            infra.runMaven(mavenOptions, jdk, null, env.MAVEN_SETTINGS, addToolEnv)
+                          }
+                    } else {
+                      echo 'NOTICE: artifacts will be downloaded directly from https://repo.jenkins-ci.org.'
```
timja, line 143: probably don't need to spam builds with this? but doesn't matter much

timja (approved):

vars/buildPlugin.groovy
lemeurherve: You're right it doesn't make sense to add this line while the functionality is opt-in, neither to nag people who would have opted out when it's switched to opt-out.

timja (dismissed):

timja (approved):

PR #505, opened 2022-10-25
feat(artifact-caching-proxy): retrieve availables providers from an env var

lemeurherve:
Avoid hardcoding them, easier to disable one if needed for maintenance or other events.

Follow-up of #502 and #504

Ref: https://github.com/jenkins-infra/helpdesk/issues/3203 & https://github.com/jenkins-infra/helpdesk/issues/2752

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -132,7 +132,7 @@ def call(Map params = [:]) {
                       final String configuredAvailableProxyProviders = env.ARTIFACT_CACHING_PROXY_AVAILABLE_PROVIDERS
                       if (configuredAvailableProxyProviders != null && configuredAvailableProxyProviders != '') {
                         final String availableProxyProviders = configuredAvailableProxyProviders.split(',')
-                        // Configure Maven settings if the requested provider is valid an available
+                        // Configure Maven settings if the requested provider is valid and available
                         if (validProxyProviders.contains(requestedProxyProvider) && availableProxyProviders.contains(requestedProxyProvider)) {
                           echo "INFO: using artifact caching proxy from '${requestedProxyProvider}' provider."
                           usingArtifactCachingProxy = true
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -143,7 +143,7 @@ def call(Map params = [:]) {
                         echo "INFO: using artifact caching proxy from '${requestedProvider}' provider."
                       }
 
-                      if (availableProxyProviders.contains(requestedProvider))
+                      if (availableProxyProviders.contains(requestedProvider)) {
                         noArtifactCachingProxyAvailable = false
                         configFileProvider(
                             [configFile(fileId: "artifact-caching-proxy-${requestedProvider}", variable: 'MAVEN_SETTINGS')]) {
@@ -153,7 +153,7 @@ def call(Map params = [:]) {
                         echo "WARNING: there is no available artifact caching proxy provider corresponding to '${requestedProvider}'."
                       }
                     }
-                    if (noArtifactCachingProxyAvailable)
+                    if (noArtifactCachingProxyAvailable) {
                       infra.runMaven(mavenOptions, jdk, null, null, addToolEnv)
                     }
                   } finally {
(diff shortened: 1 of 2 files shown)


vars/buildPlugin.groovy
```diff
@@ -123,23 +123,38 @@ def call(Map params = [:]) {
...
+                        } else {
+                          echo "INFO: using artifact caching proxy from '${requestedProvider}' provider."
+                        }
+
+                        if (availableProxyProviders.contains(requestedProvider)) {
+                          noArtifactCachingProxyAvailable = false
+                          configFileProvider(
+                              [configFile(fileId: "artifact-caching-proxy-${requestedProvider}", variable: 'MAVEN_SETTINGS')]) {
+                                infra.runMaven(mavenOptions, jdk, null, env.MAVEN_SETTINGS, addToolEnv)
+                              }
```
dduportal, line 149: - Since we introduce a fallback to repo.jenkins.io, the `defaultProxyProvider` does not make sense anymore. WDT about removing the "default" OR set it to `repo.jenkins-ci.org` (both choices would simplify the code path here)
- WDYT about the following simplified workflow:
  1. if env.ARTIFACT_CACHING_PROXY_PROVIDER is not empty AND in env.ARTIFACT_CACHING_PROXY_AVAILABLE_PROVIDERS.split(','), then test it. If test work then use it.
  2. otherwise use repo.jenkins-ci.org

lemeurherve: Having a defaultProvider set to azure is important as Azure VM agents can't be set with an env var.

dduportal: Can you add a comment then to explain why?

Otherwise WDYT of `requestProvider = env.ARTIFACT_CACHING_PROXY_PROVIDER ?: 'azure'` (groovy-isme -  https://docs.groovy-lang.org/latest/html/documentation/#_elvis_operator)

dduportal: note for later: we could set the value on Azure VM through the "Initialization Script", along with @smerle33 's datadog setup, by writing the value to `/home/jenkins/.profile` (or later `.bashrc` or whatever is eavluated in the context of a pipeline). WDYT @timja ?

lemeurherve: > Can you add a comment then to explain why?

There is already one L128 🙂 

> WDYT of requestProvider = env.ARTIFACT_CACHING_PROXY_PROVIDER ?: 'azure'

Looks nice!

timja: it should be do-able through the init script

dduportal: > even just detecting the vm is in Azure

Oh, do we have this feature in pipeline lanugage?

vars/buildPlugin.groovy
```diff
@@ -123,23 +123,31 @@ def call(Map params = [:]) {
...
-                        echo "INFO: no valid artifact caching proxy provider specified, set to '$defaultProxyProvider' by default."
+                      // As the env var ARTIFACT_CACHING_PROXY_PROVIDER can't be set on Azure VM agents,
+                      // we're using 'azure' as default provider if none is specified
+                      final String requestedProxyProvider = env.ARTIFACT_CACHING_PROXY_PROVIDER ?: 'azure'
+                      final String validProxyProviders = ['aws', 'azure', 'do']
+                      final String configuredAvailableProxyProviders = env.ARTIFACT_CACHING_PROXY_AVAILABLE_PROVIDERS
+                      if (configuredAvailableProxyProviders != null && configuredAvailableProxyProviders != '') {
+                        final String availableProxyProviders = configuredAvailableProxyProviders.split(',')
+                        // Configure Maven settings if the requested provider is valid and available
+                        if (validProxyProviders.contains(requestedProxyProvider) && availableProxyProviders.contains(requestedProxyProvider)) {
```
lemeurherve, line 136: Note: with this structure, the next step for having a fallback if the provider doesn't respond to an health check request is simple:

```suggestion
                        // Configure Maven settings if the requested provider is valid, available and respond to an health check
                        if (validProxyProviders.contains(requestedProxyProvider)
                            && availableProxyProviders.contains(requestedProxyProvider)
                            && sh(script: "curl --fail https://repo.${requestedProxyProvider}.jenkins.io/healthz", returnStatus: true) == 0
                            ) {
```
I'll open a PR as soon as this one is merged.

dduportal (approved):
LGTM 👍

PR #507, opened 2022-10-28
feat(buildPlugin): fallback to repo.jenkins-ci.org if the artifact caching proxy healcheck fails

lemeurherve:
Follow-up of https://github.com/jenkins-infra/pipeline-library/pull/502 and https://github.com/jenkins-infra/helm-charts/pull/279, this PR add a simple health check request on the requested provider before using it, and fallback to repo.jenkins-ci.org if there is an issue.

Note: `curl` is available on all our Linux and Windows agents.

Ref: https://github.com/jenkins-infra/helpdesk/issues/2752

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -128,7 +128,7 @@ def call(Map params = [:]) {
                       // As the env var ARTIFACT_CACHING_PROXY_PROVIDER can't be set on Azure VM agents,
                       // we're using 'azure' as default provider if none is specified
                       final String requestedProxyProvider = env.ARTIFACT_CACHING_PROXY_PROVIDER ?: 'azure'
-                      final String validProxyProviders = ['aws', 'azure', 'do']
+                      final String[] validProxyProviders = ['aws', 'azure', 'do']
                       final String configuredAvailableProxyProviders = env.ARTIFACT_CACHING_PROXY_AVAILABLE_PROVIDERS
                       if (configuredAvailableProxyProviders != null && configuredAvailableProxyProviders != '') {
                         final String[] availableProxyProviders = configuredAvailableProxyProviders.split(',')
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -134,12 +134,23 @@ def call(Map params = [:]) {
                         final String availableProxyProviders = configuredAvailableProxyProviders.split(',')
                         // Configure Maven settings if the requested provider is valid and available
                         if (validProxyProviders.contains(requestedProxyProvider) && availableProxyProviders.contains(requestedProxyProvider)) {
-                          echo "INFO: using artifact caching proxy from '${requestedProxyProvider}' provider."
-                          usingArtifactCachingProxy = true
-                          configFileProvider(
-                              [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
-                                infra.runMaven(mavenOptions, jdk, null, env.MAVEN_SETTINGS, addToolEnv)
-                              }
+                          boolean healthCheckOK = false
+                          final String healthCheckScript = "curl --fail https://repo.${requestedProxyProvider}.jenkins.io/healthz"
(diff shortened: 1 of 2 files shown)


vars/buildPlugin.groovy
```diff
@@ -134,12 +134,23 @@ def call(Map params = [:]) {
...
                         // Configure Maven settings if the requested provider is valid and available
                         if (validProxyProviders.contains(requestedProxyProvider) && availableProxyProviders.contains(requestedProxyProvider)) {
-                          echo "INFO: using artifact caching proxy from '${requestedProxyProvider}' provider."
-                          usingArtifactCachingProxy = true
-                          configFileProvider(
-                              [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
-                                infra.runMaven(mavenOptions, jdk, null, env.MAVEN_SETTINGS, addToolEnv)
-                              }
+                          boolean healthCheckOK = false
+                          final String healthCheckScript = "curl --fail https://repo.${requestedProxyProvider}.jenkins.io/healthz"
```
jglick, line 138: Note that this will print a warning because you are interpolating a Groovy variable in a `sh` arg.

jglick:
(BTW force pushes inside a PR break GH review flow.)

lemeurherve:
> (BTW force pushes inside a PR break GH review flow.)

Sorry about that, I though I had the time to push my last fix before you started, my mistake.

vars/buildPlugin.groovy
lemeurherve: Should I instead set an env var with the provider URL to use it latter with `sh` and `bat`?

vars/buildPlugin.groovy
```diff
@@ -134,12 +134,23 @@ def call(Map params = [:]) {
...
+                          if (isUnix()) {
+                            healthCheckOK = sh(script: healthCheckScript, returnStatus: true) == 0
+                          } else {
+                            healthCheckOK = bat(script: healthCheckScript, returnStatus: true) == 0
+                          }
```
lemeurherve, line 143: Something like that would be OK?
```suggestion
                          env.ARTIFACT_CACHING_PROXY_URL = "https://repo.${requestedProxyProvider}.jenkins.io/healthz"
                          if (isUnix()) {
                            healthCheckOK = sh(script: 'curl --fail $ARTIFACT_CACHING_PROXY_URL', returnStatus: true) == 0
                          } else {
                            healthCheckOK = bat(script: 'curl --fail $ARTIFACT_CACHING_PROXY_URL', returnStatus: true) == 0
                          }
```

jglick: Prefer `withEnv` (scoped). `env.XXX = yyy` syntax should not normally be used.

Not a big deal, just FYI.

vars/buildPlugin.groovy
```diff
@@ -128,18 +128,30 @@ def call(Map params = [:]) {
...
-                          configFileProvider(
-                              [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
-                                infra.runMaven(mavenOptions, jdk, null, env.MAVEN_SETTINGS, addToolEnv)
-                              }
+                          boolean healthCheckOK = false
+                          withEnv(["HEALTHCHECK=https://repo.${requestedProxyProvider}.jenkins.io/healthz"]) {
+                            if (isUnix()) {
+                              healthCheckOK = sh(script: 'curl --fail $HEALTHCHECK', returnStatus: true) == 0
+                            } else {
+                              healthCheckOK = bat(script: 'curl --fail $HEALTHCHECK', returnStatus: true) == 0
```
dduportal, line 142: Do we have `curl` on the windows machines? 🤔

dduportal (approved):
🚢

vars/buildPlugin.groovy
lemeurherve: Yes, tested on a VM with the images we're using.
Also noted in the PR body 🙂

PR #511, opened 2022-11-01
feat(buildPlugin): set default available artifact caching proxy providers to valid ones

lemeurherve:
So we only have to set a global env var on the controller when we need to restrict available providers.

Follow-up of #505

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -128,8 +128,9 @@ def call(Map params = [:]) {
                       // As the env var ARTIFACT_CACHING_PROXY_PROVIDER can't be set on Azure VM agents,
                       // we're using 'azure' as default provider if none is specified
                       final String requestedProxyProvider = env.ARTIFACT_CACHING_PROXY_PROVIDER ?: 'azure'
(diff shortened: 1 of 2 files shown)


vars/buildPlugin.groovy
```diff
@@ -129,22 +129,19 @@ def call(Map params = [:]) {
...
-                          echo "INFO: using artifact caching proxy from '${requestedProxyProvider}' provider."
-                          usingArtifactCachingProxy = true
-                          configFileProvider(
-                              [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
-                                infra.runMaven(mavenOptions, jdk, null, env.MAVEN_SETTINGS, addToolEnv)
-                              }
-                        } else {
-                          echo "WARNING: invalid or unavailable artifact caching proxy provider '${requestedProxyProvider}' specified, will use repo.jenkins-ci.org"
-                        }
+                      // Useful when a provider is in maintainance (or similar cases), add a global env var in Jenkins controller settings to restrict them
```
lemeurherve, line 132: This diff is a bit easier to read without whitespaces changes.

dduportal (approved):

PR #525, opened 2022-11-16
Deleting JEP-309 utilities

### LINKED ISSUE #523: cleanup request: replace all the `readYaml()` occurences by an agent command (`yq`, `sed`, whatever)
The `readYaml()` feature is used on this repository as a pipelien step provided by https://github.com/jenkinsci/pipeline-utility-steps-plugin/blob/master/docs/STEPS.md.

But this step runs on the controller, wasting precious resources that could better be used on other tasks such as... handlin build orchestration and agents.

The request is to replace the `readYaml()` occurence by a call to an agent with command line tools such as `yq` (which is installed on all Linux agents and Windows VM agents).

jglick:
Fixes #523 by deleting #30 and more. First https://github.com/jenkinsci/jenkins/pull/7397 needs to be merged, and backported to the LTS branch. As far as I can tell from code searches, nothing else is using `runATH`, and nothing whatsoever is using `runPCT`.


lemeurherve (approved):
Many thanks!

PR #529, opened 2022-11-22
Removing `customWARPackager`

jglick:
Follows up #525: https://github.com/jenkins-infra/pipeline-library/issues/523#issuecomment-1323989879

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,12 +59,6 @@
       <version>${jenkins-pipeline-unit.version}</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.yaml</groupId>
-      <artifactId>snakeyaml</artifactId>
-      <version>1.33</version>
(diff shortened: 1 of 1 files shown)


test/groovy/BaseTest.groovy
```diff
@@ -1,10 +1,8 @@
 import com.lesfurets.jenkins.unit.declarative.DeclarativePipelineTest
 import mock.CurrentBuild
-import mock.CustomWARPackager
 import mock.Infra
 import org.junit.Before
 import org.junit.Test
-import org.yaml.snakeyaml.Yaml
```
basil, line 7: Is https://github.com/jenkins-infra/pipeline-library/blob/cd87a074cbdcdb3072f7ee651afa5db7c8fdba40/pom.xml#L62-L67 still needed after the deletion of all its callers?

basil (approved):

PR #530, opened 2022-11-24
Ensure gradlew permissions are sufficient to be executable on Linux

NotMyFault:
Generating the wrapper ships a low permission set, often not being sufficient when building on Linux or not being detected at all, e.g., you are using Windows and are not prone to file permissions.
The change proposed makes the wrapper executable for the run and doesn't interrupt the workflow.

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -178,11 +178,14 @@ def call(Map params = [:]) {
                   if (skipTests) {
                     gradleOptions += '--exclude-task test'
                   }
+                  File gradleWrapper = new File("${pwd()}/gradlew")
+                  if (gradleWrapper.exists() && isUnix() && !gradleWrapper.canExecute()) {
+                    gradleWrapper.setExecutable(true)
+                  }
                   command = "gradlew ${gradleOptions.join(' ')}"
                   if (isUnix()) {
                     command = './' + command
                   }
-
                   try {
                     infra.runWithJava(command, jdk, null, addToolEnv)
                   } finally {


basil (commented):
This seems like it is a problem with the plugin maintainer not checking in the `gradlew` file to Git with the executable bit set. There are so few plugins that use Gradle to begin with, and I could not find any examples that were incorrect in my local checkout directory (I looked at `gradle-jpi-plugin`, `gradle-plugin`, `job-dsl-plugin`, and `synopsys-detect-plugin`), so I am reluctant to check in a workaround in common code for a problem that is really specific to individual plugins, and probably few if any at that. My vote is to close this PR and run the following command in any affected plugins:

```
chmod +x gradlew
git add gradlew
git commit
```

I would be OK with this PR if and only if it printed an error and then failed the Pipeline rather than setting the executable bit. That would change this PR from a workaround to an implementation of [fail-fast](https://en.wikipedia.org/wiki/Fail-fast) systems design. Failing fast is always good; workarounds are not.

jglick:
Anyway if you wanted to do this, do not do it in Pipeline script! Much simpler to change the Unix flavor from `./gradlew` to `sh gradlew`. Or `perl gradlew`! (https://askubuntu.com/a/850575 if you do not believe me)

jglick:
Anyway if you wanted to do this, do not do it in Pipeline script! Much simpler to change the Unix flavor from `./gradlew` to `sh gradlew`. Or `perl gradlew`! (https://askubuntu.com/a/850575 if you do not believe me)

NotMyFault:
> Anyway if you wanted to do this, do not do it in Pipeline script! Much simpler to change the Unix flavor from `./gradlew` to `sh gradlew`. Or `perl gradlew`! (https://askubuntu.com/a/850575 if you do not believe me)

I do believe you for sure :P

Thanks for the feedback, given this applies to a very small percent of the userbase only, we don't need to manipulate the wrapper on our end.

basil:
> Or `perl gradlew`! (https://askubuntu.com/a/850575 if you do not believe me)

Ha! But shebang lines from [Seventh Edition Unix and later](https://utcc.utoronto.ca/~cks/space/blog/unix/ExecAndShebangHistory) are so overrated: all you need to do is make the shell script executable in order to hit [this pre-Seventh Edition compatibility code path](https://github.com/bminor/bash/blob/c7bb74de78058f6381d54148465d42e8370e075a/execute_cmd.c#L6022-L6103). I joke, I joke. :smile:

PR #541, opened 2022-12-07
breaking(buildPlugin) switch default JDK from 8 to 11

dduportal:
As per the discussion on https://groups.google.com/g/jenkinsci-dev/c/pjfvsMw-EMM/m/Y7JhtkBsAQAJ?utm_medium=email&utm_source=footer, since https://github.com/jenkinsci/plugin-pom/releases/tag/plugin-4.52 has been released with JDK 8 being deprecated, we want to ensure the best contributor experience and use JDK11 by default for any default plugin build configuration (e.g. when calling `buildPlugin()` as per https://www.jenkins.io/doc/developer/tutorial-improve/add-a-jenkinsfile/.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -34,7 +34,7 @@ buildPlugin()
 * `failFast` (default: `true`) - instruct Maven tests to fail fast
 * `platforms` (default: `['linux', 'windows']`) - Labels matching platforms to
   execute the steps against in parallel
-* `jdkVersions` (default: `[8]`) - JDK version numbers, must match a version
+* `jdkVersions` (default: `[11]`) - JDK version numbers, must match a version
   number jdk tool installed
 * `jenkinsVersions`: (default: `[null]`) - a matrix of Jenkins baseline versions to build/test against in parallel (null means default,
   only available for Maven projects)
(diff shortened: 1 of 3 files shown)


basil (dismissed):

timja (approved):

basil (approved):

jglick:
Not sure this is wise. As of https://github.com/jenkinsci/archetypes/pull/546 we explicitly specify configurations for new plugins (and as of https://github.com/jenkinsci/archetypes/pull/547 we even test against JDK 17). @jtnord pointed out that switching to JDK 11 may cause build failures for plugins with older parent POMs. Seems safer and clearer to
1. Leave the behavior of `buildPlugin` without `configurations` (or I guess an element of `configurations` without a `jdk` parameter) untouched, but emit a warning.
1. Update the docs to match the current archetype behavior.
1. Fix https://issues.jenkins.io/browse/JENKINS-46795 and ship to ci.jenkins.io.
1. Use the magic tool to propose PRs to all plugins not yet using the explicit style.

The first two should be easy and something we can do now. The third I think I know how to do, though I lack time right now.

I am not blocking this PR if others feel it is best to just go ahead and do it; obviously it does not preclude the other proposals.

PR #573, opened 2023-01-30
chore(buildPlugin): revert hotfix setting the default artifact-caching-proxy provider to 'aws' instead of 'azure'

lemeurherve:
Revert 10e7c83


dduportal (approved):
As per https://github.com/jenkins-infra/helpdesk/issues/2752#issuecomment-1409006477, the Azure ACP is now performing as expected. So LGTM 👍

PR #576, opened 2023-02-01
chore(buildPlugin): better name for the artifact caching proxy activation parameter

lemeurherve:


dduportal (approved):
LGTM if tested :)

lemeurherve:
Tested successfully with the deprecated parameter name and the new one, merging.

PR #583, opened 2023-02-08
feat: move artifact caching proxy logic to its own `infra.withArtifactCachingProxy` function

lemeurherve:
This will allow easier integration for other builds like the BOM

Tested in https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/64 and https://github.com/jenkinsci/bom/pull/1741

Ref: https://github.com/jenkins-infra/helpdesk/issues/2752

PR #584, opened 2023-02-09
github app credentials lookup

lemeurherve:


vars/infra.groovy
```diff
@@ -103,6 +104,91 @@ boolean retrieveMavenSettingsFile(String settingsXml) {
...
+
+  // Check if the requested provider is invalid or unavailable
+  if (!validProxyProviders.contains(requestedProxyProvider) || !availableProxyProviders.contains(requestedProxyProvider)) {
+    echo "WARNING: invalid or unavailable artifact caching proxy provider '${requestedProxyProvider}' requested by the agent, will use repo.jenkins-ci.org"
+    useArtifactCachingProxy = false
+  }
+
+  // If the build concerns a pull request, check if there is "skip-artifact-caching-proxy" label applied in case the user doesn't want ACP
+  if ((useArtifactCachingProxy) && (!env.BRANCH_IS_PRIMARY)) {
+    final String currentCredentialsId = SCMSource.SourceByItem.findSource(currentBuild.rawBuild.parent).credentialsId
```
jglick, line 138: I do not recommend doing this except inside a dedicated `@NonCPS` method, since you do not want to try to serialize the `SCMSource` object inside the program data.

jglick (commented):
https://github.com/jenkins-infra/pipeline-library/pull/584/checks?check_run_id=11223410858 looks like you would need to mock out the newly extracted method somehow.

vars/infra.groovy
timja: just do this?
```
@NonCPS
def getBuildCredentialsId() {
  return SCMSource.SourceByItem.findSource(currentBuild.rawBuild.parent).credentialsId
}
```

jglick: and then see if you can override it in https://github.com/jenkins-infra/pipeline-library/blob/07e4c9f5e5c000bfd2b995a381973e395d126b51/test/groovy/mock/Infra.groovy#L4

lemeurherve:
Successfully tested in https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/64/, merged in #583

PR #596, opened 2023-02-17
feat(artifact-caching-proxy): ensure ARTIFACT_CACHING_PROXY is always set

lemeurherve:
Ensuring this environment variable is always set even on Azure VM agents so it can be safely used in every case.


smerle33 (approved):
LGTM

PR #605, opened 2023-02-24
chore: pin GitHub actions with their SHA-1 instead of their version number

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/3402

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -1,42 +1,42 @@
 name: updatecli
 on:
-    workflow_dispatch:
-    schedule:
-        # Every monday
-        - cron: '* * * * 1'
-    push:
-    pull_request:
+  workflow_dispatch:
+  schedule:
+    # Every monday
+    - cron: '* * * * 1'
+  push:
+  pull_request:
 jobs:
-    updatecli:
-        runs-on: ubuntu-latest
-        steps:
-            - name: Checkout
-              uses: actions/checkout@v3
-            - name: Install updatecli CLI
-              env:
-                UPDATECLI_VERSION: "v0.44.4"
-              run: |
-                curl --silent --location --show-error --output /tmp/updatecli.tgz \
-                  "https://github.com/updatecli/updatecli/releases/download/${UPDATECLI_VERSION}/updatecli_$(uname -s)_$(uname -m).tar.gz"
-                tar xzf /tmp/updatecli.tgz -C /usr/local/bin/ updatecli
-                rm -f /tmp/updatecli.tgz
-            - name: Diff
-              continue-on-error: true
-              run: |
-                updatecli diff --config ./updatecli/updatecli.d --values ./updatecli/values.yaml
-              env:
-                UPDATECLI_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-            ## This step allows to generate a short-lived token which is allowed to modify GitHub Workflow files (that the default GITHUB_TOKEN cannot)
-            ## Please make sure that the 2 secrets are defined (usually at organization level, with a per-repository access)
-            - uses: tibdex/github-app-token@v1.8
-              id: generate_token
-              if: github.ref == 'refs/heads/master'
-              with:
-                app_id: ${{ secrets.JENKINS_ADMIN_APP_ID }}
-                private_key: ${{ secrets.JENKINS_ADMIN_APP_PRIVKEY }}
-            - name: Apply
-              if: github.ref == 'refs/heads/master'
-              run: |
-                updatecli apply --config ./updatecli/updatecli.d --values ./updatecli/values.yaml
-              env:
-                UPDATECLI_GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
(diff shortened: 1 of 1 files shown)


jglick (commented):
My acid test for using a specific tag/hash in a GHA version is whether pinning it will actually do what you think. If a Dependabot PR is opened which bumps the version, and the change was actually incompatible, will there be a failing check? AFAICT in this case that will be true of the `actions/checkout` version due to https://github.com/jenkins-infra/pipeline-library/blob/b5feeb63887138138c390b8d116186a0f9be07b1/.github/workflows/updatecli.yaml#L8 but not true of the `tibdex/github-app-token` version due to https://github.com/jenkins-infra/pipeline-library/blob/b5feeb63887138138c390b8d116186a0f9be07b1/.github/workflows/updatecli.yaml#L33-L38 That is why we do not and should not pin https://github.com/jenkinsci/.github/blob/632be9b6144191d66bfc43c0bc6685f6aaacb15d/workflow-templates/cd.yaml#L12 for example: the version is only used on trunk, so a proposed change to the version would have no effect on a PR build and if the change was wrong, by the time you find out the PR has already been merged.

jglick:
To be clear, my understanding in https://github.com/jenkins-infra/pipeline-library/pull/605#pullrequestreview-1313336677 from source inspection (not experiment!) is that a specific version of `checkout` _would_ be verified by DB PRs whereas a specific version of `github-app-token` would _not_, which seems the opposite of https://github.com/jenkins-infra/pipeline-library/pull/605#event-8601765573.

(BTW it is hard for a reviewer to follow what is happening when you force-push inside the PR branch. It really only ever makes sense to do so in a PR which you expect to be rebase-merged, or merged outside GitHub via `git merge --fast-forward PR-branch`. If the PR is going to be squash-merged, you may as well keep a legible history inside the branch. If it is going to be true-merged, presumably you are not concerned with linear history to begin with and are comfortable using `-m1` options etc. where appropriate.)

lemeurherve:
> BTW it is hard for a reviewer to follow what is happening when you force-push inside the PR branch. It really only ever makes sense to do so in a PR which you expect to be rebase-merged, or merged outside GitHub via git merge --fast-forward PR-branch. If the PR is going to be squash-merged, you may as well keep a legible history inside the branch. If it is going to be true-merged, presumably you are not concerned with linear history to begin with and are comfortable using -m1 options etc. where appropriate.)

Sorry about that, I'm aware of the trouble but since I reviewed my script to remove "ratchet: org/repo" from the comment, then to avoid modifying official GitHub actions, and as I'm using multi-gitter to create these PRs I had not other (good) options than to force-push these new changes.

lemeurherve:
> If a Dependabot PR is opened which bumps the version, and the change was actually incompatible, will there be a failing check? 

We don't merge PRs blindly, even/especially dependabot ones. Personally I'm checking release notes and/or corresponding commits when an update is proposed.

SHA-1 or not it doesn't change the fact there is no automatic verification of changes proposed by dependabot.

As I borked your review comments with the force-pushed commits I don't have the corresponding references, not sure about the rest of what you said (again, sorry)

jglick:
> I'm using multi-gitter to create these PRs

Ah OK, did not realize this was a batch-filed thing.

> I'm checking release notes

That should be enough then.

> there is no automatic verification of changes proposed by dependabot

Well, again it depends on whether the thing being changed is actually exercised during the check(s) run during the PR build. Genuine dependencies are certain checked, as are tools used in a regular way during the build. updatecli is just used in a special way.

dduportal:
> My acid test for using a specific tag/hash in a GHA version is whether pinning it will actually do what you think. If a Dependabot PR is opened which bumps the version, and the change was actually incompatible, will there be a failing check? AFAICT in this case that will be true of the `actions/checkout` version due to
> 
> https://github.com/jenkins-infra/pipeline-library/blob/b5feeb63887138138c390b8d116186a0f9be07b1/.github/workflows/updatecli.yaml#L8
> but not true of the `tibdex/github-app-token` version due to
> 
> https://github.com/jenkins-infra/pipeline-library/blob/b5feeb63887138138c390b8d116186a0f9be07b1/.github/workflows/updatecli.yaml#L33-L38
> That is why we do not and should not pin https://github.com/jenkinsci/.github/blob/632be9b6144191d66bfc43c0bc6685f6aaacb15d/workflow-templates/cd.yaml#L12 for example: the version is only used on trunk, so a proposed change to the version would have no effect on a PR build and if the change was wrong, by the time you find out the PR has already been merged.

That an excellent point. But there are 2 others reasons why we (Jenkins Infra team) tend to prefer having PRs even on  untested GHA:
- (important) it allows a control of "when" the dependency is bumped ("surprise" dependency upgrades tend to add mayhem, particulary if it happens during a core release or a security advisory)
- (minor) it helps to track that there is something new (clearly we have multiple and better sources but still it helps a lot due to the ability to fine-control GitHub notifications)

jglick:
Yes it can make sense for the transparency. It just requires extra attention from repo maintainers to remember that _this_ green DB PR is potentially dangerous.

PR #611, opened 2023-03-06
fix(artifact-caching-proxy): don't add settings.xml to Maven options if no ACP is configured

lemeurherve:
When a "skip-artifact-caching-proxy" label is added to a PR, or if the artifact caching proxy health check fails, the settings.xml shouldn't be added to Maven options.

This PR ensures that's the case, and add corresponding tests.

Tested in https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/66

Ref: https://github.com/jenkins-infra/helpdesk/issues/3422 & https://github.com/jenkins-infra/helpdesk/issues/3424

### DIFF
--- a/test/groovy/InfraStepTests.groovy
+++ b/test/groovy/InfraStepTests.groovy
@@ -393,9 +393,9 @@ class InfraStepTests extends BaseTest {
     assertTrue(isOK)
     // then an healthcheck is performed on the provider
     assertTrue(assertMethodCallContainsPattern('sh', healthCheckScriptSh) || assertMethodCallContainsPattern('bat', healthCheckScriptBat))
-    // then it notices the provider isn't reachable and that it will fallback to repo.jenkins-ci.org
+    // then it does not notice the provider isn't reachable and that it will fallback to repo.jenkins-ci.org
     assertFalse(assertMethodCallContainsPattern('echo', "WARNING: the artifact caching proxy from '${anotherArtifactCachingProxyProvider}' provider isn't reachable, will use repo.jenkins-ci.org"))
-    // then there is no call to configFile containing the requested artifact caching proxy provider id
+    // then there is a call to configFile containing the requested artifact caching proxy provider id
     assertTrue(assertMethodCallContainsPattern('configFile', "artifact-caching-proxy-${anotherArtifactCachingProxyProvider}"))
     // then it succeeds
     assertJobStatusSuccess()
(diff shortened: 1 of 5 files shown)


timja (approved):

test/groovy/InfraStepTests.groovy
```diff
@@ -425,4 +425,63 @@ class InfraStepTests extends BaseTest {
...
+    // when running with useArtifactCachingProxy set to true
+    script.runMaven(['clean verify'], 11, null, null, null, true)
+    printCallStack()
+    // then an healthcheck is performed on the provider
+    assertTrue(assertMethodCallContainsPattern('sh', healthCheckScriptSh) || assertMethodCallContainsPattern('bat', healthCheckScriptBat))
+    // then it notices the provider isn't reachable and that it will fallback to repo.jenkins-ci.org
+    assertTrue(assertMethodCallContainsPattern('echo', "WARNING: the artifact caching proxy from '${anotherArtifactCachingProxyProvider}' provider isn't reachable, will use repo.jenkins-ci.org"))
+    // then there is no call to configFile containing the requested artifact caching proxy provider id
+    assertFalse(assertMethodCallContainsPattern('configFile', "artifact-caching-proxy-${anotherArtifactCachingProxyProvider}"))
+    // then it succeeds
```
jglick, line 484: And without the `infra.groovy` patch, this test fails somehow because some assertion chokes on `-s null`?

jglick (approved):
Can we get this merged soon? It seems like every build on the server is currently broken, apparently due to this.

vars/infra.groovy
```diff
@@ -179,7 +179,11 @@ Object runMaven(List<String> options, String jdk = '8', List<String> extraEnv =
   List<String> mvnOptions = ['--batch-mode', '--show-version', '--errors', '--no-transfer-progress']
   if (useArtifactCachingProxy) {
     withArtifactCachingProxy {
-      mvnOptions += "-s $env.MAVEN_SETTINGS"
+      // If an artifact caching proxy provider has been correctly configured,
+      // add the corresponding settings.xml to Maven options
```
dduportal, line 183: Post-merge review (nitpick since the PR does the job): what is the relationship between artifact caching provider configured and the undefined/defined generic variable `env.MAVEN_SETTINGS`?

As I see the patch here, it's a validation to ensure that the variable is used only if defined: is that correct?

hervelemeur: env.MAVEN_SETTINGS is defined if an artifact caching proxy has been configured (ie not skipped and not failing any health check)

test/groovy/InfraStepTests.groovy
lemeurherve: Addressed in https://github.com/jenkins-infra/pipeline-library/pull/614

PR #614, opened 2023-03-06
fix(artifact-caching-proxy): correct `infra.runMaven` tests

lemeurherve:
Follow-up of #611

Tests actually testing the fix from #611, the previous ones didn't do anything to catch the `-s null` error and passed even without the `infra.runMaven` groovy patch, as noticed by @jglick in https://github.com/jenkins-infra/pipeline-library/pull/611#discussion_r1126400780

<details>
<summary>Initial PR body</summary>

Better way of passing the fact that an artifact caching proxy has been configured or not in `infra.withArtifactCachingProxy` to `infra.runMaven` with `artifactCachingProxyConfigured` (cf @dduportal comment https://github.com/jenkins-infra/pipeline-library/pull/611#discussion_r1126494592)

Tests actually testing the fix from #611, the previous ones didn't do anything to catch the `-s null` error and passed even without the `infra.runMaven` groovy patch, as noticed by @jglick in https://github.com/jenkins-infra/pipeline-library/pull/611#discussion_r1126400780

Note: I didn't manage to retrieve in the tests the value of the `MAVEN_SETTINGS` variable defined by the config file provider here: https://github.com/jenkins-infra/pipeline-library/blob/f943c7ba139787abdcfafef8aa6d1acfd6f84233/vars/infra.groovy#L160

Any idea how/if it can be retrieved?

Ideally I'd like to pass this value in the closure instead of `artifactCachingProxyConfigured`

</details>

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -158,10 +158,10 @@ Object withArtifactCachingProxy(Closure body) {
     echo "INFO: using artifact caching proxy from '${requestedProxyProvider}' provider."
     configFileProvider(
         [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
-          body()
+          body(artifactCachingProxyConfigured = true)
         }
   } else {
-    body()
+    body(artifactCachingProxyConfigured = false)
   }
 }
 
@@ -180,8 +180,9 @@ Object runMaven(List<String> options, String jdk = '8', List<String> extraEnv = 
   if (useArtifactCachingProxy) {
     withArtifactCachingProxy {
       // If an artifact caching proxy provider has been correctly configured,
-      // add the corresponding settings.xml to Maven options
-      if (env.MAVEN_SETTINGS) {
+      // add the corresponding settings.xml path stored in MAVEN_SETTINGS env var by the config file provider
+      // to Maven options
+      if (artifactCachingProxyConfigured) {
         mvnOptions += "-s $env.MAVEN_SETTINGS"
       }
       mvnOptions.addAll(options)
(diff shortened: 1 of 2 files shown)


vars/infra.groovy
```diff
@@ -180,8 +180,9 @@ Object runMaven(List<String> options, String jdk = '8', List<String> extraEnv =
   if (useArtifactCachingProxy) {
     withArtifactCachingProxy {
       // If an artifact caching proxy provider has been correctly configured,
-      // add the corresponding settings.xml to Maven options
-      if (env.MAVEN_SETTINGS) {
+      // add the corresponding settings.xml path stored in MAVEN_SETTINGS env var
+      // by the config file provider to Maven options
+      if (artifactCachingProxyConfigured) {
```
jglick, line 185: I do not understand what this variable refers to; it is not a method parameter?

lemeurherve: @jglick it's not a method parameter. This variable refers to the one I've passed to the closure body: https://github.com/jenkins-infra/pipeline-library/pull/614/files#diff-f2d02565a16bf12d02a3181ac4ce5b6e5c7fee017458b9226d517b11c724c89aR161-R164

As noted in the PR body above, I didn't find how to retrieve `env.MAVEN_SETTINGS` value (or the equivalent variable set by the config file provider) in the tests, and as it seems always `null`, I couldn't assess this `if (env.MAVEN_SETTINGS)` path in my tests (as it never goes into it) to check if this value is passed or not to Maven options.

I then searched how to pass a value from `withArtifactCachingProxy` to `runMaven`, and found out (mostly by chance) that a variable passed in `body(aVariable)` was accessible in `runMaven`. I'm not convinced at all it's the right way, but it worked 🤷 

As I don't know if there is a better way to pass a variable I've used it, but I must admit I'm not really comfortable with it.

To reassure you, I'm scratching this code. After discussing it with Damien I think I found a way to test this properly.

jglick (approved):
Looks more straightforward now.

dduportal (approved):
🚀  👏

PR #616, opened 2023-03-08
feat: add `infra.isRelease` function (#616)Co-authored-by: Damien Duportal <damien.duportal@gmail.com>

lemeurherve:
This PR will allow restricting certain steps/stages/commands in pipelines depending on their execution location, similar to `infra.isTrusted()` and `infra.isInfra()`.

I intend to use it to prevent the use of artifact caching proxies on release.ci and trusted.ci

### DIFF
--- a/src/io/jenkins/infra/InfraConfig.groovy
+++ b/src/io/jenkins/infra/InfraConfig.groovy
@@ -39,7 +39,7 @@ class InfraConfig implements Serializable {
 
   // Returns the Docker registry hostname which this instance has credentials for
   String getDockerRegistryNamespace() {
-    if (isTrusted() || isRelease() || isInfra()) {
+    if (isTrusted() || isInfra()) {
       return 'jenkinsciinfra'
     } else {
       return 'jenkins4eval'
--- a/src/io/jenkins/infra/InfraConfig.groovy
+++ b/src/io/jenkins/infra/InfraConfig.groovy
@@ -25,17 +25,21 @@ class InfraConfig implements Serializable {
     return isJenkinsURLcontains('trusted.ci')
   }
 
+  Boolean isRelease() {
+    return isJenkinsURLcontains('release.ci')
+  }
+
   Boolean isInfra() {
     return isJenkinsURLcontains('infra.ci')
   }
 
   Boolean isRunningOnJenkinsInfra() {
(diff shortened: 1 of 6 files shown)


src/io/jenkins/infra/InfraConfig.groovy
```diff
@@ -25,17 +25,21 @@ class InfraConfig implements Serializable {
...
 
   Boolean isRunningOnJenkinsInfra() {
-    return isCI() || isTrusted() || isInfra()
+    return isCI() || isTrusted() || isRelease() || isInfra()
   }
 
   // Returns the Docker registry hostname which this instance has credentials for
   String getDockerRegistryNamespace() {
-    if (isTrusted() || isInfra()) {
+    if (isTrusted() || isRelease() || isInfra()) {
```
lemeurherve, line 42: Not quite sure of this change, WDYT?

dduportal: this line should not be changed. Today, no Docker image is built on release.ci.jenkins.io.

If it ever happens, it would be `jenkinsci` namespace only, as the role of release.ci.jenkins.io is only to produce artifacts for the Jenkins users (e.g. Jenkins Core WAR, packages and eventuall Docker image) benefit.


`jenkinsciinfra` is a DockerHub namespace only aimed at the benefit of the Jenkins infrastructure: it's the role of infra.ci.jenkins.io.

vars/infra.txt
```diff
@@ -11,6 +11,12 @@
     Return true if this Pipeline is executing on a trusted CI environment.
 </p>
 
+<strong>infra.isRelease()</strong>
+
+<p>
+    Return true if this Pipeline is executing on a release CI environment.
```
dduportal, line 17:
```suggestion
    Return true if this Pipeline is executing on the Jenkins release CI environment.
```

dduportal (changes requested):
LGTM escept the docker namespace

src/io/jenkins/infra/InfraConfig.groovy
lemeurherve, line 42:
```suggestion
    if (isTrusted() || isInfra()) {
```

dduportal (approved):
LGTM (assuming you'll clean up the unit tests ;) )

PR #629, opened 2023-03-27
More clearly communicate to Gradle users that they are not on a fully supported path

basil:
In https://groups.google.com/g/jenkinsci-dev/c/lHQAiEepBiw/m/1efu_EUBBQAJ @lemeurherve notes that

> for those using `buildPluginWithGradle()` from the shared pipeline library, we could add a note about it in the logs

which makes sense to me. Here I am adding such a deprecation check. While I was here, I also cleaned up the preceding warning to use the GitHub Checks API rather than an `echo` step. To test this change I successfully ran `mvn clean verify` locally.


dduportal (approved):
Thanks!

PR #630, opened 2023-03-29
Add `launchable` step

basil:
It was requested that I do a prototype of Launchable. This provides me with functionality to install and run the Launchable CLI from within our CI system. I tested this on my local machine and on `ci.jenkins.io` by replaying `text-finder` plugin builds with both container-based and VM Linux agents. I did not implement the functionality for Windows agents because I determined that both our container-based and VM Windows agents do not have Python installed.

### DIFF
--- a/test/groovy/LaunchableStepTests.groovy
+++ b/test/groovy/LaunchableStepTests.groovy
@@ -1,0 +1,52 @@
+import static org.junit.Assert.assertTrue
+
+import org.junit.Before
+import org.junit.Test
+
+class LaunchableStepTests extends BaseTest {
+  static final String scriptName = 'vars/launchable.groovy'
+
+  @Override
+  @Before
+  void setUp() throws Exception {
+    super.setUp()
+  }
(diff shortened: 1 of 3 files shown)


basil:
Since I now have three open PRs that depend on this PR, as well as several more in development that depend on those PRs, I would appreciate a review and merge of this one.

dduportal (approved):
Nice job, thanks for the details on the PR descrption.

Note: I just learnt about the `launchable/bin/pip --require-virtualenv` form to avoid sourcing the virtual env `bin/activate`  script, really useful for non interactive scripts!

dduportal:
@basil does it make sense if the jenkins infra team takes care of installing python3 on the Windows machines?

Also, does it make sense if we install launchable on all templates?

basil:
Yes, it does make sense, and that would allow this code to be simplified to just a few lines. In the meantime, this rather basic solution helps me to get started with the prototype.

PR #552, opened 2022-12-22
feat(buildPlugin): skip ACP if a `skip-artifact-caching-proxy` label has been applied to the pull request

lemeurherve:
This PR add the possibility for contributors to disable the Artifact Caching Proxy temporary on their pull requests by applying a `skip-artifact-caching-proxy` label to them.

It comes in complement of the possibility for ci.jenkins.io admins to disable some or all ACP providers via the `ARTIFACT_CACHING_PROXY_AVAILABLE_PROVIDERS` global settings. (cf https://github.com/jenkins-infra/helpdesk/issues/3302#issuecomment-1357319280)

Ref: https://github.com/jenkins-infra/helpdesk/issues/2752

PR #635, opened 2023-03-31
chore(`infra.withArtifactCachingProxy`): use `pipeline-github-plugin` to retrieve PR labels

lemeurherve:
As suggested by @basil in https://github.com/jenkins-infra/pipeline-library/pull/552#discussion_r1152685801, this PR replaces the GitHub API `curl` call with a credentials by the use of the `pullRequest` object provided by https://github.com/jenkinsci/pipeline-github-plugin.

> > What would be the benefit(s) of switching to this method?
> 
> The current code uses a `@NonCPS` method to call `SCMSource.SourceByItem.findSource` against `currentBuild.rawBuild.parent`, and using `currentBuild.rawBuild` goes against Pipeline best practices (I can't remember the reference now, but I could look it up if this helps). Also, starting a new (CPS-based) `sh`/`bat` durable task step to invoke `curl` is more expensive than the simple Groovy code I showed above (which could be optimized even further with `@NonCPS`). So the alternative is cleaner and more efficient.

Tested in https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/66 with and without the `skip-artifact-caching-proxy` label.

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -116,7 +116,7 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure 
     final String skipACPLabel = 'skip-artifact-caching-proxy'
     // Note: the pullRequest object is provided by https://github.com/jenkinsci/pipeline-github-plugin
     if (pullRequest.labels.contains(skipACPLabel)) {
-      echo "INFO: the label 'skip-artifact-caching-proxy' has been applied to the pull request, will use repo.jenkins-ci.org"
+      echo "INFO: the label '$skipACPLabel' has been applied to the pull request, will use repo.jenkins-ci.org"
       useArtifactCachingProxy = false
     }
   }
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -112,8 +112,7 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure 
   }
 
   // If the build concerns a pull request, check if there is "skip-artifact-caching-proxy" label applied in case the user doesn't want ACP
-  if ((useArtifactCachingProxy) && (env.CHANGE_URL)) {
-    boolean prLabelsContainSkipACP = false
+  if (useArtifactCachingProxy && env.CHANGE_URL && pullRequest != null) {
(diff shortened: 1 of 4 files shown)


vars/infra.groovy
```diff
@@ -121,23 +113,12 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure
 
   // If the build concerns a pull request, check if there is "skip-artifact-caching-proxy" label applied in case the user doesn't want ACP
   if ((useArtifactCachingProxy) && (env.CHANGE_URL)) {
```
jglick, line 115: perhaps

```suggestion
  if (useArtifactCachingProxy && env.CHANGE_URL && pullRequest != null) {
```

vars/infra.groovy
```diff
@@ -121,23 +113,12 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure
...
-      if (isUnix()) {
-        prLabelsContainSkipACP = sh(script: 'curl --silent -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $GH_TOKEN" ' + pullrequestLabelsApiURL + ' | grep --ignore-case "skip-artifact-caching-proxy"', returnStatus: true) == 0
-      } else {
-        prLabelsContainSkipACP = bat(script: 'curl --silent -H "Accept: application/vnd.github+json" -H "Authorization: Bearer %GH_TOKEN%" ' + pullrequestLabelsApiURL + ' | findstr /i "skip-artifact-caching-proxy"', returnStatus: true) == 0
-      }
-      if (prLabelsContainSkipACP) {
-        echo "INFO: the label 'skip-artifact-caching-proxy' has been applied to the pull request, will use repo.jenkins-ci.org"
-        useArtifactCachingProxy = false
-      }
+    boolean prLabelsContainSkipACP = false
```
jglick, line 116: Forgotten?

```suggestion
```

jglick (dismissed):

vars/infra.groovy
lemeurherve: Good catch, thanks!

basil (approved):
Very nice!

Issue #514, opened 2022-11-04
Default `buildPlugin()` and `buildPluginWithGradle()` to Java 11


NotMyFault:
Both calls still default to Java 8, when the minimum Java version to compile plugins against modern baselines require Java 11.

It would be nice to have Java 11 available as default, given that's the lowest supported Java version by now.

lemeurherve:
I though about that but as @dduportal noted doing so would break all plugins compatible with Java 8 only (AFAIU).

NotMyFault:
> I though about that but as @dduportal noted doing so would break all plugins compatible with Java 8 only (AFAIU).

Those would just need to specify JDK 8 then, no?

lemeurherve:
> > I though about that but as @dduportal noted doing so would break all plugins compatible with Java 8 only (AFAIU).
> 
> Those would just need to specify JDK 8 then, no?

Yes, that all they need, but until they've done it (if ever), their plugin build will be broken.
We can't break an undetermined number of plugins like that (un)fortunately.

dduportal:
> > > I though about that but as @dduportal noted doing so would break all plugins compatible with Java 8 only (AFAIU).
> > 
> > 
> > Those would just need to specify JDK 8 then, no?
> 
> Yes, that's all they need, but until they've done it (if ever), their plugin build will be broken. We can't break an undetermined number of plugins like that (un)fortunately.

+1 with what @lemeurherve said. However this task could be associated to a communication to all plugins developers + eventually a batch of PRs to change on plugins and maintainers could then choose to merge or not WDYT?

@NotMyFault worth an email on the dev mailing list to boostrap the subject?

NotMyFault:
> eventually a batch of PRs to change on plugins and maintainers could then choose to merge or not

Yup, I remember we did similar stuff in the past to accomplish infrastructure changes, like switching to the reusable CD workflow.
Although, I'd imagine that not everyone who calls `buildPlugin()` only is stuck on Java 8 (thanks to major upgrade campaigns like [JENKINS-52012](https://issues.jenkins.io/browse/JENKINS-52012), [JENKINS-51805](https://issues.jenkins.io/browse/JENKINS-51805) and [JENKINS-52284](https://issues.jenkins.io/browse/JENKINS-52284)), the vast majority amount of people possible relies on the one-liner out of convenience, rather having a test matrix to run a subset of tasks on different JDKs.

> worth an email on the dev mailing list to boostrap the subject?

[...]

PR #640, opened 2023-04-07
update default JDK version

jetersen:
JDK 11 have been required as the default for newer plugins at what time is it good thing to switch this here in the buildPlugin?

It's been 4 almost 5 months since blog post about JDK 11

https://www.jenkins.io/blog/2022/12/14/require-java-11/

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -349,7 +349,7 @@ List<Map<String, String>> getConfigurations(Map params) {
   }
 
   def platforms = params.containsKey('platforms') ? params.platforms : ['linux', 'windows']
-  def jdkVersions = params.containsKey('jdkVersions') ? params.jdkVersions : ['8']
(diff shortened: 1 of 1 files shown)


lemeurherve:
See the comments on https://github.com/jenkins-infra/pipeline-library/issues/514

Doing that would require batch updates of every plugin's Jenkinsfile first. (planned)

MarkEWaite:
As far as I can tell as a first approximation, that change will break 600 of the 1150+ plugins that have a Jenkinsfile.  I think it is much better to submit the pull requests to the individual plugins proposing to make their use of Java 8 explicit.

As a different question, is there enough benefit to the Jenkins project to make this change ever?  I agree that plugins should be moving to require Jenkins 2.361.4 or newer and should require Java 11, but that is a long process that will take years.

jglick (changes requested):
Discussed in #514.

MarkEWaite:
I'm closing this pull request because we'd rather retain compatibility for existing users and place the changes into the callers.  It can be reopened if we find the need in the future.

PR #645, opened 2023-04-11
feat(artifact-caching-proxy): set `MAVEN_ARGS` env var

lemeurherve:
As suggested by @jglick in https://github.com/jenkinsci/bom/pull/1955#discussion_r1163088843 and as this env var isn't used in any @jenkinsci or @jenkins-infra repositories and is [supported since Maven 3.9.0](https://maven.apache.org/configure.html#maven_args-environment-variable), we can set it to make the settings file be used automatically.

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -141,10 +141,9 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure 
   if (useArtifactCachingProxy) {
     echo "INFO: using artifact caching proxy from '${requestedProxyProvider}' provider."
     configFileProvider(
-        [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", targetLocation: '/tmp/settings.xml')]) {
+        [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
           withEnv([
-            'MAVEN_SETTINGS=/tmp/settings.xml',
-            'MAVEN_ARGS=-s /tmp/settings.xml',
+            "MAVEN_ARGS=-s $env.MAVEN_SETTINGS",
             "ARTIFACT_CACHING_PROXY_ORIGIN=https://repo.${requestedProxyProvider}.jenkins.io"
           ]) {
             body()
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -165,7 +165,7 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure 
  * @see withArtifactCachingProxy
  */
 Object runMaven(List<String> options, String jdk = '8', List<String> extraEnv = null, Boolean addToolEnv = true, Boolean useArtifactCachingProxy = true) {
-  List<String> mvnOptions = ['--batch-mode', '--show-version', '--errors']
+  List<String> mvnOptions = ['--batch-mode', '--show-version', '--errors', '--no-transfer-progress']
   withArtifactCachingProxy(useArtifactCachingProxy) {
     mvnOptions.addAll(options)
     mvnOptions.unique()
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -142,10 +142,7 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure 
     echo "INFO: using artifact caching proxy from '${requestedProxyProvider}' provider."
     configFileProvider(
         [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
-          withEnv([
-            "MAVEN_ARGS=-s $env.MAVEN_SETTINGS",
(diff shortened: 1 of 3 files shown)


vars/infra.groovy
```diff
@@ -159,14 +165,8 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure
  * @see withArtifactCachingProxy
  */
 Object runMaven(List<String> options, String jdk = '8', List<String> extraEnv = null, Boolean addToolEnv = true, Boolean useArtifactCachingProxy = true) {
-  List<String> mvnOptions = ['--batch-mode', '--show-version', '--errors', '--no-transfer-progress']
+  List<String> mvnOptions = ['--batch-mode', '--show-version', '--errors']
```
lemeurherve, line 168:
```suggestion
  List<String> mvnOptions = ['--batch-mode', '--show-version', '--errors', '--no-transfer-progress']
```

vars/infra.groovy
```diff
@@ -141,8 +141,14 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure
   if (useArtifactCachingProxy) {
     echo "INFO: using artifact caching proxy from '${requestedProxyProvider}' provider."
     configFileProvider(
-        [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
-          body()
+        [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", targetLocation: '/tmp/settings.xml')]) {
```
jglick, line 144: Why the switch to a hard-coded location? Fine for a container agent but could break concurrent activity on some agents.

vars/infra.groovy
lemeurherve: Confirmed the correct download from the proxy in https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/lemeurherve-patch-2/2/console

vars/infra.groovy
lemeurherve: I'm not really happy with that too but I had issues with the variable, I didn't managed to retrieve its value while running the unit tests.
Will try again.

lemeurherve: > could break concurrent activity on some agents.

Maybe adding a random value and/or an agent identifier would avoid breaking concurrent activity or is it simply a bad idea?
I need to check that more thoroughly tomorrow.

vars/infra.groovy
```diff
@@ -141,8 +141,14 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure
...
+          withEnv([
+            'MAVEN_SETTINGS=/tmp/settings.xml',
+            'MAVEN_ARGS=-s /tmp/settings.xml',
```
lemeurherve, line 147: re https://github.com/jenkins-infra/pipeline-library/pull/645#discussion_r1163250167
My initial version was like
```suggestion
        [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
          withEnv([
            "MAVEN_ARGS=-s $MAVEN_SETTINGS",
```
But when running the tests I saw only `MAVEN_ARGS=-s null` in the outputs. (Probably my mistake though)
I've also tested
```suggestion
        [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
          withEnv([
            "MAVEN_ARGS=-s $env.MAVEN_SETTINGS",
```
same `null` output.

vars/infra.groovy
```diff
@@ -142,7 +142,12 @@ Object withArtifactCachingProxy(boolean useArtifactCachingProxy = true, Closure
     echo "INFO: using artifact caching proxy from '${requestedProxyProvider}' provider."
     configFileProvider(
         [configFile(fileId: "artifact-caching-proxy-${requestedProxyProvider}", variable: 'MAVEN_SETTINGS')]) {
-          body()
+          withEnv([
+            "MAVEN_ARGS=-s $env.MAVEN_SETTINGS",
+            "ARTIFACT_CACHING_PROXY_ORIGIN=https://repo.${requestedProxyProvider}.jenkins.io"
```
jglick, line 147: Just for fun, or do you expect something to use this?

jglick (approved):

vars/infra.groovy
lemeurherve: I've added this for https://github.com/jenkins-infra/javadoc/blob/master/Jenkinsfile#L35
(and other similar potential changes)

Not mandatory at all, I can keep my current Javadoc version.

lemeurherve:
Confirmed working as intended: https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/lemeurherve-patch-2/3/console

vars/infra.groovy
jglick: I see, can be used for tools not running Maven like https://github.com/jenkins-infra/javadoc/blob/595c5188e24434669bd522bfb1aefaa23cdab2be/scripts/generate-javadoc.sh#L21 :+1:

jglick (approved):

PR #646, opened 2023-04-11
chore: disable concurrent builds

lemeurherve:


dduportal (commented):
Not sure it is worth it. What is the reason?

MarkEWaite (approved):

basil:
This is in line with established precedent in the project. GitHub Actions takes the opposite stance, testing each commit by default. I fail to comprehend the long-term economic viability of such an approach.

hervelemeur:
> > 7 concurrent builds started on my previous pipeline library pull request, wasting resources for nothing.
> 
> Make sense, but word of advise: if doing a lot of commit wastes resources, it means that commiting a bit less would help also ;)

Sure, but you also sometime need these multiple commits, for example to take in account review comments, or to test in situ that your changes work as intended by temporarily modifying the code then reverting it. (In my case the `--no-tranfer-progress` switch, 4 of these 7 commits)

PR #653, opened 2023-04-20
feat: activate `launchable` in `buildPlugin` on Windows, remove `launchable.install` from `buildPlugin` and add corresponding deprecation notice

lemeurherve:
Follow-up of https://github.com/jenkins-infra/packer-images/pull/605 & https://github.com/jenkins-infra/docker-inbound-agents/pull/71

As Launchable is now available on every agent, this PR:
- activate `launchable` on Windows builds
- removes `launchable.install` from `buildPlugin`
- add a deprecation notice when `launchable.install` is called

Tested in https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/82

Ref: https://github.com/jenkins-infra/helpdesk/issues/3484

### DIFF
--- a/vars/launchable.groovy
+++ b/vars/launchable.groovy
@@ -44,7 +44,7 @@ def call(String args) {
       currentBuild.result = 'SUCCESS'
       echo 'Failed to run Launchable; continuing build.'
     } finally {
-      def errorsCount = bat(script: 'type ' + launchableStderrFile + ' | find /c /v ""', returnStdout: true)
+      def errorsCount = pwsh(script: "(Get-Content $launchableStderrFile).Length", returnStdout: true)
       echo "laucnhable errors count: $errorsCount"
       if (errorsCount > 2) {
         echo 'Launchable errors log (please ignore the two first lines); continuing build.'
--- a/vars/launchable.groovy
+++ b/vars/launchable.groovy
@@ -44,11 +44,11 @@ def call(String args) {
       currentBuild.result = 'SUCCESS'
       echo 'Failed to run Launchable; continuing build.'
     } finally {
-      def errorsCount = Integer.parseInt(pwsh(script: "(Get-Content $launchableStderrFile | Select-Object -Skip 2) | Set-Content $launchableStderrFile; (Get-Content $launchableStderrFile).Length", returnStdout: true))
+      def errorsCount = Integer.parseInt(pwsh(script: "(Get-Content $launchableStderrFile).Length", returnStdout: true).trim())
       echo "launchable errors count: $errorsCount"
       if (errorsCount > 2) {
-        echo 'Launchable errors log (please ignore the two first lines); continuing build.'
-        bat 'type ' + launchableStderrFile
+        echo 'Launchable errors log below; continuing build.'
+        pwsh(script: "(Get-Content launchable.stderr | Select-Object -Skip 2)", returnStdout: false)
       }
     }
   }
--- a/vars/launchable.groovy
+++ b/vars/launchable.groovy
@@ -44,8 +44,8 @@ def call(String args) {
       currentBuild.result = 'SUCCESS'
       echo 'Failed to run Launchable; continuing build.'
     } finally {
-      def errorsCount = pwsh(script: "(Get-Content $launchableStderrFile).Length", returnStdout: true)
-      echo "laucnhable errors count: $errorsCount"
+      def errorsCount = Integer.parseInt(pwsh(script: "(Get-Content $launchableStderrFile | Select-Object -Skip 2) | Set-Content $launchableStderrFile; (Get-Content $launchableStderrFile).Length", returnStdout: true))
+      echo "launchable errors count: $errorsCount"
       if (errorsCount > 2) {
         echo 'Launchable errors log (please ignore the two first lines); continuing build.'
         bat 'type ' + launchableStderrFile
--- a/vars/launchable.groovy
+++ b/vars/launchable.groovy
@@ -45,10 +45,9 @@ def call(String args) {
       echo 'Failed to run Launchable; continuing build.'
     } finally {
       def errorsCount = Integer.parseInt(pwsh(script: "(Get-Content $launchableStderrFile).Length", returnStdout: true).trim())
-      echo "launchable errors count: $errorsCount"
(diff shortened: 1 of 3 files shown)


basil:
Why would we want to do this? The idea is that the caller should only call `launchable.install()` once per node. Plus this was intended to be temporary code anyway: the permanent solution originally envisioned was to bake the Launchable PyPI package into each relevant system at (container or VM) image creation time so that it could be invoked from the `PATH` by running `launchable` without Pipeline scripts having to worry about whether or not to install it.

lemeurherve:
@basil If I'm reading your help desk issue and your comment here correctly, I think we're on the same page.

I've only kept the `install` part here for now as Launchable isn't installed on the ACI Windows images yet (it's installed on the other images since https://github.com/jenkins-infra/packer-images/pull/605), hence the draft status of this PR.

I intend to remove `launchable.install` from `buildPlugin` when it's done.

basil:
> Launchable isn't installed on the ACI Windows images yet

Note that I will have an immediate use case for when it is available on ACI Windows images, as I intend to use Launchable to decrease the 6-hour long Windows core builds to a maximum of 90 minutes.

lemeurherve:
It took me much more time than anticipated but I finally got launchable and python installed on Windows ACI agent images.

I tried several ways for a working python install on the Nanoserver image, finally used a multi-stage image build.

On this Nanoserver image I didn't managed to get `launchable` working when called directly, even if the plugin is correctly detected as installed: https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/PR-82/24/console

I had to call `launchable` via python (`python -m launchable --version` for example), and still got 2 non blocking errors in the logs: https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/PR-82/26/console
```
python.exe -m launchable --version 
 pythoncom error: CoInitializeEx failed (0x80004001)
 pythoncom error: OLE initialization failed! (0x80004005)
 launchable-cli, version 1.63.0
```

I've then tried to copy the ole32.dll from a Windows Server image to the Nanoserver one in the Dockerfile, as it seems these two functions come from this DLL ([list of Nano server API](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/mt588480(v=vs.85)) > [OleInitialize](https://learn.microsoft.com/en-gb/windows/win32/api/ole2/nf-ole2-oleinitialize?redirectedfrom=MSDN) & [CoInitializeEx](https://learn.microsoft.com/en-gb/windows/win32/api/combaseapi/nf-combaseapi-coinitializeex?redirectedfrom=MSDN)), but it didn't worked and resulted in a blocking error:
```
Step 11/17 : COPY --from=core C:/windows/system32/ole32.dll C:/windows/system32/ole32.dll
 ---> Using cache
 ---> a91209838916
Step 12/17 : RUN "C:/tools/python/python.exe" -m launchable --version
 ---> Running in b07d7845ac7b
Failed to load the dll from [C:\Program Files\PowerShell\coreclr.dll], HRESULT: 0x8007007F
The command 'pwsh.exe -Command $ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue'; "C:/tools/python/python.exe" -m launchable --version' returned a non-zero code: 3221226505
```

In this PR I'm calling launchable via `bat 'python -m launchable ' + args + ' 2>NUL'`, I've tested it in https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/PR-82/30/console, it seems to work as intended.

lemeurherve:
(My comment wasn't finished, continuing here)

If the method used in this PR with the Nano server isn't acceptable, we could switch our docker inbound agent images from Nano Server to Windows Server, at the price of a bigger image, taking more time to be pulled for these ACI agents running on VMs.

We intend later to stop using these ACI VM Windows agents with these docker inbound agent images, but instead building an image with Packer, which imply the use of Windows Server instead of Nano Server.
This packer image could then be run from a Kubernetes agent like the Linux ones, but this work hasn't started.

@basil WDYT about this current state?
Does it seem OK for you as it is or should we move right now to a Windows Server image at the price of longer (Windows) builds?

basil:
This is fine as a short-term workaround. I reported the issue to our Launchable product manager in a corporate Slack channel. If you can come up with a minimal reproducible test case, it might be worth filing an issue about Nanoserver support at https://github.com/launchableinc/cli/issues

vars/launchable.groovy
```diff
@@ -1,23 +1,43 @@
...
-      sh 'launchable/bin/launchable ' + args
+      sh 'launchable ' + args
     }
   } else {
-    error 'The Launchable CLI is not yet implemented for Windows'
+    catchError(buildResult: 'SUCCESS', catchInterruptions: false, message: 'Failed to run Launchable; continuing build.') {
+      // Avoid logging the errors due to Nano server used in docker-inbound-agent
+      // TODO remove ' 2>NUL' when switching to Windows Server images instead of Nano Server images
+      bat 'python -m launchable ' + args + ' 2>NUL'
+    }
```
dduportal, line 41:
```suggestion
    def launchableStderrFile = 'launchable.stderr
    try {
      // Avoid logging the errors due to Nano server used in docker-inbound-agent
      // TODO remove ' 2>NUL' when switching to Windows Server images instead of Nano Server images
      bat 'python -m launchable ' + args + ' 2>' + launchableStderrFile
    catch (err) {
        echo "Caught: ${err}"
        currentBuild.result = 'SUCCESS'
    finally {
      bat 'type ' + launchableStderrFile
    }
```

dduportal: https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/

dduportal: As seen in 1:1: the goal would be to avoid masking error messages from launchable when it fails (but when it succeeeds , e.g. nominal case, then we don't want to see stderr)

dduportal (approved):
LGTM : code looks good, worth testing the latest version (if not already done) before merging

lemeurherve:
> LGTM : code looks good, worth testing the latest version (if not already done) before merging

Testing done with the last code in https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/82

PR #655, opened 2023-04-21
fix(terraform): add toleration to use infracipool

lemeurherve:
To avoid cutting off the branch we're sitting on when modifying privatek8s node pools (ex: https://github.com/jenkins-infra/azure/pull/329 and its infra.ci.jenkins.io job which ran on the "linuxpool" node pool impacted by this change), this PR sets a toleration so terraform jobs will run on infracipool only.


dduportal (approved):
LGTM

PR #522, opened 2022-11-15
Feat(agent label): use a specific label for windows if not a container

smerle33:
in order to avoid the `windows` label as per https://github.com/jenkins-infra/helpdesk/issues/3223#issuecomment-1305703973 we set a specific label.

PR #658, opened 2023-04-26
fix(buildPlugin) ensure the custom label is passed to node when addin…

dduportal:
…g custom platform

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -58,8 +58,8 @@ def call(Map params = [:]) {
           label = 'vm && linux'
           break
         default:
-          echo "WARNING: Unknown platform '${platform}'. Agent label set to fallback value 'linux'"
-          label = 'linux'
+          echo "WARNING: Unknown Virtual Machine platform '${platform}'. Set useContainerAgent to 'true' unless you want to be in uncharted territory."
+          label = platform
       }
     }
 
--- a/test/groovy/BuildPluginStepTests.groovy
+++ b/test/groovy/BuildPluginStepTests.groovy
@@ -125,14 +125,42 @@ class BuildPluginStepTests extends BaseTest {
     // when running without any parameters
     script.call([:])
     printCallStack()
-    // then it runs in a linux node
-    assertTrue(assertMethodCallContainsPattern('node', 'linux'))
-    // then it runs in a windows node
-    assertTrue(assertMethodCallContainsPattern('node', 'windows'))
+    // then it runs a stage in a linux VM by default
+    assertTrue(assertMethodCallContainsPattern('node', 'vm && linux'))
+    // then it runs a stage in a Windows VM by default
+    assertTrue(assertMethodCallContainsPattern('node', 'docker-windows'))
     // then it runs the junit step by default
     assertTrue(assertMethodCall('junit'))
     // then it runs the junit step with the maven test format
     assertTrue(assertMethodCallContainsPattern('junit', '**/target/surefire-reports/**/*.xml,**/target/failsafe-reports/**/*.xml'))
+    assertJobStatusSuccess()
+  }
+
+  @Test
+  void test_buildPlugin_with_container_agents() throws Exception {
+    def script = loadScript(scriptName)
+    // when running with useContainerAgent set to true
+    script.call([useContainerAgent: true])
+    printCallStack()
+    // then it runs a stage in a linux container by default
+    assertTrue(assertMethodCallContainsPattern('node', 'maven'))
+    // then it runs a stage in a Windows container by default
+    assertTrue(assertMethodCallContainsPattern('node', 'maven-windows'))
+    assertJobStatusSuccess()
+  }
+
+  @Test
+  void test_buildPlugin_with_customplatforms() throws Exception {
+    def script = loadScript(scriptName)
+    // when running with useContainerAgent set to true
+    script.call(platforms: ['openbsd', 'maven-windows-experimental'])
+    printCallStack()
+    // then it runs a stage in an openbsd node with a warning message
+    assertTrue(assertMethodCallContainsPattern('node', 'openbsd'))
+    assertTrue(assertMethodCallContainsPattern('echo', 'WARNING: Unknown Virtual Machine platform \'openbsd\''))
+    // then it runs a stage in a maven-windows-experimental node with a warning message
+    assertTrue(assertMethodCallContainsPattern('node', 'maven-windows-experimental'))
+    assertTrue(assertMethodCallContainsPattern('echo', 'WARNING: Unknown Virtual Machine platform \'maven-windows-experimental\''))
     assertJobStatusSuccess()
   }
 


vars/buildPlugin.groovy
```diff
@@ -58,8 +58,8 @@ def call(Map params = [:]) {
           label = 'vm && linux'
           break
         default:
-          echo "WARNING: Unknown platform '${platform}'. Agent label set to fallback value 'linux'"
-          label = 'linux'
+          echo "WARNING: Unknown Virtual Machine platform '${platform}'. Set useContainerAgent to 'true' unless you want to be in uncharted territory."
```
jtnord, line 61: this still seems at odds with the [documentation](https://github.com/jenkins-infra/pipeline-library/blob/master/README.adoc?plain=1#L41-L42)  (ie this is supposed to support labels - so using a label (even a label expression if I read this doc correct) should be ok here and not produce a warning?)

jtnord (approved):
Looks correct - not sure about the warning given the documentation around this case.

vars/buildPlugin.groovy
dduportal: yep, updating the documentation is the next step 👍  thanks for pointing out, I would have forgotten to be honest.

We are trying to update the label models because the current one is not usable in 2023. Some examples:

- Should you expect `docker` to be available when specifying `maven-windows` (no obvious answer of course :D). Which JDK should you have?
- When you say `linux`, do you mean `x86` or `arm64`? With or without Docker Engine?

=> our current strategy is to move to explicit tags, #522 was driven by this idea. Becasue most of the labels are specified by the pipeline library, only edge cases uses custom labels.


=> a takeaway, I strongly suggest to change 

```
buildPlugin(platforms: ['linux','windows'])
``` 

by 

```
buildPlugin(
  useContainerAgent: true,
  configurations: [
    [platform: 'linux', jdk: 11],
    [platform: 'windows', jdk: 11],
])
```

jglick (approved):
I am still unclear on the takeaway for plugins—what _should_ they specify to build on

* A Linux VM with Docker available.
* A Windows container (without Docker).

?

dduportal:
> I am still unclear on the takeaway for plugins—what _should_ they specify to build on
> 
>     * A Linux VM with Docker available.
> 
>     * A Windows container (without Docker).
> 
> 
> ?

Ideally,  container agents (linux or windows) should be the default **unless** Docker engine is required.
Docker CLI should be expected on all kinds to be found though to allow "client side" operation (wether with buildx or compose) without involving containers (for instance: parsing a `bake.hcl` or `compose.yml` file)


The function `buildPlugin` and the way to specify platform is confusing to me though, as there seems to be tons of options to bypass the default: might need thinking this a bit more thoroughly. But  short term, fixing the documentation (by infra team) would help a lot

jglick:
> the default **unless** Docker engine is required

Right, my question is what `buildPlugin` format is recommended for a plugin which _does_ require Docker (server supporting Linux containers, as well as CLI) for some tests, and also wishes to run non-Docker-based tests on Windows.

dduportal:
> > the default **unless** Docker engine is required
> 
> Right, my question is what `buildPlugin` format is recommended for a plugin which _does_ require Docker (server supporting Linux containers, as well as CLI) for some tests, and also wishes to run non-Docker-based tests on Windows.

I thought setting `useContainerAgent` to `false` would be it, but I realize that this attribute is scope to the whole set of platforms/configuration 🤦  

Good point!

Short term (modulo the JDK version you might want to change):

```
buildPlugin(
  useContainerAgent: false,
  configurations: [
    [platform: 'linux', jdk: 11],
    [platform: 'windows', jdk: 11],
])
```

would ensure that VM are used:
- For Linux, to ensure Docker is present
- For Windows, not needed but at least should work for now

=> got plenty of feedback for improvement thanks!

jglick:
But the above will build on a Windows VM, right? So if the Windows branch can run fine in a container, then you can still specify `maven-windows` as in https://github.com/jenkinsci/docker-workflow-plugin/pull/222#discussion_r471640290?

dduportal:
> But the above will build on a Windows VM, right? So if the Windows branch can run fine in a container, then you can still specify `maven-windows` as in [jenkinsci/docker-workflow-plugin#222 (comment)](https://github.com/jenkinsci/docker-workflow-plugin/pull/222#discussion_r471640290)?

Yes, absolutely (haven't thought about it). WDYT about an "explicit" docker request, something like

```
buildPlugin(platforms: [
    'linux && docker',
    'maven-windows && !docker'
])
```

vars/buildPlugin.groovy
jtnord: > When you say linux, do you mean x86 or arm64? With or without Docker Engine?

I mean I do not care :)  (currently as it was restricted - linux is ambiguos and the useage of it in this library has not helped plugin authors)..  

But in reality it may depend on if this is actually allowing a label, or a label expression.  If it would be a label expression then a developer could say `linux && docker && x64` 
Would that pick up `highmen` - as is very likely - but we should be able to make `highmen` (and other expensive options) only available if a specific label is actually given.  This may require a plugin (not sure off hand) but would seem like a useful thing to have to the wider community.

jtnord:
> Yes, absolutely (haven't thought about it). WDYT about an "explicit" docker request, something like

as for `linux` being a mostly bad label you have the similarly bad `docker`.
what does this mean?  the `docker` cli is installed?  the docker cli is installed and points to a working docker install (local / remote?)  the docker cli is installed and supports specific platforms (linux/arm64 windows....)?

what if I do want to run the docker (linux containers) tests on windows (and such support was available)?

```
buildPlugin(platforms: [
    'linux && docker',
    'maven-windows && docker && docker_linux_container'
])
```

and for things that want windows containers

```
buildPlugin(platforms: [
    'linux && docker',
    'maven-windows && docker && docker_windows_container'
])
```

see how `docker` is redundant here?   linux with docker should become `linux && docker_linux_containers` 

and this does not discuss [`testcontainers.cloud`](https://www.testcontainers.cloud/) which needs a docker cli but not necessarily a docker daemon (I have not played - may be incorrect)

vars/buildPlugin.groovy
dduportal: Absolutely, our analysis matches yours! That would be the road to a (breaking) change in the provided labels on ci.jenkins.io.

The behavior in label matching you describe is the reason why, in 2021, for 2 months, we used ARM VM agents for a lot of builds without even knowing it, because `linux` was resolved by both ARM and x86 machines. 
=> As soon as Jenkins find an agent template matching the labels, it uses it.


We are thinking about a policy with looooong tags such as `<os>-<cpu>-<qualifier>` but still note sure where to put the cursor with or without too much details as we have different dimensions but it might not be needed (or event wanted) to provide all of them:

- OS (Linux Ubuntu, Windows Core Server)
- OS version (22.04 for Ubuntu, 2019/2022 for Windows Core Server)
- size (such as `highmem`)
- Compute type: VM, container
- Tools (`docker`, `maven`, `jdk17`)
- Feature: `jdk-next` (for edge non stable JDK), `experiemental`, etc.
- Cloud/Location (`aws`, `azure`, etc.)
- Infrastructure (such as cluster name or region name: `doks`, etc.)


We though about the platform labeler plugin, but haven't look or tried closely. Any idea or tip is welcome.

PR #664, opened 2023-05-18
Remove use of Launchable shared library functionality

basil:
Now that Launchable is installed on our container agents, there is no longer a need to use the Launchable shared library, which is slower and flakier.

Note that Jenkins core is still consuming the shared library. I plan to remove consumption from there. Once that consumption is removed (and backported!) I plan to delete `test/groovy/LaunchableStepTests.groovy`, `vars/launchable.groovy`, and `vars/launchable.txt` from this repository.

### Testing done

`buildPlugin`: https://ci.jenkins.io/job/Plugins/job/text-finder-plugin/job/PR-172/18/ (Linux) and https://ci.jenkins.io/job/Plugins/job/text-finder-plugin/job/PR-172/19/ (Windows)

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -279,10 +279,14 @@ def call(Map params = [:]) {
                    * plugins' PR builds could not be consumed by anything else anyway, and all
                    * plugins currently in the BOM are incrementalified.
                    */
-                  if (incrementals && currentBuild.currentResult == 'SUCCESS') {
+                  if (incrementals && currentBuild.currentResult == 'SUCCESS' && useContainerAgent) {
                     withCredentials([string(credentialsId: 'launchable-jenkins-bom', variable: 'LAUNCHABLE_TOKEN')]) {
-                      launchable('verify')
-                      launchable('record commit')
+                      if (isUnix()) {
+                        sh 'launchable verify && launchable record commit'
+                      } else {
+                        // TODO launchable.exe still not working for some reason
+                        bat 'python -m launchable verify && python -m launchable record commit'
+                      }
                     }
                   }
                 } else {
--- a/test/groovy/BuildPluginStepTests.groovy
+++ b/test/groovy/BuildPluginStepTests.groovy
@@ -1,6 +1,5 @@
 import mock.CurrentBuild
 import mock.Infra
-import mock.Launchable
 import org.junit.Before
 import org.junit.Test
 import static org.junit.Assert.assertEquals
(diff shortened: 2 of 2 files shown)


vars/buildPlugin.groovy
```diff
@@ -279,10 +279,14 @@ def call(Map params = [:]) {
                    * plugins' PR builds could not be consumed by anything else anyway, and all
                    * plugins currently in the BOM are incrementalified.
                    */
-                  if (incrementals && currentBuild.currentResult == 'SUCCESS') {
+                  if (incrementals && currentBuild.currentResult == 'SUCCESS' && useContainerAgent) {
```
basil, line 282: This might be unnecessary, but I only tested in containers so I thought I should rather be safe than sorry. If the operators assure me that this is not needed, I can remove it, but I do not plan to do more thorough testing because commit collection is only needed on a best-effort basis anyway, so from a correctness point of view it is not a big deal if we are missing some commits. Note that I did test this in containers in both Linux and Windows.

vars/buildPlugin.groovy
```diff
@@ -279,10 +279,14 @@ def call(Map params = [:]) {
...
                     withCredentials([string(credentialsId: 'launchable-jenkins-bom', variable: 'LAUNCHABLE_TOKEN')]) {
-                      launchable('verify')
-                      launchable('record commit')
+                      if (isUnix()) {
+                        sh 'launchable verify && launchable record commit'
+                      } else {
+                        // TODO launchable.exe still not working for some reason
```
basil, line 287: https://github.com/jenkins-infra/helpdesk/issues/3484#issuecomment-1552280288

lemeurherve: Should be addressed by https://github.com/jenkins-infra/docker-inbound-agents/pull/83

lemeurherve (approved):

vars/buildPlugin.groovy
basil: https://ci.jenkins.io/job/Plugins/job/text-finder-plugin/job/PR-172/24/console is still failing, but I am not sure whether that is because https://github.com/jenkins-infra/docker-inbound-agents/pull/83 failed to fix the problem or simply has not been deployed to production yet.

lemeurherve: It wasn't deployed yet, done in https://github.com/jenkins-infra/jenkins-infra/pull/2845

Confirmed working in https://ci.jenkins.io/job/Plugins/job/text-finder-plugin/job/PR-172/25/console

PR #673, opened 2023-05-23
feat(updatecli): add `debug` parameter

lemeurherve:
This PR add the `debug` parameter (set to `false` by default) so we can activate the debug output (by adding `--debug` to the CLI call)

Tested in https://github.com/jenkins-infra/docker-inbound-agents/pull/87 (commit https://github.com/jenkins-infra/docker-inbound-agents/pull/87/commits/dd3eb0051715b70f2f0e726b8ccadcb5d113a0f4)


dduportal:
I have mixed feelings as the debug is potentially leaking credentials in its output, so having it as a default sounds dangerous

PR #691, opened 2023-06-27
Remove support for Gradle from `buildPlugin`

NotMyFault:
Support for Gradle in buildPlugin has been deprecated for more than 4 years. I'd propose to remove its support from buildPlugin, in favor of buildPluginWithGradle.

I've proposed PRs updating the existing repositories from buildPlugin to buildPluginWithGradle, builds are successful this time ;)

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -68,7 +68,6 @@ def call(Map params = [:]) {
         node(label) {
           try {
             timeout(timeoutValue) {
-              boolean isMaven
               // Archive artifacts once with pom declared baseline
               boolean doArchiveArtifacts = !jenkinsVersion && !archivedArtifacts
               if (doArchiveArtifacts) {
@@ -78,7 +77,6 @@ def call(Map params = [:]) {
 
               stage("Checkout (${stageIdentifier})") {
                 infra.checkoutSCM(repo)
-                isMaven = !fileExists('gradlew')
                 incrementals = fileExists('.mvn/extensions.xml') &&
                     readFile('.mvn/extensions.xml').contains('git-changelist-maven-extension')
                 final String gitUnavailableMessage = '[buildPlugin] Git CLI may not be available'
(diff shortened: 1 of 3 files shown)


vars/buildPlugin.groovy
```diff
@@ -303,11 +277,7 @@ def call(Map params = [:]) {
                     publishingIncrementals = true
                   } else {
                     String artifacts
-                    if (isMaven) {
-                      artifacts = '**/target/*.hpi,**/target/*.jpi,**/target/*.jar'
-                    } else {
-                      artifacts = '**/build/libs/*.hpi,**/build/libs/*.jpi'
-                    }
+                    artifacts = '**/target/*.hpi,**/target/*.jpi,**/target/*.jar'
                     archiveArtifacts artifacts: artifacts, fingerprint: true
```
basil, line 281: The only reason to pull this out into a variable would have been to support the conditional structure that is now being deleted. Now that the conditional logic is being deleted, leaving behind a variable that is only ever used one time does not make sense and only serves to make the code less readable by adding an unnecessary layer of indirection. In general, any variable that is used only one time should be inlined unless there is a specific reason otherwise.

```suggestion
                    archiveArtifacts artifacts: '**/target/*.hpi,**/target/*.jpi,**/target/*.jar', fingerprint: true
```

basil (approved):
Thanks for the PR!

vars/buildPlugin.groovy
```diff
@@ -303,12 +277,7 @@ def call(Map params = [:]) {
                     publishingIncrementals = true
                   } else {
                     String artifacts
```
jglick, line 279:
```suggestion
```

Note that you could have accepted https://github.com/jenkins-infra/pipeline-library/pull/691#discussion_r1244343602 (which included this deletion) directly from the GH web UI.

jglick (approved):

vars/buildPlugin.groovy
NotMyFault: > Note that you could have accepted https://github.com/jenkins-infra/pipeline-library/pull/691#discussion_r1244343602 (which included this deletion) directly from the GH web UI.

GH refused to push the comment, because of deleted lines, apparently :/

jglick: Oh right, GH has problems with suggestions that span edits already in the PR.

basil (approved):

vars/buildPlugin.groovy
basil: I miss Review Board.

PR #705, opened 2023-07-27
Skip Incrementals deployment for branch build of CD-enabled repo

jglick:
Fixes https://github.com/jenkins-infra/helpdesk/issues/3687. Not sure whether it is feasible to test something like this in advance. @timja can the infra test plugin be set up to pick up this PR throughout a plugin PR test, merge, and release cycle? Would still not be verifying that there is no behavioral change for a non-CD repo like `jenkinsci/jenkins`.

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -343,11 +343,13 @@ void prepareToPublishIncrementals() {
 void maybePublishIncrementals() {
   if (new InfraConfig(env).isRunningOnJenkinsInfra() && currentBuild.currentResult == 'SUCCESS') {
     if (env.CHANGE_ID == null) {
+      def skip
       catchError(message: 'Could not check whether repo has enabled CD', buildResult: 'SUCCESS', stageResult: 'UNSTABLE', catchInterruptions: false) {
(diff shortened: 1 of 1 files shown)


timja:
of course, see https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/82

vars/infra.groovy
```diff
@@ -342,6 +342,14 @@ void prepareToPublishIncrementals() {
  */
 void maybePublishIncrementals() {
   if (new InfraConfig(env).isRunningOnJenkinsInfra() && currentBuild.currentResult == 'SUCCESS') {
+    if (env.CHANGE_ID == null) {
+      catchError(message: 'Could not check whether repo has enabled CD', buildResult: 'SUCCESS', stageResult: 'UNSTABLE', catchInterruptions: false) {
```
jglick, line 346: In case `readTrusted` fails, for example because `.mvn/maven.config` does not exist (though in that case it would be impossible for Incrementals publishing to work anyway), or because it has been modified in an untrusted revision (though this should only happen when `CHANGE_ID` is set). Erring on the side of the status quo.

timja (approved):

Issue #285, opened 2022-01-17
tune the number of forks run for surefire


jtnord:
the CI nodes may (or may not have a large number of CPUs), and by default we only run one surefire fork for all tests.

This could be a bit wasteful for many plugins where multiple tests run in parallel could hugely decrease the plugins build time.
(e.g. https://github.com/jenkinsci/git-plugin/pull/1173#discussion_r785025999) 

therefore the number of forks should be tuned for all plugins to a reasonable default (we already tune the surefire JVM args to a set JVM heap so this number could reasonably be caclulated knowing the underlying hardware).

Additionally this tuning may not work for all plugins (for example warnings ng or plugins using `RealJenkinsRule` that require more memory, so allowing this number to be tweaked / overriddien would be required at the same time (or the number chosen would have to be low enough to account for these plguins which could be deemed as wastefull).

Quicker builds on the same hardware will also reduce the number of agents we require which should reduce the cloud costs for the product.

dduportal:
Do you think that this value could be defined based on the amount of vCPUs availables?

PR #707, opened 2023-07-28
Add forkCount argument to buildPlugin

MarkEWaite:
## Add forkCount argument to buildPlugin

Allow CI jobs to declare parallel execution rather than using a configuration in the pom.xml file.

Fixes https://github.com/jenkins-infra/pipeline-library/issues/285 to allow users to tune the number of forks based on their knowledge of the tests in their plugin.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -34,6 +34,7 @@ buildPlugin(
 * `repo` (default: `null`  inherit from Multibranch) - custom Git repository to check out
 * `useContainerAgent` (default: `false`) - uses a link:https://github.com/jenkins-infra/documentation/blob/main/ci.adoc#container-agents[Container agent] instead of a Virtual Machine: usually faster to start and generates less costs for the project
 ** Please note that the implementation of "containers" can be changed over time
+* `forkCount` (default: null) - Execute tests with `forkCount` Java virtual machines. Common values include `0.45C`, `1`, `2`, and `1C`.
 * `failFast` (default: `true`) - instruct Maven tests to fail fast
 * `platforms` (default: `['linux', 'windows']`) - Labels matching platforms to
   execute the steps against in parallel
--- a/README.adoc
+++ b/README.adoc
@@ -34,7 +34,7 @@ buildPlugin(
 * `repo` (default: `null`  inherit from Multibranch) - custom Git repository to check out
 * `useContainerAgent` (default: `false`) - uses a link:https://github.com/jenkins-infra/documentation/blob/main/ci.adoc#container-agents[Container agent] instead of a Virtual Machine: usually faster to start and generates less costs for the project
 ** Please note that the implementation of "containers" can be changed over time
-* `forkCount` (default: null) - Execute tests with `forkCount` Java virtual machines. If no value is specified, a single JVM is used.  Common values include `0.45C`, `1`, `2`, and `1C`. If you terminate the value with a 'C', that value will be multiplied by the number of available CPU cores. This controls the number of JVM's used for the test code, including unit tests, integration tests (like JenkinsRule tests), and additional JVM's launched by end to end tests (like agent JVM's and RealJenkinsRule tests).
+* `forkCount` (default: null) - Execute tests with `forkCount` Java virtual machines. If no value is specified, a single JVM is used.  Common values include `0.45C`, `1`, `2`, and `1C`. If you terminate the value with a 'C', that value will be multiplied by the number of available CPU cores. This controls the number of JVM's used for the test code.  Additional JVM's will be launched if the test code starts agents or uses `RealJenkinsRule`.
 * `failFast` (default: `true`) - instruct Maven tests to fail fast
 * `platforms` (default: `['linux', 'windows']`) - Labels matching platforms to
   execute the steps against in parallel
--- a/README.adoc
+++ b/README.adoc
@@ -34,7 +34,7 @@ buildPlugin(
 * `repo` (default: `null`  inherit from Multibranch) - custom Git repository to check out
 * `useContainerAgent` (default: `false`) - uses a link:https://github.com/jenkins-infra/documentation/blob/main/ci.adoc#container-agents[Container agent] instead of a Virtual Machine: usually faster to start and generates less costs for the project
 ** Please note that the implementation of "containers" can be changed over time
-* `forkCount` (default: null) - Execute tests with `forkCount` Java virtual machines. Common values include `0.45C`, `1`, `2`, and `1C`.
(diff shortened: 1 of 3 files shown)


dduportal (approved):

dduportal:
I would want a secondary review from pipeline library specialist (but it looks 👍 from pure infra side)

README.adoc
```diff
@@ -34,6 +34,7 @@ buildPlugin(
 * `repo` (default: `null`  inherit from Multibranch) - custom Git repository to check out
 * `useContainerAgent` (default: `false`) - uses a link:https://github.com/jenkins-infra/documentation/blob/main/ci.adoc#container-agents[Container agent] instead of a Virtual Machine: usually faster to start and generates less costs for the project
 ** Please note that the implementation of "containers" can be changed over time
+* `forkCount` (default: null) - Execute tests with `forkCount` Java virtual machines. Common values include `0.45C`, `1`, `2`, and `1C`.
```
lemeurherve, line 37: What's the difference between `1` and `1C`?

lemeurherve: Found an explanation here: https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html

> if you terminate the value with a 'C', that value will be multiplied with the number of available CPU cores in your system

lemeurherve, line 37: Optional addition to the comment:
```suggestion
* `forkCount` (default: null) - Execute tests with `forkCount` Java virtual machines. Common values include `0.45C`, `1`, `2`, and `1C` (if you terminate the value with a 'C', that value will be multiplied with the number of available CPU cores).
```

jtnord (approved):
🥳  thank you @MarkEWaite

README.adoc
MarkEWaite: The `1C` value forks a virtual machine per core on the machine.  If the machine has 16 cores, then 16 virtual machines are run in parallel.  `0.45C` is the value that @jglick uses on his macOS laptop with good results.  I use `1C` on the older separate machines that I use to run tests.

jglick: Ahem, I use a ThinkPad with Linux. :grin:

README.adoc
```diff
@@ -34,6 +34,7 @@ buildPlugin(
 * `repo` (default: `null`  inherit from Multibranch) - custom Git repository to check out
 * `useContainerAgent` (default: `false`) - uses a link:https://github.com/jenkins-infra/documentation/blob/main/ci.adoc#container-agents[Container agent] instead of a Virtual Machine: usually faster to start and generates less costs for the project
 ** Please note that the implementation of "containers" can be changed over time
+* `forkCount` (default: null) - Execute tests with `forkCount` Java virtual machines. Common values include `0.45C`, `1`, `2`, and `1C` (if you terminate the value with a 'C', that value will be multiplied by the number of available CPU cores).
```
jglick, line 37: My local setting of `0.45C` is probably inappropriate for a CI machine. I chose this specifically because I do _not_ want a desktop OS to be entirely consumed with tests since it can cause GUI applications to freeze up.

lemeurherve (approved):

test/groovy/BuildPluginStepTests.groovy
```diff
@@ -148,6 +148,28 @@ class BuildPluginStepTests extends BaseTest {
...
   }
 
+  @Test
+  void test_buildPlugin_with_forkCount() throws Exception {
+    def script = loadScript(scriptName)
+    // Run with forkCount=1C
+    script.call([forkCount: '1C'])
+    printCallStack()
+    // and confirm expected message is output
+    assertTrue(assertMethodCallContainsPattern('echo', 'Running parallel tests with forkCount'))
```
jglick, line 158: Would it not be more relevant to assert the actual `-DforkCount` parameter in the command?

README.adoc
jglick, line 37: May want to explain that this controls the number of JVMs used for the test code itself. That will also include code under test in the case of unit tests and `JenkinsRule`, but additional JVMs may be launched if you attach agents or use `RealJenkinsRule`, and especially in those cases setting this too high can lead to memory problems.

jglick, line 37: May be helpful to note that the default of `null` means to use Surefire’s default, which would normally be the same as `1`: one test at a time, run serially.

jglick (approved):
Straightforward. Note that for plugins with a lot of tests, like `workflow-cps`, much more significant speedups would be accomplished with use of https://plugins.jenkins.io/parallel-test-executor/.

test/groovy/BuildPluginStepTests.groovy
MarkEWaite: I tried to assert the actual argument through multiple paths and could not find one that would detect it.  The tests do not display (or execute) the maven command, so the maven command is not visible in the output.  Since it is not visible in the output, I could not assert its presence or absence.

I used this technique (with a little bit of frustration) because it is the technique that is used by one of the other tests.  I think that the message may actually be helpful to users if they are puzzled if CI tests fail differently than their desktop tests.

README.adoc
MarkEWaite: Very good point.

MarkEWaite: Done in 942490c939a853e94f38f43fcfe9df1776dfafc3

MarkEWaite: done in 942490c939a853e94f38f43fcfe9df1776dfafc3

test/groovy/BuildPluginStepTests.groovy
jglick: > The tests do not display (or execute) the maven command

Right but I thought there was something that mocked out the `sh` step so you can run assertions against it. But I am not very familiar with the tests in this repo.

jglick: > the message may actually be helpful to users if they are puzzled if CI tests fail differently than their desktop tests

Perhaps, though if in doubt you should always refer to the actual command being run according to the log:

```
+ mvn …
```

README.adoc
jglick: The explanation is mangled I think. The fork count should be the (maximum) number of concurrent JVMs running test code and any same-process code under test, such as controller code when using `JenkinsRule`. There will be _additional_ JVMs if you start agents or use `RealJenkinsRule`.

MarkEWaite: Thanks.  I had misunderstood.  Corrected in fe9fb49f869230697992caf599e3de4e4b02313a

MarkEWaite: Strange that my local git client reports that the commit fe9fb49 has been pushed, but the GitHub user interface does not show it as pushed. I can pull the change from another machine, but the UI does not show it...

timja: https://www.githubstatus.com/

image

jglick (approved):

Issue #643, opened 2023-04-10
Library build fails on Java 17


uhafner:
Is there a special environment required to get a green build locally? 

When I run `mvn --no-transfer-progress -B clean verify`

I get 
```
Apache Maven 3.9.1 (2e178502fcdbffc201671fb2537d0cb4b4cc58f8)23
Maven home: /usr/local/Cellar/maven/3.9.1/libexec
Java version: 17.0.6, vendor: Homebrew, runtime: /usr/local/Cellar/openjdk@17/17.0.6/libexec/openjdk.jdk/Contents/Home
Default locale: de_DE, platform encoding: UTF-8
OS name: "mac os x", version: "13.3.1", arch: "x86_64", family: "mac"
 [ ...]
[ERROR] Failures: 
[ERROR]   BuildPluginStepTests.test_getConfigurations_implicit_with_jdkVersions:107 expected:<[null, null, null, null]> but was:<4>
[ERROR]   BuildPluginStepTests.test_getConfigurations_implicit_with_jenkinsVersions:96 expected:<[null, null, null, null]> but was:<4>
[ERROR]   BuildPluginStepTests.test_getConfigurations_implicit_with_platforms:85 expected:<[null, null]> but was:<2>
[ERROR]   LaunchableStepTests.test_launchable_install:21
[INFO] 
[ERROR] Tests run: 106, Failures: 4, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
```

dduportal:
Could you try with JDK11 (as per https://github.com/jenkins-infra/pipeline-library/blob/d6e19b741e7ea049c4e6761763f125e24945b06c/Jenkinsfile#L4).

I'm not sure if the documentation is listing JDK11 as a requirement, but I clearly recall JDK17 failing 8-10 month ago

uhafner:
Yes, this is exactly the problem. When switching to Java 11 the build is green. 

(It would be also helpful if the documentation would describe, how we could import the project in IntelliJ, so that the tests can be run in the IDE and not in the command line).

dduportal:
Gotcha for the jdk17 part 👍🏼

i won’t be able to help for the IDE part, as i do not one. I assume IntelliJ has a « maven project » support that should be able to run tests no ?

PR #708, opened 2023-07-28
Fix Java 17 tests

MarkEWaite:
## Fix Java 17 tests

Call size as a method, not an attribute.

Fixes https://github.com/jenkins-infra/pipeline-library/issues/643


dduportal (approved):
LGTM! Tested with Maven 3.9.3 and Temurin JDK 17.0.7+7 with success

PR #709, opened 2023-07-28
Make tests work on Java 21 and newer

NotMyFault:
Addresses asm shenanigans to make tests work on Java 21 and above.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
     <dependency>
       <groupId>org.codehaus.groovy</groupId>
       <artifactId>groovy-all</artifactId>
-      <version>3.0.13</version>
+      <version>3.0.18</version>
       <!-- Type pom replaces the über "groovy-all.jar", for version 3+ -->
       <type>pom</type>
       <scope>test</scope>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -132,8 +132,9 @@
         <extensions>true</extensions>
       </plugin>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
NotMyFault, line 135: Dependabot looked for `org.codehaus.groovy:maven-compiler-plugin` for some reason, that's why this wasn't caught.

MarkEWaite (approved):
Confirmed locally that with this change the tests pass for Java 11 and Java 17, but they fail for me when run with Java 21 early access `OpenJDK Runtime Environment (build 21-ea+32-2482)` on amd64 Linux (Ubuntu 22.04).

What version of Java 21 early access are you running and on what platform?

timja (approved):

NotMyFault:
> What version of Java 21 early access are you running and on what platform?

MacOS 13.4, build 21-ea+33-2490 from https://jdk.java.net/21/ with maven 3.9.3, calling the same steps the Jenkinsfile does.

jglick (approved):

PR #724, opened 2023-08-08
feat(docker): multiplatforms builds v2

smerle33:
POC

### DIFF
--- a/resources/io/jenkins/infra/docker/Makefile
+++ b/resources/io/jenkins/infra/docker/Makefile
@@ -34,8 +34,8 @@ lint: ## Lint the $(IMAGE_DOCKERFILE) content
 
 build: ## Build the Docker Image $(IMAGE_NAME) from $(IMAGE_DOCKERFILE)
 	@echo "== Building $(IMAGE_NAME) from $(call FixPath,$(IMAGE_DOCKERFILE)) with docker:"
-	docker buildx ls
-	docker context ls
+	docker buildx create --use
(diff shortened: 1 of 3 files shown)


resources/io/jenkins/infra/docker/Makefile
```diff
@@ -34,9 +34,9 @@ lint: ## Lint the $(IMAGE_DOCKERFILE) content
 
 build: ## Build the Docker Image $(IMAGE_NAME) from $(IMAGE_DOCKERFILE)
 	@echo "== Building $(IMAGE_NAME) from $(call FixPath,$(IMAGE_DOCKERFILE)) with docker:"
-	docker buildx create --use
-    docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
-	docker buildx build \
+		docker buildx create --use
```
dduportal, line 37: Could you give a name to this "builder instance" and only create it if not existing?

Also, can you remove it after the `docker run` is finished with success (no need to take care of failures: that's what the condition would be there for)

smerle33: added in https://github.com/jenkins-infra/pipeline-library/pull/724/commits/75be361a5695998ba0f4f9e66d2da7c949d1c483

smerle33: removed as per rubber duck

smerle33:
close as per `ERROR: docker exporter does not currently support exporting manifest lists`

PR #783, opened 2023-11-06
chore(infra.withDockerCredentials): fix `--password via the CLI is insecure` warning

lemeurherve:
This PR removes the first following warning from logs when calling `infra.withDockerCredentials` function on Windows agents:

```console
powershell.exe : WARNING! Using --password via the CLI is insecure. Use --password-stdin.
 At C:\Jenkins\agent\workspace\Packaging_docker_PR-1770@tmp\durable-ba669f40\powershellWrapper.ps1:3 char:1
 + & powershell -NoProfile -NonInteractive -ExecutionPolicy Bypass -Comm ...
 + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     + CategoryInfo          : NotSpecified: (WARNING! Using ...password-stdin.:String) [], RemoteException
     + FullyQualifiedErrorId : NativeCommandError
  
 WARNING! Your password will be stored unencrypted in C:\Windows\system32\config\systemprofile\.docker\config.json.
 Configure a credential helper to remove this warning. See
 https://docs.docker.com/engine/reference/commandline/login/#credentials-store
 
 Login Succeeded
```

Note: the second warning should also be dealt with (not in this PR).

Tested in https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/107

https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/PR-107/2/console ✅ 

Closes https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/107

### DIFF
--- a/test/groovy/BaseTest.groovy
+++ b/test/groovy/BaseTest.groovy
@@ -65,6 +65,7 @@ class BaseTest extends DeclarativePipelineTest {
 
     helper.registerAllowedMethod('sh', [String.class], { s -> s })
     helper.registerAllowedMethod('powershell', [String.class], { s -> s })
+    helper.registerAllowedMethod('pwsh', [String.class], { s -> s })
     helper.registerAllowedMethod('stage', [String.class], { s -> s })
     helper.registerAllowedMethod('timeout', [String.class], { s -> s })
     helper.registerAllowedMethod('timeout', [Integer.class, Closure.class], { list, body -> body() })
--- a/test/groovy/InfraStepTests.groovy
+++ b/test/groovy/InfraStepTests.groovy
@@ -135,7 +135,7 @@ class InfraStepTests extends BaseTest {
     printCallStack()
     assertTrue(isOK)
     assertJobStatusSuccess()
-    assertTrue(assertMethodCallContainsPattern('powershell', 'Invoke-Expression "${Env:CONTAINER_BIN} login --username ${Env:DOCKER_CONFIG_USR} --password ${Env:DOCKER_CONFIG_PSW}"'))
+    assertTrue(assertMethodCallContainsPattern('powershell', 'Invoke-Expression "${Env:CONTAINER_BIN} login --username ${Env:DOCKER_CONFIG_USR} --password ${Env:DOCKER_CONFIG_PSW} 2>&1 > $tmpFolder/discarded_output.log"'))
   }
 
   @Test
--- a/test/groovy/InfraStepTests.groovy
+++ b/test/groovy/InfraStepTests.groovy
@@ -135,7 +135,7 @@ class InfraStepTests extends BaseTest {
     printCallStack()
     assertTrue(isOK)
     assertJobStatusSuccess()
-    assertTrue(assertMethodCallContainsPattern('powershell', 'Invoke-Expression "${Env:CONTAINER_BIN} login --username ${Env:DOCKER_CONFIG_USR} --password ${Env:DOCKER_CONFIG_PSW} 2>&1 > $tmpFolder/discarded_output.log"'))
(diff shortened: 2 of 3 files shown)


timja (approved):

dduportal:
The warning is there for a reason: we should better spend time fixing it instead of hiding it.

- As per https://docs.docker.com/engine/reference/commandline/login/#password-stdin, the `--password-stdin` flag should be used in a non-interactive setup such as a Jenkins pipeline
- https://stackoverflow.com/questions/51489359/docker-using-password-via-the-cli-is-insecure-use-password-stdin as plenty example including Windows powershell and batch for using this flag

dduportal:
As for the second warning, the message is pretty self-explanatory:

```
WARNING! Your password will be stored unencrypted in C:\Windows\system32\config\systemprofile\.docker\config.json.
 Configure a credential helper to remove this warning. See
 https://docs.docker.com/engine/reference/commandline/login/#credentials-store
```

https://docs.docker.com/engine/reference/commandline/login/#credentials-store explains what to do :)

lemeurherve:
> The warning is there for a reason: we should better spend time fixing it instead of hiding it.
> 
>     * As per https://docs.docker.com/engine/reference/commandline/login/#password-stdin, the `--password-stdin` flag should be used in a non-interactive setup such as a Jenkins pipeline
> 
>     * https://stackoverflow.com/questions/51489359/docker-using-password-via-the-cli-is-insecure-use-password-stdin as plenty example including Windows powershell and batch for using this flag

I tried that, see https://github.com/jenkins-infra/pipeline-library/pull/783/commits/0b6c170e1231568d8053f0136d853f40c1087ea1

With the following command `powershell 'Write-Output ${env:DOCKER_CONFIG_PSW} | & ${Env:CONTAINER_BIN} login --username ${Env:DOCKER_CONFIG_USR} --password-stdin'`, I'm successfully log in locally, but it doesn't seem to work on infra.ci.jenkins.io.

I suspect that `withCredentials` is masking the password between the first `Write-Output` and the right part of the pipe on Windows (while the same pattern is working fine on Linux). I don't explain why it doesn't work otherwise, but I could be completely wrong, if you have any idea I'm all ears!

lemeurherve:
> > The warning is there for a reason: we should better spend time fixing it instead of hiding it.
> > 
> > * As per https://docs.docker.com/engine/reference/commandline/login/#password-stdin, the `--password-stdin` flag should be used in a non-interactive setup such as a Jenkins pipeline
> > * https://stackoverflow.com/questions/51489359/docker-using-password-via-the-cli-is-insecure-use-password-stdin as plenty example including Windows powershell and batch for using this flag
> 
> I tried that, see [0b6c170](https://github.com/jenkins-infra/pipeline-library/commit/0b6c170e1231568d8053f0136d853f40c1087ea1)
> 
> With the following command I'm successfully logging in locally, but it doesn't seem to work on infra.ci.jenkins.io:
> 
> ```powershell
> Write-Output ${env:DOCKER_CONFIG_PSW} | & ${Env:CONTAINER_BIN} login --username ${Env:DOCKER_CONFIG_USR} --password-stdin
> ```
> 
> I suspect that `withCredentials` is masking the password between the first `Write-Output` and the right part of the pipe on Windows (while the same pattern is working fine on Linux). I don't explain why it doesn't work otherwise, but I could be completely wrong, if you have any idea I'm all ears!

The `withCredentials` wasn't responsible at all, switching from `powershell` (version 5) to `pwsh` (version 7) fixed the command above 🎉

dduportal (approved):

PR #809, opened 2023-12-13
chore: update README.adoc with latest archetype

lemeurherve:
This PR updates the `buildPlugin` example with the latest archetype content from https://github.com/jenkinsci/archetypes/blob/master/common-files/Jenkinsfile

I've picked the last Weekly release as example for Jenkins version parameter.


dduportal (approved):

jglick:
Would it not be better to just link to that official template?

PR #838, opened 2024-02-10
chore(terraform): show with a GitHub check that shipping change failed

lemeurherve:
This PR adds a status check to know when shipping change failed and that the build is in standby waiting for input from user, with a link in the check run redirecting to the build console log URL where the input links are.

### DIFF
--- a/vars/terraform.groovy
+++ b/vars/terraform.groovy
@@ -129,7 +129,12 @@ def call(userConfig = [:]) {
                   sh makeCliCmd + ' deploy'
                 } catch(Exception e) {
                   // If the deploy failed, keep the pod until a user catch the problem (cloud be an errored state, or many reason to keep the workspace)
-                  input message: 'An error happened while applying the terraform plan. Keeping the agent up and running. Delete the agent?'
+                  final String msg = 'An error happened while applying the terraform plan. Keeping the agent up and running. Delete the agent?'
+                  input message: msg
+                  publishChecks name: 'deploy-error',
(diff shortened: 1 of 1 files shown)


vars/terraform.groovy
```diff
@@ -129,7 +129,12 @@ def call(userConfig = [:]) {
                   sh makeCliCmd + ' deploy'
                 } catch(Exception e) {
                   // If the deploy failed, keep the pod until a user catch the problem (cloud be an errored state, or many reason to keep the workspace)
-                  input message: 'An error happened while applying the terraform plan. Keeping the agent up and running. Delete the agent?'
+                  final String msg = 'An error happened while applying the terraform plan. Keeping the agent up and running. Delete the agent?'
+                  input message: msg
+                  publishChecks name: 'deploy-error',
```
dduportal, line 134: At first glance, this won't work as you expect (unless you tested it successfully?).

The step `input` is synchronous so the `publishChecks` step won't be executed unless an admin clicks on "continue" in the input step (which.. defeats your purpose here).


I wonder if inverting the 2 steps would not be enough (the risk of false positive looks low but only a gut feeling).

dduportal (changes requested):

vars/terraform.groovy
lemeurherve: I've put the `publishCheck` in a `try` block, with the `input` as `finally`.

dduportal (approved):

PR #846, opened 2024-03-21
feat: allow using storage account key to generate a file share SAS token

lemeurherve:
This PR allows using a storage account access key to generate a file share SAS token instead of a service principal, useful when this access key is already available in the script execution context.

Tested in situ with:
- https://github.com/jenkins-infra/plugin-site/pull/1661

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3414


dduportal (approved):
👏 🚀

PR #869, opened 2024-07-01
debug: remove the NTP option for debug

dduportal:
Used for debugging: https://github.com/jenkins-infra/helpdesk/issues/4157


lemeurherve:
As it's not the first nor the last time we're doing this, WDYT of adding an option for disabling this NTP?

dduportal:
Closing as not needed anymore

dduportal:
> As it's not the first nor the last time we're doing this, WDYT of adding an option for disabling this NTP?

Could be an option. Please open an issue or open a PR if interested.

PR #877, opened 2024-08-10
fix(buildPlugin) setup Artifact Caching Proxy using only the env. var `ARTIFACT_CACHING_PROXY_SERVERID`

dduportal:
Ref. https://github.com/jenkins-infra/jenkins-infra/pull/3592

### DIFF
--- a/test/groovy/InfraStepTests.groovy
+++ b/test/groovy/InfraStepTests.groovy
@@ -10,10 +10,7 @@ import mock.PullRequest
 
 class InfraStepTests extends BaseTest {
   static final String scriptName = "vars/infra.groovy"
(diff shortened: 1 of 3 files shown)


dduportal:
Tested with success with build replays https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/master/243 and https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/master/244 covering the 4 main new cases:

- Nominal case with the environment variable set at the agent level (Linux container and Linux VM)
- Case with no environment variable set in the agent template (Windows VM)
- Case with environment variable set but unreachable (Windows Container)

PR #879, opened 2024-08-28
feat(BuildDockerAndPublishImage): publish checks only on other branch than `main` to avoid GH rate limit

smerle33:
as per https://www.jenkins.io/doc/pipeline/steps/warnings-ng/ 
setting `skipPublishingChecks` to false on the main branch will limit to bump into the GH rate limit 

```
Could not update commit status. Message: {"message":"API rate limit exceeded for installation ID <redacted>. If you reach out to GitHub Support for help, please include the request ID <redacted> and timestamp 2024-08-27 16:49:25 UTC.","documentation_url":"https://docs.github.com/rest/overview/rate-limits-for-the-rest-api","status":"403"}
```

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -196,6 +196,7 @@ def call(String imageShortName, Map userConfig=[:]) {
               }
             } finally {
               recordIssues(
+                  skipPublishingChecks: (env.BRANCH_IS_PRIMARY),
                   enabledForFailure: true,
                   aggregatingResults: false,
                   tool: hadoLint(id: hadolintReportId, pattern: hadoLintReportFile)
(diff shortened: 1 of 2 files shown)


vars/buildDockerAndPublishImage.groovy
```diff
@@ -196,7 +196,7 @@ def call(String imageShortName, Map userConfig=[:]) {
               }
             } finally {
               recordIssues(
-                  skipPublishingChecks: (env.BRANCH_IS_PRIMARY),
+                  skipPublishingChecks: "${env.BRANCH_IS_PRIMARY}",
```
dduportal, line 199:
```suggestion
                  skipPublishingChecks: env.BRANCH_IS_PRIMARY,
```

There are no need to interpolate here: we are in Groovy so you can directly call the variable. Types are compatible (both are booleans) so you're good

dduportal (approved):

PR #880, opened 2024-08-30
fix(buildDockerAndPublishImage) avoid wrong type inference for `recordIssues(skipPublishingChecks=)` argument

dduportal:
This change fixes the error below caused by #879.

~I'm not sure why we have this error as #879 was tested by a dummy PR on a docker-jenkins repo.~
(edited: found the cause: we tested before my suggestion - https://github.com/jenkins-infra/pipeline-library/pull/879#discussion_r1735795782. The code before suggestion worked as it was inffering the `null` to string and then to boolean)

Let's retry with 2 PRs end to end testing this time (including openvpn)

```
Found unhandled java.lang.IllegalArgumentException exception:

Could not instantiate {skipPublishingChecks=null, enabledForFailure=true, aggregatingResults=false, tool=@hadoLint(id=jenkinsciinfra-openvpn-hadolint-1725004944771,pattern=jenkinsciinfra-openvpn-hadolint-1725004944771.json)} for io.jenkins.plugins.analysis.core.steps.RecordIssuesStep: java.lang.IllegalArgumentException
```

----

Tested with 2 pipeline replay builds prefixed with the instruction `@Library('pipeline-library@pull/880/head') _`

- Case with principal branch in https://infra.ci.jenkins.io/job/docker-jobs/job/docker-404/job/main/138/ => checks were not published
- Case with a PR in https://infra.ci.jenkins.io/job/docker-jobs/job/docker-openvpn/job/PR-353/2 => checks were published in the PR as expected

----

### DIFF
--- a/test/groovy/BuildDockerAndPublishImageStepTests.groovy
+++ b/test/groovy/BuildDockerAndPublishImageStepTests.groovy
@@ -117,9 +117,13 @@ class BuildDockerAndPublishImageStepTests extends BaseTest {
   // Return if the usual static checks had been recorded with the usual pattern
   Boolean assertRecordIssues(String imageName = fullTestImageName) {
     final String reportId = "${imageName}-hadolint-${mockedTimestamp}".replaceAll('/','-').replaceAll(':', '-')
+    boolean skipChecks = false
+    if (env.BRANCH_IS_PRIMARY) {
+      skipChecks = true
+    }
     return assertMethodCallContainsPattern(
         'recordIssues',
-        "{skipPublishingChecks=${env.BRANCH_IS_PRIMARY}, enabledForFailure=true, aggregatingResults=false, tool={id=${reportId}, pattern=${reportId}.json}}",
+        "{skipPublishingChecks=${skipChecks}, enabledForFailure=true, aggregatingResults=false, tool={id=${reportId}, pattern=${reportId}.json}}",
         )
   }
 
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -195,8 +195,12 @@ def call(String imageShortName, Map userConfig=[:]) {
                 powershell 'make lint'
               }
             } finally {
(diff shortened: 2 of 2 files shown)


test/groovy/BuildDockerAndPublishImageStepTests.groovy
```diff
@@ -117,9 +117,13 @@ class BuildDockerAndPublishImageStepTests extends BaseTest {
   // Return if the usual static checks had been recorded with the usual pattern
   Boolean assertRecordIssues(String imageName = fullTestImageName) {
     final String reportId = "${imageName}-hadolint-${mockedTimestamp}".replaceAll('/','-').replaceAll(':', '-')
+    boolean skipChecks = false
+    if (env.BRANCH_IS_PRIMARY) {
```
timja, line 121: you tested this? as an env variable will be a string this _may_ not work as expected

dduportal: Yep, I've updated the PR body with the tests after your comment.

The issue is that, due to the pipeline engine, the env.BRANCH_IS_PRIMARY was undefined (`null`) when running in a PR (hence the error message we had). No type inference was done so the `recordIssue()` function, expecting a boolean value for the argument `skipPublishingSteps` receieved a `null` value.

With this new code, using the Groovy features, https://groovy-lang.org/semantics.html#_strings empty string is coerced to `false` for the `if` condition and to `true` if not empty.

It's a codestyle tradeoff: it adds a code step, but it avoid letting the pipeline engine coercing types in the step call

dduportal: Groovy languages recommends to rely the emptiness coercion https://groovy-lang.org/semantics.html#_strings => it is a feature language we can rely on when using pure groovy instruction.

The inference issue came from calling Groovy feature in a pipeline step which can have.. weird behavior (hence the choice to move it in the  `if` case


Your suggestion is valid but un-needed. Would be true on Javascript though \o/

PR #893, opened 2024-12-17
Allow publishing attachment using junit-attachments

jonesbusy:
My use case is integration tests using maven invoker to run `plugin-modernizer` recipes againts sample plugin.

Since the `junit-attachment` plugin is installed, every maintainer can use it. I would propose to enable the publisher by default when we archive junit results

I've tested this PR by doing a print of the attachement

```java
System.out.printf("[[ATTACHMENT|%s]]%n", Plugin.build(plugin).getLogFile().toAbsolutePath());
```

https://ci.jenkins.io/job/Tools/job/plugin-modernizer-tool/view/change-requests/job/PR-463/35/testReport/io.jenkins.tools.pluginmodernizer.cli/CommandLineITCase/linux_21___Build__linux_21____testBuildMetadata_PluginMetadata__WireMockRuntimeInfo__1_/

attachment

### DIFF
--- a/test/groovy/BaseTest.groovy
+++ b/test/groovy/BaseTest.groovy
@@ -20,6 +20,7 @@ class BaseTest extends DeclarativePipelineTest {
     binding.setProperty('mvnSettingsFile', 'settings.xml')
 
     helper.registerAllowedMethod('archiveArtifacts', [Map.class], { true })
+    helper.registerAllowedMethod('attachments', [])
     helper.registerAllowedMethod('checkout', [String.class], { 'OK' })
     helper.registerAllowedMethod('configFile', [Map.class], { 'OK' })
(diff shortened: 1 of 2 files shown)


jonesbusy:
Hey @lemeurherve @dduportal what do you think about that ?

I ping you because the codeowner contains error and not sure if anyone will receive the notification

codeowner_errors

Thanks!

timja (approved):
Makes sense

timja:
> Hey @lemeurherve @dduportal what do you think about that ?
> 
> I ping you because the codeowner contains error and not sure if anyone will receive the notification
> 
> codeowner_errors
> 
> Thanks!

continue this in https://github.com/jenkins-infra/helpdesk/issues/4264

PR #897, opened 2025-01-07
Default to Java 11 for `jdkVersions`

### LINKED ISSUE #514: Default `buildPlugin()` and `buildPluginWithGradle()` to Java 11

jonesbusy:
Fix #514 but maybe the opinion is different 2.5 years after

There are still few plugin still using implicit configuration which default to Java 8 still.

Those plugin are difficult for non-maintainer or non-administrator to modernize because it require the Jenkinsfile to be replayed

It would broke some plugin maybe (some are already broken) but I will argue it will benefit a lot modernizing many plugin to Java 11 and above (In more less 6 month Java 17 will be the recommended version for plugin bom in any case)

If for some some plugin still need to build on Java 8 they can move to explicit configuration until Java 8 is completely removed from ci.jenkins.io

Any though ?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -38,7 +38,7 @@ buildPlugin(
 * `failFast` (default: `true`) - instruct Maven tests to fail fast
 * `platforms` (default: `['linux', 'windows']`) - Labels matching platforms to
   execute the steps against in parallel
-* `jdkVersions` (default: `[8]`) - JDK version numbers, must match a version
+* `jdkVersions` (default: `[11]`) - JDK version numbers, must match a version
   number jdk tool installed
 * `jenkinsVersions`: (default: `[null]`) - a matrix of Jenkins baseline versions to build/test against in parallel (null means default,
   only available for Maven projects)
--- a/test/groovy/BuildPluginStepTests.groovy
+++ b/test/groovy/BuildPluginStepTests.groovy
@@ -68,8 +68,8 @@ class BuildPluginStepTests extends BaseTest {
     def configurations = script.getConfigurations([:])
 
     def expected = [
-      ['platform': 'linux', 'jdk': '8', 'jenkins': null],
-      ['platform': 'windows', 'jdk': '8', 'jenkins': null],
+      ['platform': 'linux', 'jdk': '11', 'jenkins': null],
+      ['platform': 'windows', 'jdk': '11', 'jenkins': null],
     ]
     assertEquals(expected, configurations)
     printCallStack()
@@ -142,9 +142,9 @@ class BuildPluginStepTests extends BaseTest {
     script.call([useContainerAgent: true])
     printCallStack()
     // then it runs a stage in a linux container by default
-    assertTrue(assertMethodCallContainsPattern('node', 'maven'))
+    assertTrue(assertMethodCallContainsPattern('node', 'maven-11'))
     // then it runs a stage in a Windows container by default
-    assertTrue(assertMethodCallContainsPattern('node', 'maven-windows'))
+    assertTrue(assertMethodCallContainsPattern('node', 'maven-11-windows'))
     assertJobStatusSuccess()
   }
 
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -345,7 +345,7 @@ List<Map<String, String>> getConfigurations(Map params) {
   }
 
   def platforms = params.containsKey('platforms') ? params.platforms : ['linux', 'windows']
-  def jdkVersions = params.containsKey('jdkVersions') ? params.jdkVersions : ['8']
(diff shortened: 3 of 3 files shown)


dduportal:
Thanks for the PR @jonesbusy ! Looks a good idea. 

The SRE in me says it is a good idea, but this is a collective decision to be made: are you ok to add this to the agenda of an upcoming SIG platform please?

=> additionally, it require a bit more work, as we should identify the plugins which will be broken to make sure we don't have bad surprises. Maybe we can accept that this "work" would be "merge this PR, trigger build on all ci.jio plugins and check which one are failing". This won't be exhaustive (as not all plugins are using ci.jio) but it would cover the most used plugin as far as I know.


Any thoughts other key players here @basil @MarkEWaite @daniel-beck @wadeck @jglick @jtnord @poddingue @timja @mawinter69 @slides ?

timja (approved):
Fine with me, they'll likely still fail though due to the same problems with old parents etc though won't they? Although I guess this makes it easier to fix for non maintainers

timja:
> are you ok to add this to the agenda of an upcoming SIG platform please?

I'm unsure if many of the involved people would attend that, I think either this PR or the mailing list would be most appropriate.

jonesbusy:
Answered  https://groups.google.com/g/jenkinsci-dev/c/pjfvsMw-EMM

gounthar:
I have [a list](https://github.com/gounthar/jdk8-removal/tree/main/reports) of plugins without a Jenkinsfile, plugins using an old Java version, and plugins not defining a Java version, just in case. 🤷

jtnord:
> Those plugin are difficult for non-maintainer or non-administrator to modernize because it require the Jenkinsfile to be replayed

if you are attempting to modernise a plugin though, why would you use a java 11 version over java17 which has been required since 2.462.1? 

How is this different to a non maintainer modernising a plugin to support java21 / jakartaee or the like where the maintainer has explicitly put in versions?

Personally I am of the belief that the version should be explicit and the default should eventually die, adn this PR makes this harder by keeping it alive. Whilst that makes it harder for non maintainers, unless we are constantly bumping the default (for java17 then java21) then we are just kicking a can down the road for the few that have not explicitly set the version.

I am not going to die on this hill.

jonesbusy:
> though

Yes but at least PR that modernize them will not

> if you are attempting to modernise a plugin though, why would you use a java 11 version over java17 which has been required since 2.462.1?

If I'm not wrong Java 17 is required since 2.479.1. We still recommend 2.452.4 which is still compatible with Java 11 (https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/#currently-recommended-versions) https://www.jenkins.io/doc/book/platform-information/support-policy-java/#running-jenkins-system

The reason we have an archetype Jenkinsfile with Java 17 and 17 is because we didn't detected issues that would occur on Java 11 and not 17 or 21 (This save also cost by removing a matrix from the build).

Don't get me wrong, I think those implicit version are bad and must be removed at some point.

The point is with a default on Java 8 it make it difficult to align them to the archetype and would cause many build issue for non-maintainer when opening PR (Which lower a lot chances for a PR gets merged)

My opinion is that this PR is a step further not a step back

jtnord:
> If I'm not wrong Java 17 is required since 2.479.1. We still recommend 2.452.4 which is still compatible with Java 11 (https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/#currently-recommended-versions) https://www.jenkins.io/doc/book/platform-information/support-policy-java/#running-jenkins-system

Yup, my mistake.

jonesbusy:
Make sense, but perhaps it could be the next steps (maybe when 2.479.1 is the minimum recommended version ?

Maybe a less "big bang" approach 

1) We switch to default Java 11 on infra library 
2) We deprecate implicit settings and communicate (without faling build yet)
3) We modernize as much plugin as possible using minimum Java 11 and Jenkinsfile archetype (Java 17 and Java 21) so that PR doesn't fail and anyone including non-maintainer can contribute 
4) We remove implicit settings, and fail build 

What do you think?

jglick (changes requested):
https://github.com/jenkins-infra/pipeline-library/pull/897#issuecomment-2575573499 I think we should just start failing the build now if you do not specify your environment explicitly. I do not see any benefit in trying to keep CI passing for new pull requests filed by contributors against completely unmaintained plugins; the first thing that needs to happen on such a plugin is that the maintainer merge a PR to update metadata files to something recent.

Alternately, just close this PR (which serves no purpose I can see), and do 2–4.

PR #917, opened 2025-03-19
feat(buildPlugin): use a spot agent for the first try, normal one for the second

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4598

try on a spot instance first, and fall back on a non spot one for the second try

⚠️ Need to be tested, do not merge ⚠️

### DIFF
--- a/vars/buildPlugin.groovy
+++ b/vars/buildPlugin.groovy
@@ -38,30 +38,35 @@ def call(Map params = [:]) {
     boolean skipTests = params?.tests?.skip
     boolean addToolEnv = !useContainerAgent
 
-    if (useContainerAgent) {
-      if (platform == 'linux' || platform == 'windows') {
-        def agentContainerLabel = jdk == '8' ? 'maven' : 'maven-' + jdk
-        if (platform == 'windows') {
-          agentContainerLabel += '-windows'
-        }
-        label = agentContainerLabel
-      }
-    } else {
-      switch(platform) {
-        case 'windows':
-          label = 'docker-windows'
-          break
-        case 'linux':
-          label = 'vm && linux'
-          break
-        default:
-          echo "WARNING: Unknown Virtual Machine platform '${platform}'. Set useContainerAgent to 'true' unless you want to be in uncharted territory."
-          label = platform
-      }
-    }
-
     tasks[stageIdentifier] = {
-      retry(count: 3, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+      int retryCounts = 1
+      retry(count: 2, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+        if (useContainerAgent) {
+          if (platform == 'linux' || platform == 'windows') {
+            def agentContainerLabel = jdk == '8' ? 'maven' : 'maven-' + jdk
+            if (platform == 'windows') {
+              agentContainerLabel += '-windows'
+            }
+            label = agentContainerLabel
+          }
+        } else {
+          switch(platform) {
+            case 'windows':
+              label = 'docker-windows'
+              break
+            case 'linux':
+              label = 'vm && linux'
+              break
+            default:
+              echo "WARNING: Unknown Virtual Machine platform '${platform}'. Set useContainerAgent to 'true' unless you want to be in uncharted territory."
+              label = platform
+          }
+        }
+        if (retryCounts == 1 && platform != 'windows') { // no spot instances for windows for now TODO change when available
+          // Use a spot instance for the first try
+          label += ' && spot'
+        }
+        retryCounts = retryCounts + 1
         node(label) {
           try {
             timeout(timeoutValue) {


smerle33:
test : https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/master/305/
spot instance launch: `Running on [EC2 (aws-us-east-2) - Spot Ubuntu 22.04 LTS x86_64 with JDK17 set as default java CLI (i-06d00494ca1448496)](https://ci.jenkins.io/computer/EC2%20(aws-us-east-2)%20-%20Spot%20Ubuntu%2022.04%20LTS%20x86_64%20with%20JDK17%20set%20as%20default%20java%20CLI%20(i-06d00494ca1448496)/)`

but in real non spot 

Capture d’écran 2025-03-19 à 18 15 24

may need to ask for spot instance of this type

smerle33:
This new build https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/master/306/, ran on a spot instance: 

Capture d’écran 2025-03-20 à 09 55 12

vars/buildPlugin.groovy
```diff
@@ -61,7 +61,14 @@ def call(Map params = [:]) {
...
     tasks[stageIdentifier] = {
-      retry(count: 3, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+      int retryCounts = 1
+      retry(count: 2, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+        if (retryCounts == 2 && platform != 'windows') {
+          // no spot instances for windows for now TODO change when available
+          // Use a spot instance for the first try and nonspot for second
+          label = 'ubuntu-22-amd64-maven17-nonspot'
+        }
+        retryCounts = retryCounts + 1
```
timja, line 71: why change this from 3?
```suggestion
      int retryCount = 0
      retry(count: 3, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
        if (retryCount == 1 && platform != 'windows') {
          // no spot instances for windows for now TODO change when available
          // Use a spot instance for the first try and nonspot for second
          label = 'ubuntu-22-amd64-maven17-nonspot'
        }
       retryCount++
```

smerle33: only 2 retry instead of 2, it's just to match one on spot, one on non spot ...
we can also change line 66 to != 0 then nonspot and put back 3 tries ... (1 spot, 3 nonspot)

WDYT ?

dduportal: @smerle33 I missed the old value of 3. But it make sense to retry twice (with spot) and only fallback ondemand for the last one (that was the previous behavior). Tim's suggestion looks good.

vars/buildPlugin.groovy
```diff
@@ -61,7 +61,14 @@ def call(Map params = [:]) {
     }
 
     tasks[stageIdentifier] = {
-      retry(count: 3, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+      int retryCounts = 1
+      retry(count: 2, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
```
dduportal, line 65: You need to add the `agent()` condition to map what we have in ATH: https://github.com/jenkinsci/acceptance-test-harness/blob/ecaf4fc38523b77e0997daa4285999965276a3e2/Jenkinsfile#L116

I don't know if it is cumulative or mutually exclusive with kubernetesAgent though: if we can only spcify on of each, we'll need a tiny function with a closure to sleect retrys based on a condition, and then run the body.

dduportal: Or maybe not needed thanks to `handleNonKubernetes` (ref. https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#retry-retry-the-body-up-to-n-times) 🤔

smerle33: I think that it's why `handleNonKubernetes` is for :  https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#retry-retry-the-body-up-to-n-times
```
    handleNonKubernetes : boolean (optional)
    Behave like the generic agent() (Agent errors) when applied to a non-Kubernetes agent. Useful in cases where it is hard to predict in a job definition whether a Kubernetes or other sort of agent will be used.
```

smerle33: o sorry, the comments were not uptodate

vars/buildPlugin.groovy
```diff
@@ -61,7 +61,14 @@ def call(Map params = [:]) {
...
 
     tasks[stageIdentifier] = {
+      int retryCounts = 1
       retry(count: 3, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+        if (retryCounts >= 3 && platform != 'windows') {
+          // no spot instances for windows for now TODO change when available
+          // Use a spot instance for the 2 first try and nonspot for third and last
+          label = 'ubuntu-22-amd64-maven17-nonspot'
+        }
+        retryCounts = retryCounts + 1
```
timja, line 71: this is cleaner (and basically what I suggested before..)

count should start at 0 as its not being retried

```suggestion
      int retryCounter = 0
      retry(count: 3, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
        if (retryCounter > 1 && platform != 'windows') {
          // no spot instances for windows for now TODO change when available
          // Use a spot instance for the 2 first try and nonspot for third and last
          label = 'ubuntu-22-amd64-maven17-nonspot'
        }
        retryCounter++
```

vars/buildPlugin.groovy
```diff
@@ -61,7 +61,14 @@ def call(Map params = [:]) {
     }
 
     tasks[stageIdentifier] = {
+      int retryCounter = 0
       retry(count: 3, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+        if (retryCounter > 1 && platform != 'windows') {
+          // no spot instances for windows for now TODO change when available
+          // Use a spot instance for the 2 first try and nonspot for third and last
+          label = 'ubuntu-22-amd64-maven17-nonspot'
```
dduportal, line 69:
```suggestion
          label += '-nonspot'
```

to avoid fragility of the string

smerle33: but the `original` agent label is 'vm && linux' https://github.com/jenkins-infra/pipeline-library/pull/917/files#diff-db2794773b738dbb5566313f115cdf3ff1f45e13e57cbc37aa1bd181819ed5fbL55 in order to use the nonspot we define the complete label but the "spot" one is 'vm && linux' ...

jglick, line 69: Why not just use labels the way they were intended?
```suggestion
        if (retryCounter > 1) {
          // Use a spot instance (if applicable) for the 2 first try and nonspot for third and last
          label += ' && !spot'
```
assuming of course that you have marked the usual Linux agents with the `spot` label. Then there is no need to hard-code a particular label expression for the rest, and if we do add Windows agents supporting Spot later, fine: just mark them with the `spot` label, and offer an alternate pool without it for on-demand.

smerle33: if we just add a `spot` label and use your idea of && spot or && !spot it's working for the shared pipeline library but we miss the point of other usage, meaning we want almost all call to agent to fall on spot ones, so default is spot ... and exception (or retry) is non spot ...

if both spot and nonspot got the same set of label, lets say vm && linux ... both can be called ...

if I understood correctly ...

jglick: Yeah you would want to use `&& spot` for the first two tries. Are there that many other pieces of Pipeline script running the system? `bom/Jenkinsfile`, `jenkins/Jenkinsfile`, and a few miscellaneous low-traffic projects I guess?

jglick: (You may be able to designate Spot-only `Cloud`s as first in the list and hope they service requests preferentially. Not sure.)

dduportal: > (You may be able to designate Spot-only Clouds as first in the list and hope they service requests preferentially. Not sure.)

With EC2 plugins, for sure it does not work :'( (it works with ACI, kubernetes and auzr-evm though)

smerle33: see https://github.com/jenkins-infra/jenkins-infra/pull/3978 that set the labels automatically and this one have been updated to match spot and then nonspot agents

vars/buildPlugin.groovy
```diff
@@ -61,7 +61,15 @@ def call(Map params = [:]) {
     }
 
     tasks[stageIdentifier] = {
+      int retryCounter = 0
       retry(count: 3, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+        if (retryCounter > 1) {
+          // Use a spot instance for the 2 first try [try 0 and 1] and nonspot for third and last [2]
```
dduportal, line 67:
```suggestion
          // Use a spot instance for the 2 first tries and nonspot for the last
```

dduportal:
Note: Until https://github.com/jenkinsci/ec2-plugin/pull/1015 (or equivalent) is merged, the detection of a failed EC2 VM agent takes 10 min.

vars/buildPlugin.groovy
```diff
@@ -49,19 +50,29 @@ def call(Map params = [:]) {
     } else {
```
lemeurherveCB, line 50: question(not directly related to this PR): should a warning be displayed and `label` set to `platform` in an `else` L49 if this one is different than "linux" or "windows"?

PR #919, opened 2025-03-29
fix(infra) fix a potential memory leak related to missing type for `orgAndCredentialsId`

dduportal:
This change aims at getting fixing the warning message below seen in our private controller pipeline outputs:

```
Did you forget the `def` keyword? infra seems to be setting a field named orgAndCredentialsId (to a value of type LinkedHashMap) which could lead to memory leaks or other issues.
```


MarkEWaite (approved):
Thanks!

PR #937, opened 2025-05-23
bump default java version to 21 in pipeline lib

clintonsteiner:


dduportal:
Hi @clintonsteiner what is the rationale behind this change? There are reasons why we use JDK17 on Windows.

dduportal:
> the default was set to jdk 8 which is out of support for jenkins

That is untrue for the use case here. We still need to provide JDK8 to many plugin users for their builds and suddendly changing the default would impact them.

If you want to help, please get in touch with the maintainers (community forum, elements chats, issues) explaining the problem you see and your plan to solve it. Sending PRs without any heads up shows you did not make the due diligence to evaluate impact of your proposed change.

JDK8 EOL is in 5 years as per https://endoflife.date/oracle-jdk so we (Jenkins infrastructure team) keeps providing it.

dduportal:
Hello there, I'm sorry I was too brief on the explanations.

But there has been an history around this "challenge" of using a certain JDK or not:

- https://github.com/jenkins-infra/pipeline-library/pull/897
- https://github.com/jenkins-infra/pipeline-library/pull/541
- https://github.com/jenkins-infra/pipeline-library/pull/937
- https://github.com/jenkins-infra/pipeline-library/issues/514

=> as you can see, there are no real consensus over the different actors, except for the fact that we should head towards forcing to define explicit environment (instead of relying to an implicit default which can change and breaks things)?

The complicated part is that bumping suddenly the default open the can of worms:
- It means that not only language feature should be updated in the code,
- But also the minimum supported Jenkins baseline (using pom parent version bumps) need to be bumped, which adds more potential code changes to apply

=> all of this is under a non controlled schedule for plugins contributors, plus adding the fact that some plugins are not maintained actively but still used.

If you are willing to help, it would be really useful to go in the direction of this comment: https://github.com/jenkins-infra/pipeline-library/pull/897#pullrequestreview-2535285086, e.g. starting to implement a failure when no environment is specified, along with the batch PRs to pipelines without environment set up.

PR #938, opened 2025-05-29
feat(publishBuildStatusReport): publishes controller job status on reports.jenkins.io

jayfranco999:
As per - https://github.com/jenkins-infra/helpdesk/issues/2843#issuecomment-2883040380

`publishBuildStatusReport` will publish the build status report on reports.jenkins.io using pre-existing `publishReports` shared library script.

Once completed we can move on to the next phase of monitoring the reports. 

Testing with https://github.com/jenkins-infra/docker-404/pull/46

### DIFF
--- a/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
+++ b/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -euo pipefail
+set -x
 
 # Extract hostname from JENKINS_URL
 CONTROLLER_HOSTNAME=$(echo "$JENKINS_URL" | sed 's|https\?://||' | cut -d'/' -f1)
--- a/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
+++ b/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
@@ -32,11 +32,3 @@ cd "$REPORT_DIR"
 azcopy logout >/dev/null 2>&1 || true
 test -z "${AZURE_FEDERATED_TOKEN_FILE:-}" || export AZCOPY_AUTO_LOGIN_TYPE=WORKLOAD
 azcopy login --identity
-
-# Try azcopy with error handling
-if ! azcopy copy "status.json" "$DESTINATION_URL"; then
-    echo "azcopy failed, collecting logs for debugging"
-    # Retrieve azcopy logs to archive them
-    cat /home/jenkins/.azcopy/*.log > "$WORKSPACE/azcopy.log" 2>/dev/null
-    exit 1
-fi
--- a/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
+++ b/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
@@ -23,7 +23,7 @@ cat > "$REPORT_FILE" << EOF
 EOF
 
 # Upload with azcopy
-DESTINATION_URL="https://buildsreportsjenkinsio.file.core.windows.net/builds-reports-jenkins-io/build_status_reports/$CONTROLLER_HOSTNAME/$JOB_NAME"
+DESTINATION_URL="https://buildsreportsjenkinsio.file.core.windows.net/builds-reports-jenkins-io/build_status_reports/$JENKINS_URL/$JOB_NAME"
 
 cd "$REPORT_DIR"
 azcopy logout 2>/dev/null || true
--- a/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
+++ b/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
@@ -1,9 +1,6 @@
 #!/bin/bash
 set -euo pipefail
 
-# Extract controller hostname from JENKINS_URL
-CONTROLLER_HOSTNAME=$(echo "$JENKINS_URL" | sed 's|https\?://||' | cut -d'/' -f1 | cut -d':' -f1 | sed 's/[^a-zA-Z0-9.-]/_/g')
-
 # Build file path
 REPORT_DIR="$WORKSPACE/build_status_reports/$CONTROLLER_HOSTNAME/$JOB_NAME"
 REPORT_FILE="$REPORT_DIR/status.json"
@@ -17,7 +14,7 @@ REPORT_TIMESTAMP=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
 # Generate JSON report
 cat > "$REPORT_FILE" << EOF
 {
-  "controller_hostname": "$CONTROLLER_HOSTNAME",
+  "controller_url": "$JENKINS_URL",
   "job_name": "$JOB_NAME",
   "build_number": "$BUILD_NUMBER",
   "build_status": "$BUILD_STATUS",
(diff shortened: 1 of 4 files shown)


dduportal (commented):
Unit tests are missing :)

vars/publishReports.groovy
```diff
@@ -31,7 +31,7 @@ def call(List<String> files, Map params = [:]) {
             uploadFlags = '--content-type="application/json"'
             break
           case ~/(?i).*\.js/:
-            uploadFlags = '--content-type="application/javascript"'
+            uploadFlags = '--content-type="application/javascript"'3
```
jayfranco999, line 34:
```suggestion
            uploadFlags = '--content-type="application/javascript"'
```

test/groovy/PublishBuildStatusReportStepTests.groovy
```diff
@@ -0,0 +1,79 @@
...
+    assertMethodCallContainsPattern('withEnv', 'ENV_CONTROLLER_HOSTNAME=unknown_controller')
+    assertMethodCallContainsPattern('withEnv', 'ENV_JOB_NAME=unknown_job')
+    assertMethodCallContainsPattern('withEnv', 'ENV_BUILD_NUMBER=unknown_build')
+    assertMethodCallContainsPattern('withEnv', 'ENV_BUILD_STATUS=UNKNOWN')
+  }
+
+  @Test
+  void it_does_nothing_on_non_principal_branch() throws Exception {
+    def script = loadScript(scriptName)
+    // No mockPrincipalBranch() call
```
jayfranco999, line 68:
```suggestion
```

jayfranco999:
Update:

Added unit tests.

dduportal (changes requested):
Thanks @jayfranco999 . The code still have many things to review: it works but it is really verbose for not a lot of values (LLM generated?) which adds a lot of complexity and instructions.

As a general rule of thumb: the less you do in Pipeline (e.g. Groovy), the safer we are.
You should always aim at running one `sh` step and build thing around it (such as the required `withEnv() wrapping)`.


=> A good start would be to remove the utility function for parsing Date and time, adn group the `sh` into a single step.

Don't hesitate to use my suggestions as starters

vars/publishBuildStatusReport.groovy
```diff
@@ -0,0 +1,42 @@
...
+ *     targetplatforms: 'linux/amd64,linux/arm64',
+ * ])
+ * node('linux-arm64') {
+ *     stage('[Post-Build] Status Report') {
+ *         publishBuildStatusReport()
+ *     }
+ * }
+ */
+def call(Map config = [:]) {
+  // if (env.BRANCH_IS_PRIMARY == null || !env.BRANCH_IS_PRIMARY.toBoolean()) {
```
dduportal, line 22:
```suggestion
  // if (!env.BRANCH_IS_PRIMARY) {
```

Ref. https://docs.groovy-lang.org/latest/html/documentation/core-semantics.html#the-groovy-truth and https://docs.groovy-lang.org/latest/html/documentation/core-semantics.html#_object_references in particular

=> Groovy considers `null` to be `false` when called in a logical expression.

Art39print (approved):

Art39print (approved):

resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
```diff
@@ -0,0 +1,42 @@
...
+cd "$REPORT_DIR"
+azcopy logout >/dev/null 2>&1 || true
+test -z "${AZURE_FEDERATED_TOKEN_FILE:-}" || export AZCOPY_AUTO_LOGIN_TYPE=WORKLOAD
+azcopy login --identity
+
+# Try azcopy with error handling
+if ! azcopy copy "status.json" "$DESTINATION_URL"; then
+    echo "azcopy failed, collecting logs for debugging"
+    # Retrieve azcopy logs to archive them
+    cat /home/jenkins/.azcopy/*.log > "$WORKSPACE/azcopy.log" 2>/dev/null
```
dduportal, line 40: What if the username (or user home) changes? ;)

resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
```diff
@@ -0,0 +1,34 @@
...
+}
+EOF
+
+# Upload with azcopy
+DESTINATION_URL="https://buildsreportsjenkinsio.file.core.windows.net/builds-reports-jenkins-io/build_status_reports/$CONTROLLER_HOSTNAME/$JOB_NAME"
+
+cd "$REPORT_DIR"
+azcopy logout >/dev/null 2>&1 || true
+test -z "${AZURE_FEDERATED_TOKEN_FILE:-}" || export AZCOPY_AUTO_LOGIN_TYPE=WORKLOAD
+azcopy login --identity
```
dduportal, line 34: You are missing the `azcopy copy <...>` command (accidentally remove in your last commit)

dduportal (approved):
LGTM! We tested all cases on jenkins-infra/docker-404 and it works as expected: https://builds.reports.jenkins.io/build_status_reports/infra.ci.jenkins.io/docker-jobs/docker-404/main/status.json

lemeurherve:
> LGTM! We tested all cases on jenkins-infra/docker-404 and it works as expected: https://builds.reports.jenkins.io/build_status_reports/infra.ci.jenkins.io/docker-jobs/docker-404/main/status.json

How did you tested the report of a failure?

Current docker-404 pipeline can't catch error in `buildAndPublishDockerImage` AFAICT:
```groovy
buildDockerAndPublishImage('404', [
    publishToPrivateAzureRegistry: true,
    targetplatforms: 'linux/amd64,linux/arm64',
]) // test

node('jnlp-linux-arm64') {
    stage('[Post-Build] Status Report') {
        echo "[Docker-404] After main call. Current result: ${currentBuild.currentResult}, Branch Primary: ${env.BRANCH_IS_PRIMARY}"
        publishBuildStatusReport()
    }
}
```
Jenkins will exit before reaching `publishBuildStatusReport`.

Tested a pipeline with `try..catch`, but the build status seems to always return `SUCCESS`:
```groovy
node('jnlp-linux-arm64') {
    try {
        stage('Test') {
            sh 'echo "Fail!"; exit 1'
        }
        echo 'This will run only if successful'
    } catch (e) {
        echo 'This will run only if failed'

        // Since we're catching the exception in order to report on it,
        // we need to re-throw it, to ensure that the build is marked as failed
        throw e
    } finally {
        def currentResult = currentBuild.currentResult ?: 'UNKNOWN'
        echo "currentResult: $currentResult"

        publishBuildStatusReport()
    }
}
```
https://infra.ci.jenkins.io/job/docker-jobs/job/docker-404/job/main/219/console

Also, this seems to require editing the pipeline for all jobs we want, which is less than ideal.

Is there a way to run `publishBuildStatusReport` independently of the pipelines of the jobs we want to monitor?

dduportal:
> How did you tested the report of a failure?
> 
> Current docker-404 pipeline can't catch error in `buildAndPublishDockerImage` AFAICT:
> 
> ```groovy
> buildDockerAndPublishImage('404', [
>     publishToPrivateAzureRegistry: true,
>     targetplatforms: 'linux/amd64,linux/arm64',
> ]) // test
> 
> node('jnlp-linux-arm64') {
>     stage('[Post-Build] Status Report') {
>         echo "[Docker-404] After main call. Current result: ${currentBuild.currentResult}, Branch Primary: ${env.BRANCH_IS_PRIMARY}"
>         publishBuildStatusReport()
>     }
> }
> ```
> 
> Jenkins will exit before reaching `publishBuildStatusReport`.

The staleness is the more important thing for us as per what has been specified to Jay. We can think about failure detection in a second eventual step, but what we want is "last success was before/after XX min". If the staleness is reached due to (uncaught in current step as you underlined) failure, then the monitor is triggered.

=> Need a sync between @jayfranco999 @lemeurherve to clarify, but we are in the right direction

PR #940, opened 2025-06-09
feat(buildDockerAndPublishImage) add `cache-to` parameter to build docker images

jayfranco999:
As per https://github.com/jenkins-infra/helpdesk/issues/4683:

This PR add `cache-to` parameter while using the pipeline library script

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -4,7 +4,7 @@ import java.util.Date
 import java.text.DateFormat
 
 // makecall is a function to concentrate all the call to 'make'
-def makecall(String action, String imageDeployName, String targetOperationSystem, String specificDockerBakeFile, String dockerBakeTarget, String cacheToString = '') {
+def makecall(String action, String imageDeployName, String targetOperationSystem, String specificDockerBakeFile, String dockerBakeTarget, String cacheTo) {
   final String bakefileContent = libraryResource 'io/jenkins/infra/docker/jenkinsinfrabakefile.hcl'
   // Please note that "make deploy" and the generated bake deploy file uses the environment variable "IMAGE_DEPLOY_NAME"
   if (isUnix()) {
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -14,13 +14,13 @@ def makecall(String action, String imageDeployName, String targetOperationSystem
     }
     withEnv(
         [
-          "DOCKER_BAKE_FILE=${specificDockerBakeFile}",
-          "DOCKER_BAKE_TARGET=${dockerBakeTarget}",
-          "IMAGE_DEPLOY_NAME=${imageDeployName}"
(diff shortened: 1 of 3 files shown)


vars/buildDockerAndPublishImage.groovy
```diff
@@ -4,22 +4,24 @@ import java.util.Date
 import java.text.DateFormat
 
 // makecall is a function to concentrate all the call to 'make'
-def makecall(String action, String imageDeployName, String targetOperationSystem, String specificDockerBakeFile, String dockerBakeTarget) {
+def makecall(String action, String imageDeployName, String targetOperationSystem, String specificDockerBakeFile, String dockerBakeTarget, String cacheToString = '') {
```
dduportal, line 7:
```suggestion
def makecall(String action, String imageDeployName, String targetOperationSystem, String specificDockerBakeFile, String dockerBakeTarget, String cacheTo) {
```

Nit (Not sure why appending the type - `String` in the variable name?)

vars/buildDockerAndPublishImage.groovy
```diff
@@ -4,22 +4,24 @@ import java.util.Date
...
-    ]) {
-      sh 'export BUILDX_BUILDER_NAME=buildx-builder; docker buildx use "${BUILDX_BUILDER_NAME}" 2>/dev/null || docker buildx create --use --name="${BUILDX_BUILDER_NAME}"'
-      sh "make bake-$action"
-    }
+    withEnv(
+        [
+          "DOCKER_BAKE_FILE=${specificDockerBakeFile}",
+          "DOCKER_BAKE_TARGET=${dockerBakeTarget}",
+          "IMAGE_DEPLOY_NAME=${imageDeployName}"
+        ] + ((cacheToString && !cacheToString.isEmpty()) ? ["DOCKER_CACHE_TO_VALUE=${cacheToString}"] : []) // Conditionally add cacheTo
```
dduportal, line 20:
```suggestion
        ] + ((cacheToString && !cacheToString.isEmpty()) ? ["DOCKER_CACHE_TO=${cacheToString}"] : []) // Conditionally add cacheTo
```

What is the reason to add `_VALUE` at the end of the name?

vars/buildDockerAndPublishImage.groovy
```diff
@@ -50,6 +52,7 @@ def call(String imageShortName, Map userConfig=[:]) {
     unstash: '', // Allow to unstash files if not empty
     dockerBakeFile: '', // Specify the path to a custom Docker Bake file to use instead of the default one
     dockerBakeTarget: 'default', // Specify the target of a custom Docker Bake file to work with
+    cacheTo: '', // New parameter for Docker build cache export using cache-to
```
dduportal, line 55: You need unit test to cover both the default case and with a string passed as argument

vars/buildDockerAndPublishImage.groovy
jayfranco999: Just to avoid ambiguity since `cache-to` is a flag.

dduportal: But it is an environment variable passed to shell, there are no ambiguity (besides it is prefixed by `DOCKER_`)

dduportal (approved):
👏

Issue #945, opened 2025-06-13
`infra.getBuildAgentLabel(...)` Fails in `buildPluginWithGradle


TT-S3-D1:
### Reproduction steps

The issue occurs when calling the step as follows (but also with any other specification of platform and jdk):

```groovy
buildPluginWithGradle(
    configurations: [
        [platform: 'linux', jdk: 17],
        [platform: 'windows', jdk: 17]
    ],
    failFast: false,
)
```

### Expected Results

The pipeline should proceed without error and execute builds for the specified configurations.

### Actual Results

The following error is thrown during pipeline execution:

```
No such DSL method 'getBuildAgentLabel' found among steps [...] or globals [...]
```

## Description
There are runtime failures in the `buildPluginWithGradle` shared library step when invoking `infra.getBuildAgentLabel(...)`.

For example [this ecu-test-execution plugin run](https://github.com/jenkinsci/ecu-test-execution-plugin/runs/40477023173) fails with the following error.

## Additional Context

The method is defined in `vars/infra.groovy` and is used successfully in other parts of the shared library, including the `buildPlugin` (Maven-based) step. 

This regression appears to have started with commit [050aae0](https://github.com/jenkins-infra/pipeline-library/commit/050aae058e71a42a75f8d7c83c63f9ea9acfb65e), which introduced the call to getBuildAgentLabel(...) into the Gradle build logic.

dduportal:
Thanks for reporting, good catch! 

I thought I did enough testing with the `buildPluginWithGradle()` in https://github.com/jenkins-infra/pipeline-library/pull/926 (the commit you properly identified) but looks like it wasn't enough.

Let us check the problem.

dduportal:
FIx ready in https://github.com/jenkins-infra/pipeline-library/pull/946 and tested with https://ci.jenkins.io/job/Plugins/job/ecu-test-execution-plugin/job/PR-237/8/console (replay with the line `@Library('pipeline-library@pull/946/head') _` added in top of the `Jenkinsfile`)

PR #946, opened 2025-06-13
fix(buildPluginWithGradle) use valid signature when calling `infra.getBuildAgentLabel()` method (`jdk` should be a String)

dduportal:
Ref. https://github.com/jenkins-infra/pipeline-library/issues/945

Fixup of #926 

This PR is a bugfix (started as debug) which corrects the call to the `infra.getBuildAgentLabel()` method.

The second argument is expected to be the JDK major version but as a **String** (yes poor design I made in #926), while the `config.jdk` seems to be casted implicitly to something else.

The bugfix defines local variables inside the loop which are explicitly typed as `String` which fix the error message `No such DSL method 'getBuildAgentLabel' found among steps [...] or globals [...]` by using the proper method signature call.

Tested with https://ci.jenkins.io/job/Plugins/job/ecu-test-execution-plugin/job/PR-237/8/console (replay with the line `@Library('pipeline-library@pull/946/head') _` added in top of the `Jenkinsfile`)

### DIFF
--- a/vars/buildPluginWithGradle.groovy
+++ b/vars/buildPluginWithGradle.groovy
@@ -20,10 +20,10 @@ def call(Map params = [:]) {
   boolean archivedArtifacts = false
   Map tasks = [failFast: failFast]
   buildPlugin.getConfigurations(params).each { config ->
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):
Looks good to me.

PR #956, opened 2025-08-28
feat(publishBuildStatusReport): change report timestamp from ISO 8601 to Unix epoch format

jayfranco999:
As per https://github.com/jenkins-infra/helpdesk/issues/2843#issuecomment-3187753080

This PR

- Update generateAndWriteBuildStatusReport.sh to use Unix timestamp
- Change from ISO format to epoch seconds for better monitoring compatibility
- Prepare for timestamp age validation in Datadog synthetics


smerle33 (approved):
🚀

PR #1001, opened 2026-03-31
Fix terraform production credentials using wrong config map

### LINKED ISSUE #1000: terraform step uses defaultConfig instead of finalConfig for production credentials
## Description

While reviewing [`terraform.groovy`](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy), I noticed that the production stage passes `defaultConfig.productionCredentials` instead of `finalConfig.productionCredentials` to `withCredentials`.

## Impact

Any caller supplying custom production credentials via the `productionCredentials` parameter has those credentials silently ignored. `defaultConfig` is initialized with an empty array on [line 9](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L9) and is never modified. Only `finalConfig` ([line 18](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L18)) contains the merged user configuration after the `<<` operator is applied.

**Affected code ([line 74](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L74)):**
```groovy
withCredentials(defaultConfig.productionCredentials) {  // ❌ Always []
```

**Correct pattern already used for staging ([line 49](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L49)):**
```groovy
withCredentials(finalConfig.stagingCredentials) {  // ✅ Correct
```

## Reproduction

Any Jenkinsfile calling `terraform()` with custom production credentials is affected:
```groovy
terraform(
  productionCredentials: [
    usernamePassword(credentialsId: 'aws-prod', ...)
  ]
)
```

The credentials are merged into `finalConfig` but never used — `withCredentials` on [line 74](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L74) reads from `defaultConfig` instead.

## Proposed Fix
```diff
- withCredentials(defaultConfig.productionCredentials) {
+ withCredentials(finalConfig.productionCredentials) {
```

esosaoh:
Fixes #1000

## Changes

Replaces `defaultConfig.productionCredentials` with `finalConfig.productionCredentials` on [line 74](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L74) of [`vars/terraform.groovy`](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy).

The `defaultConfig` map is initialized with empty arrays and never modified. User-supplied credentials are merged into `finalConfig` via the `<<` operator ([line 18](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L18)), but the production stage was reading from the wrong map, causing all custom production credentials to be silently ignored.

This matches the existing pattern used for staging credentials ([line 49](https://github.com/jenkins-infra/pipeline-library/blob/master/vars/terraform.groovy#L49)).

## Testing Done

- Verified the change doesn't introduce new test failures: `mvn test` shows the same 73 pre-existing failures as `master` (unrelated to this change)
- The fix is a single-variable rename with no logic changes
- Existing `TerraformStepTests` pass for staging credentials (which correctly use `finalConfig`); production credentials were untested because the bug prevented them from ever being used

## Submitter Checklist

- [x] Make sure you are opening from a topic/feature/bugfix branch (right side) and not your main branch
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira

### DIFF
--- a/vars/terraform.groovy
+++ b/vars/terraform.groovy
@@ -71,7 +71,7 @@ def call(userConfig = [:]) {
     parallelStages['production'] = {
       stage('Production') {
         agentTemplate(finalConfig.agentLabel, {
-          withCredentials(defaultConfig.productionCredentials) {
+          withCredentials(finalConfig.productionCredentials) {
             final String planFileName = 'terraform-plan-for-humans.txt'
             def scmOutput
             stage('🦅 Generate Terraform Plan') {


dduportal (approved):
Thank you, good catch!

This change makes sense. It does not fix any behavior as the `defaultConfig` map has the exact same value as `finalConfig` since we do not touch it at all. It is mutated by the `<<` operator and then `finalConfig` is assigned with the same content: explains why it worked without any problem.

However, your fix make sense:
- staging and production will have the same behavior (consistency)
- Accidental mutation of `defaultConfig` *after* the initialization of `finalConfig` won't create any mayhem if it happens
- `finalConfig` is immutable once assigned: safer!


I'll let @lemeurherve merge at his own convienience so he can monitor the effect under his own planning

lemeurherve (approved):

PR #1009, opened 2026-04-20
feat(publishBuildStatusReport): allow to pass `AZCOPY_LOGIN_IDENTITY_RESOURCE_ID`

lemeurherve:
This change allows with a92c1e9 to pass `AZCOPY_LOGIN_IDENTITY_RESOURCE_ID` to specify the UAID to use (ex from https://reports.jenkins.io/jenkins-infra-data-reports/azure.json) when calling `generateAndWriteBuildStatusReport.sh` script directly. (In a freestyle job like update_center on trusted.ci.jenkins.io for example).

It also applies shellcheck, requires parameters and uses lowercase for internal variable in 8ff17cd, and removes the unused `BUILD_STATUS` variable in a316e50 so it doesn't have to be set when calling the script directly.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2843

#### Testing done
`mvn test -Dtest=PublishBuildStatusReportStepTests`

### DIFF
--- a/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
+++ b/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
@@ -8,7 +8,6 @@ set -euxo pipefail
 # Required parameters that must be set by the caller
 : "${JOB_NAME:?JOB_NAME is not set}"
 : "${BUILD_NUMBER:?BUILD_NUMBER is not set}"
-: "${BUILD_STATUS:?BUILD_STATUS is not set}"
 
 # Extract hostname from JENKINS_URL
 controller_hostname=$(echo "${JENKINS_URL}" | sed 's|https\?://||' | cut -d'/' -f1)
@@ -29,7 +28,6 @@ cat > "${report_file}" << EOF
   "controller_url": "${JENKINS_URL}",
   "job_name": "${JOB_NAME}",
   "build_number": "${BUILD_NUMBER}",
-  "build_status": "${BUILD_STATUS}",
   "report_timestamp": "${report_timestamp}"
 }
 EOF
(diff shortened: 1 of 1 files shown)


resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
```diff
@@ -1,36 +1,46 @@
...
 azcopy logout >/dev/null 2>&1 || true
 test -z "${AZURE_FEDERATED_TOKEN_FILE:-}" || export AZCOPY_AUTO_LOGIN_TYPE=WORKLOAD
-azcopy login --identity
-azcopy copy "status.json" "$DESTINATION_URL" --recursive
+if [[ -n "${AZCOPY_LOGIN_IDENTITY_RESOURCE_ID:-}" ]]; then
+  azcopy login --identity --identity-resource-id "${AZCOPY_LOGIN_IDENTITY_RESOURCE_ID}"
+else
+  azcopy login --identity
+fi
+azcopy copy "status.json" "${destination_url}"
```
dduportal:
```suggestion
azcopy copy "status.json" "${destination_url}" --recursive
```

Flag has been removed but it should not (takes care of creating sub directories if not existing)

dduportal:
Looks good, could you try it, though, on a PR in infra.ci to be sure we don't have any regression? (as the maven unit tests only allows checking for pipeline code, not shell)

lemeurherve:
> Looks good, could you try it, though, on a PR in infra.ci to be sure we don't have any regression? (as the maven unit tests only allows checking for pipeline code, not shell)

Test in a replay on `main` job of docker-404 (as the pipeline library function exit early when running from a PR): ✅ 
https://infra.ci.jenkins.io/job/docker-jobs/job/docker-404/job/main/224/replay/diff

dduportal (approved):

PR #1010, opened 2026-04-21
fix(generateAndWriteBuildStatusReport): quote `SCRIPT_PATH` to handle workspace paths with spaces

jayfranco999:
### Ref: 

- https://github.com/jenkins-infra/helpdesk/issues/2843
- https://github.com/jenkins-infra/release/pull/891

Jobs with spaces in their name (e.g. "Infra Agents Health") cause the `publishBuildStatusReport` step to fail with exit code 127:

```
+ bash /home/jenkins/agent/workspace/Infra Agents Health@tmp/generateAndWriteBuildStatusReport.sh
bash: /home/jenkins/agent/workspace/Infra: No such file or directory
```

The unquoted `${SCRIPT_PATH}` in the `sh` block undergoes word splitting, bash receives `/home/jenkins/agent/workspace/Infra` as the script path instead of the full string.

### DIFF
--- a/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
+++ b/resources/io/jenkins/infra/pipeline/generateAndWriteBuildStatusReport.sh
@@ -12,8 +12,11 @@ set -euxo pipefail
 # Extract hostname from JENKINS_URL
 controller_hostname=$(echo "${JENKINS_URL}" | sed 's|https\?://||' | cut -d'/' -f1)
 
+# Sanitize JOB_NAME for filesystem and URL safety
+sanitized_job_name=$(echo "${JOB_NAME}" | tr ' ' '-')
+
 # Build file path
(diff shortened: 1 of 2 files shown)


dduportal (dismissed):
You seem to have misunderstood the problem in https://github.com/jenkins-infra/release/pull/891#issuecomment-4286459301.

The pipeline library was never able to even run the script. Your change in its current state is non functional (and most probably will break other things). Using space in file name or URL is valid and legit: it only requires a bit of caution. By adding a sanitize function here, you'll have to build "unsanitize" functions on each consumer: sounds like a tedious thing to implement.

The error that @lemeurherve had is the following:

Image

```
+ bash /home/jenkins/agent/workspace/Infra Agents Health@tmp/generateAndWriteBuildStatusReport.sh
21:59:12  bash: /home/jenkins/agent/workspace/Infra: No such file or directory
script returned exit code 127
```

Fixing this error needs to start by analysing its root cause. Lets move the discussion to the issue :)

dduportal (approved):
Looks good! Thanks!

PR #1022, opened 2026-06-04
fix(terraform) replace `TERRAFORM_ENVIRONMENT` by `TF_VAR_environment`

dduportal:
Fixup of #1021

Defining the environment as env. var is really useful. But if Terraform is able to read the value as input variable it is even better

### DIFF
--- a/vars/terraform.groovy
+++ b/vars/terraform.groovy
@@ -51,7 +51,7 @@ def call(userConfig = [:]) {
       parallelStages['staging'] = {
         stage('Staging') {
           agentTemplate(finalConfig.agentLabel, {
-            withEnv(['TERRAFORM_ENVIRONMENT=staging']) {
+            withEnv(['TF_VAR_environment=staging']) {
               withCredentials(finalConfig.stagingCredentials) {
                 stage('🔎 Validate Terraform for Staging Environment') {
                   getInfraSharedTools(sharedToolsSubDir)
@@ -78,7 +78,7 @@ def call(userConfig = [:]) {
     parallelStages['production'] = {
       stage('Production') {
         agentTemplate(finalConfig.agentLabel, {
-          withEnv(['TERRAFORM_ENVIRONMENT=production']) {
+          withEnv(['TF_VAR_environment=production']) {
             withCredentials(finalConfig.productionCredentials) {
               final String planFileName = 'terraform-plan-for-humans.txt'
               def scmOutput


lemeurherve (dismissed):

vars/terraform.groovy
```diff
@@ -51,7 +51,7 @@ def call(userConfig = [:]) {
       parallelStages['staging'] = {
         stage('Staging') {
           agentTemplate(finalConfig.agentLabel, {
-            withEnv(['TERRAFORM_ENVIRONMENT=staging']) {
+            withEnv(['TF_VAR_environment=staging']) {
```
lemeurherve, line 54:
```suggestion
            withEnv(['TF_VAR_tfenvironment=staging']) {
```
More explicit

vars/terraform.groovy
```diff
@@ -78,7 +78,7 @@ def call(userConfig = [:]) {
     parallelStages['production'] = {
       stage('Production') {
         agentTemplate(finalConfig.agentLabel, {
-          withEnv(['TERRAFORM_ENVIRONMENT=production']) {
+          withEnv(['TF_VAR_environment=production']) {
```
lemeurherve, line 81:
```suggestion
          withEnv(['TF_VAR_tfenvironment=production']) {
```
Ditto

vars/terraform.groovy
dduportal: The goal is to have an input variable named `environment` such as:

```
variable "environment" {
  type    = string
  default = "staging"
}
```

I'm not sure what is not explicit in `environment`?

lemeurherve: In #1021 you stated the goal as "allow our common tooling to determine what kind of environment is being used".

Having a `tf` prefix makes it explicit that the environment in question relates to terraform and not something else.

lemeurherve: Those suggestions aren't blocking, fine for me if you merge as it is.

dduportal: > Those suggestions aren't blocking, fine for me if you merge as it is.

Ack, I'm merging then


> Having a tf prefix makes it explicit that the environment in question relates to terraform and not something else.

What would be the something else? I'm genuinely don't understand?

lemeurherve: I don't know, just wanted to be sure. Your usage isn't clear to me yet.

lemeurherve: BTW, if we start relying on those env vars, shouldn't their values be checked in tests?

PR #1028, opened 2026-06-23
feat(infra) implement Windows case for `loadMavenLocalCacheIfAny()` and refine existing shell code

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/5194, this PR implements the Windows (powershell) case for the `infra.loadMavenLocalCacheIfAny()` function.

It includes 2 changes:

- Implementation in Powershell
- A set of refinements in the Linux shell code (feedbacks from Claude Code, my shell habits and feedbacks from @lemeurherve during code review) gathered along the way following #1025 

----

Note: currently being tested in https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/PR-220/3/stages/?selected-node=7 (https://github.com/jenkinsci/jenkins-infra-test-plugin/pull/220/changes/09fa9c382cbcbdc156c31e6898a74279e63f88e7)

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -299,16 +299,22 @@ Object loadMavenLocalCacheIfAny(String mvnLocalRepo, String cachePath = '/cache/
   if (isUnix()) {
     withEnv(["MVN_LOCAL_REPO=${mvnLocalRepo}", "MVN_CACHE_PATH=${cachePath}",]) {
       sh '''
-      test -n "${MVN_LOCAL_REPO}" || export MVN_LOCAL_REPO="$HOME/.m2/repository"
+      : "${MVN_CACHE_PATH:?MVN_CACHE_PATH must be set}"
+      export MVN_LOCAL_REPO="${MVN_LOCAL_REPO:-$HOME/.m2/repository}"
+
       mkdir -p "${MVN_LOCAL_REPO}"
-      if test -f "${MVN_CACHE_PATH}";
+
+      if test -f "${MVN_CACHE_PATH}"
       then
-        pushd "${MVN_LOCAL_REPO}"
-        cache_archive_name=../"$(basename "${MVN_CACHE_PATH}")"
+        # MVN_CACHE_PATH might served from a CSI S3 volume which does not support seeking.
+        # tar requires a seekable source, so we copy the archive locally first.
+        # The copy lands in the parent of MVN_LOCAL_REPO which has sufficient disk space.
+        cache_archive_name="$(dirname "${MVN_LOCAL_REPO}")/$(basename "${MVN_CACHE_PATH}")"
         time cp "${MVN_CACHE_PATH}" "${cache_archive_name}"
-        time tar xzf "${cache_archive_name}" ./
+        time tar xzf "${cache_archive_name}" -C "${MVN_LOCAL_REPO}"
         rm -f "${cache_archive_name}"
-        popd
+      else
+        echo "${MVN_CACHE_PATH} archive not found: skipping maven cache retrieval."
       fi
       '''
     }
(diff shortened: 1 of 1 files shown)


vars/infra.groovy
```diff
@@ -299,21 +299,49 @@ Object loadMavenLocalCacheIfAny(String mvnLocalRepo, String cachePath = '/cache/
...
+      export MVN_LOCAL_REPO="${MVN_LOCAL_REPO:-$HOME/.m2/repository}"
+
       mkdir -p "${MVN_LOCAL_REPO}"
-      if test -f "${MVN_CACHE_PATH}";
+
+      if test -f "${MVN_CACHE_PATH}"
       then
-        pushd "${MVN_LOCAL_REPO}"
-        cache_archive_name=../"$(basename "${MVN_CACHE_PATH}")"
+        # MVN_CACHE_PATH might served from a CSI S3 volume which does not support seeking.
```
timja:
```suggestion
        # MVN_CACHE_PATH might be served from a CSI S3 volume which does not support seeking.
```

lemeurherve (approved):

dduportal:
> Note: currently being tested in https://ci.jenkins.io/job/Plugins/job/jenkins-infra-test-plugin/job/PR-220/3/stages/?selected-node=7 (https://github.com/jenkinsci/jenkins-infra-test-plugin/commit/09fa9c382cbcbdc156c31e6898a74279e63f88e7)

- No regression on Linux: archive is still uncompressed and takes ~10s
- No regression on Window: warning message is printed that no archive is found

timja:
ATH failing: https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-2685/7/stages/?selected-node=465

https://github.com/jenkinsci/acceptance-test-harness/pull/2685

dduportal:
> ATH failing: https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-2685/7/stages/?selected-node=465
> 
> [jenkinsci/acceptance-test-harness#2685](https://github.com/jenkinsci/acceptance-test-harness/pull/2685)

I'm looking into it. IF no quick fix found I'll revert

dduportal:
@timja tracking in https://github.com/jenkins-infra/helpdesk/issues/5193#issuecomment-4787267159 (reopened the issue). Note: the problem seems to be present since #1025

vars/infra.groovy
```diff
@@ -299,21 +299,49 @@ Object loadMavenLocalCacheIfAny(String mvnLocalRepo, String cachePath = '/cache/
...
+
       mkdir -p "${MVN_LOCAL_REPO}"
-      if test -f "${MVN_CACHE_PATH}";
+
+      if test -f "${MVN_CACHE_PATH}"
       then
-        pushd "${MVN_LOCAL_REPO}"
-        cache_archive_name=../"$(basename "${MVN_CACHE_PATH}")"
+        # MVN_CACHE_PATH might be served from a CSI S3 volume which does not support seeking.
+        # tar requires a seekable source, so we copy the archive locally first.
```
jtnord, line 310: it doesn't, or you would never be able to `untar` from a pipe :)  

`tar` should automatically detect if it has a seekable or non-seekable source, if its failing in this specific case then use the `--no-seek` flag :) 

(seeking only helps when unpacking a subset of files as it can offset without reading, but in our case we are extracting everything so should not make a material difference)

dduportal: Yes, but the pipeline-library can't tell what kind of mountpoint is used (otherwise we would need a terribly wrong if/then/elseif/..) as it depends on the kind of agent: container Linux with S3 CSI? EmptyDir? Other? Windows VM with local FS? or data disk NVMe? etc.)

So we keep the 2 times process (cp + tar) because it's does not have real negative impact but handles all cases properly (unless we missed something)

dduportal: > (seeking only helps when unpacking a subset of files as it can offset without reading, but in our case we are extracting everything so should not make a material difference)

With the CSI S3 mountpoint, the time goes through the roof: we are at ~2:30m instead of ~6 to 10s when "un-taring" from S3's mount point. And of course, the amount of request to S3 endpoints goes from 3 to more than 100s.

PR #1039, opened 2026-08-20
feat(infra.maybePublishIncrementals): allow to pass an alternative build id

lemeurherve:
This change allows to pass an alternative build URL to incrementals-publisher service on publication.

Use case: `infra.prepareToPublishIncrementals()` artifacts prepared and archived in a previous build, then calling `infra.maybePublishIncrementals` in a subsequent build which doesn't have those archives.

### Testing done

Build of https://github.com/jenkinsci/bom/pull/7324 (testing https://github.com/jenkinsci/bom/pull/7319)

https://ci.jenkins.io/job/Tools/job/bom/job/PR-7324/12/stages/?selected-node=89
> image

### DIFF
--- a/vars/infra.groovy
+++ b/vars/infra.groovy
@@ -526,7 +526,7 @@ void prepareToPublishIncrementals() {
  * Call at the end of the build, outside any node, when #prepareToPublishIncrementals may have been called previously.
  * See INFRA-1571 and JEP-305.
  */
-void maybePublishIncrementals(String optionalAlternativeBuildUrl = '') {
+void maybePublishIncrementals(int optionalAlternativeBuildId = 0) {
   if (new InfraConfig(env).isRunningOnJenkinsInfra() && currentBuild.currentResult == 'SUCCESS') {
     if (env.CHANGE_ID == null) {
       def skip
@@ -540,9 +540,9 @@ void maybePublishIncrementals(String optionalAlternativeBuildUrl = '') {
     }
     stage('Deploy') {
       def buildUrlToUse = env.BUILD_URL
-      if (optionalAlternativeBuildUrl && optionalAlternativeBuildUrl != buildUrlToUse) {
-        echo "Using alternative build URL ${optionalAlternativeBuildUrl} instead of ${buildUrlToUse}"
-        buildUrlToUse = optionalAlternativeBuildUrl ?: env.BUILD_URL
+      if (optionalAlternativeBuildId > 0 && optionalAlternativeBuildId != env.BUILD_ID) {
+        buildUrlToUse = "${env.JOB_URL}/${optionalAlternativeBuildId}/"
+        echo "Using alternative build URL ${buildUrlToUse} instead of ${env.BUILD_URL}"
       }
 
       withCredentials([string(credentialsId: 'incrementals-publisher-token', variable: 'FUNCTION_TOKEN')]) {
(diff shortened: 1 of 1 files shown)


dduportal:
Mmmh this change feels a bit unsafe as it allows passing a full url. Is there any way to have a specific url building insise the function and only use a boolean argument?

lemeurherve:
> Mmmh this change feels a bit unsafe as it allows passing a full url. Is there any way to have a specific url building insise the function and only use a boolean argument?

Proposing to pass only the build id as the rest of the URL should remain the same.

lemeurherve:
> > Mmmh this change feels a bit unsafe as it allows passing a full url. Is there any way to have a specific url building insise the function and only use a boolean argument?
> 
> Proposing to pass only the build id as the rest of the URL should remain the same.

An alternative could be to check for the presence of a file containing the build ID at a specific location (ex: `target/build-if-ref-for-incrementals.txt`) and retrive the build id to use from that file.

Pro: unchanged method signature
Con: less explicit

I personally prefer the build id implementation.

jglick (commented):
This makes me uncomfortable from a security perspective. Part of the Incrementals model is that trusted infrastructure uploads artifacts associated with the current build, which would not otherwise be permitted to make that upload. In fact the publisher is supposed to be calling back to the build to verify the contents. Please study the code on both sides and think very, very carefully about this before proceeding.

jglick:
What exactly is the purpose of this change? The PR description does not clarify why the use case would make any sense. If you are publishing incremental artifacts from a given commit of a given PR, the expectation is that the CI passed all tests, and produced those artifacts from the given source snapshot.

lemeurherve:
> What exactly is the purpose of this change? The PR description does not clarify why the use case would make any sense. If you are publishing incremental artifacts from a given commit of a given PR, the expectation is that the CI passed all tests, and produced those artifacts from the given source snapshot.

In https://github.com/jenkinsci/bom/pull/7319 for https://github.com/jenkinsci/bom/issues/7317, I'm preparing an archive from prep.sh result, then I'm letting infra.prepareIncrementals archiving artifacts needed for incrementals-publisher.

If the first build where those artifacts are stored doesn't pass (ex: a flaky test), infra.maybePublishIncrementals skip the deployment.

With https://github.com/jenkinsci/bom/pull/7319 subsequent builds of the bom job will retrieve the prep archive but won't run again infra.prepareIncrementals as it would publish new artifacts, screwing the retrieval of the prep archive with copyArtifact, unless I rearchive the prep archive, something I'd like to avoid in order to have only one build with that relatively large prep archive (1Go).

To avoid that, I'm storing the id of the first build where the incrementals artifacts have been archived (alongside the prep archive), so infra.maybePublishIncrementals could "deploy" (ie send the request to incrementals-publisher service) from the first build if all pct tests passed.

> This makes me uncomfortable from a security perspective. Part of the Incrementals model is that trusted infrastructure uploads artifacts associated with the current build, which would not otherwise be permitted to make that upload.

>  from a given commit of a given PR, the expectation is that the CI passed all tests, and produced those artifacts from the given source snapshot.

In my bom changes, the archive name contains the current revision, forcing the builds to prepare a new archive (and thus new incrementals artifacts) whenever a new commit is pushed, ensuring that any incrementals publication correspond to the commit actually tested in bom even if not the same build id.

---
As an alternative to preparing that prep archive from bom builds directly and to avoid this kind of complication preventing to use archiveArtifacts, I've got a different version using a downstream job in charge of preparing that archive only, but it's a bit more code, hence the continuation with https://github.com/jenkinsci/bom/pull/7319 and this PR to allow every potential successful bom build to publish a bom incrementals.

lemeurherve:
Closing, see https://github.com/jenkinsci/bom/issues/7317#issuecomment-5360700687

PR #1043, opened 2026-08-28
feat(buildDockerAndPublishImage) add support of `cert.ci.jenkins.io`

dduportal:
Requires #1045 #1047 and #1048

This PR introduces the support of the library `buildDockerAndPublishImage()` in cert.ci.jenkins.io to allow the Jenkins Security team to build and publish privately Docker images of the Jenkins infrastructure for autonomy during security advisories.

Notes:
- It does NOT cover publishing the `jenkinsci/docker*` images neither it is a staging for Docker images.
- The target is the (private) container registry in Azure which requires credentials set up per job inside cert.ci.jenkins.io
- For cert.ci.jenkins.io, we want the lint phase to happen (to avoid bad surprises for the Jenkins Security team when publishing code fixes) but we print the hadolint report in the stdout as the `warnings` plugin (providing the `recordIssue pipeline step` is not available
  - Also the case of trusted.ci.jenkins.io. Never had it until now but could be.
- The container image's tag must be provided by the calling pipeline as part of the "imageName" string parameter. cert.ci.jenkins.io is not expected to build from git tag or principal branch.
- We also disable publication of build report: no infra monitoring for these builds. The goal is to avoid leaking any reference

----

Tested with success in cert.ci.jenkins.io:

```
pushing manifest for *****/******/account-app:SECURITY-testddu@sha256:2381ea5af801316ac4ed69773c651d6234b6d30a1d2403af1d95a738fbcc7dce 1.5s done
```

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -143,7 +143,6 @@ def call(String imageShortName, Map userConfig =[:]) {
       "BAKE_TARGETPLATFORMS=${finalConfig.targetplatforms}",
       "REGISTRY=${registryHost}",
     ]) {
-      String nextVersion = ''
       stage("Prepare ${imageName}") {
         checkout scm
         if (finalConfig.unstash != '') {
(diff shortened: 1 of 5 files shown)


vars/buildDockerAndPublishImage.groovy
```diff
@@ -209,30 +235,23 @@ def call(String imageShortName, Map userConfig =[:]) {
...
+        } else if (env.TAG_NAME) {
+          // if a tag is scanned, it need to be the nextVersion
+          nextVersion = env.TAG_NAME
+          echo "nextVersion: ${nextVersion}"
         } else {
-          if (env.TAG_NAME) {
-            // if a tag is scanned, it need to be the nextVersion
-            nextVersion = env.TAG_NAME
-          }
+          echo "INFO: no 'nextVersion' detected. Relying on the user-provided image name for Docker tag."
```
dduportal: Nit: improve logging by printing the value of the user-provided image name

lemeurherve (approved):
🚀

PR #1047, opened 2026-08-31
feat(buildDockerAndPublishImage): allow execution in the current agent instead of allocating a new one. opt-in feature.

dduportal:
This PR introduces a new feature required by #1043 (cert.ci.jenkins.io) but which could be useful for some container images of our infrastructure

### DIFF
--- a/vars/buildDockerAndPublishImage.groovy
+++ b/vars/buildDockerAndPublishImage.groovy
@@ -39,6 +39,17 @@ def makecall(String action, String imageDeployName, String targetOperationSystem
   } // unix agent
 }
 
+def withAgent(Boolean runInCurrentAgent, String agentLabels, Closure body) {
+  if (runInCurrentAgent) {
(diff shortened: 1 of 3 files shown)


vars/buildDockerAndPublishImage.groovy
```diff
@@ -39,6 +39,17 @@ def makecall(String action, String imageDeployName, String targetOperationSystem
   } // unix agent
 }
 
+def withAgent(Boolean runInCurrentAgent, String agentLabels, Closure body) {
```
lemeurherve, line 42: Should `agentLabels` parameter have a default value to ensure it's never empty?

lemeurherve (approved):

vars/buildDockerAndPublishImage.groovy
dduportal: Could be, but such defensive pattern would need to write down validations first, before the logic. I'm assuming this function is not standalone (even tough the dynamic nature of Groovy makes my assertion wrong) and only called in this library. If an external consumer fails to pass proper values, I consider it their loss.

PR #1052, opened 2026-09-07
fix(buildDockerAndPublishImage): handle "non-nextversion" tags where the container tag is provided as suffix of the `imageName` parameter

dduportal:
Bug detected while working on #1043 

In such case (`buildDockerAndPublishImage("myapp:customtag", [automaticSemanticVersioning: false])`), the build fails with:

```
ERROR: invalid tag "myapp:customtag:latest": invalid reference format
````

This change fixes this behavior with 2 elements:
- Removes the concept of `NEXTVERSION` from the bake file in favor of a list of tags
  - Defaults is empty tags: it will imply an implicit `:latest` tag
- Makefile is responsible to detect the presence of a tag marker (the colon `:` character) in the provided image name.
  - If detected, it is passed as the only tag
  - Otherwise (nominal case) we passe the `NEXTVERSION` and `latest` explicit tags

NOTE: the bakefile does not have all HCL functions: it lacks `strcontains()` so we can't easily detect the colon inside bakefile

----

Tests in progress with 3 kinds of builds:

- [infra.ci] Replay a CD build for docker-404 to cover usual tagging when a release is published with a "NEXT VERSION" - https://infra.ci.jenkins.io/job/docker-jobs/job/docker-404/job/main/256/ 
- [infra.ci] Replay a PR build to ensure no non regression - https://infra.ci.jenkins.io/job/docker-jobs/job/docker-jenkins-lts/view/change-requests/job/PR-1120/4/
- [cert.ci] rebasing #1043 against this PR and retriggering a build to verify the "cert.ci" case with no NEXTVERSION

### DIFF
--- a/resources/io/jenkins/infra/docker/Makefile
+++ b/resources/io/jenkins/infra/docker/Makefile
@@ -13,6 +13,10 @@ export ARCH ?= $(shell case $(current_arch) in (x86_64) echo "amd64" ;; (i386) e
 
 IMAGE_NAME ?= helloworld
 IMAGE_DEPLOY_NAME ?= "$(IMAGE_NAME)"
+ifeq (,$(findstring :,$(IMAGE_DEPLOY_NAME)))
+	TAGS_LIST := $(NEXTVERSION),latest
+	export TAGS_LIST
+endif
 BUILD_TARGETPLATFORM ?= linux/amd64
 # Paths
 # Detect Windows (only OS separating paths in $PATH with ";") to set specific absolute paths for linux (bake) only
--- a/resources/io/jenkins/infra/docker/Makefile
+++ b/resources/io/jenkins/infra/docker/Makefile
@@ -13,9 +13,11 @@ export ARCH ?= $(shell case $(current_arch) in (x86_64) echo "amd64" ;; (i386) e
 
 IMAGE_NAME ?= helloworld
 IMAGE_DEPLOY_NAME ?= "$(IMAGE_NAME)"
+# If there are no ':' character in IMAGE_DEPLOY_NAME (nominal case) then we assume both $NEXTVERSION and 'latest' tags should be built/deployed
+# We use tedious Makefile because the HCL 'strcontains()' function is unavailable in Docker Bake
 ifeq (,$(findstring :,$(IMAGE_DEPLOY_NAME)))
-	TAGS_LIST := $(NEXTVERSION),latest
-	export TAGS_LIST
+	CUSTOMTAGS := $(NEXTVERSION),latest
+	export CUSTOMTAGS
 endif
 BUILD_TARGETPLATFORM ?= linux/amd64
 # Paths
--- a/resources/io/jenkins/infra/docker/jenkinsinfrabakefile.hcl
+++ b/resources/io/jenkins/infra/docker/jenkinsinfrabakefile.hcl
@@ -4,7 +4,7 @@ variable "REGISTRY" {
   default = "docker.io"
 }
 
-variable "NEXT_VERSION" {
+variable "TAGS_LIST" {
   default = ""
 }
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: resources/io/jenkins/infra/docker/Makefile [hunk 1]
```diff
@@ -13,6 +13,10 @@ export ARCH ?= $(shell case $(current_arch) in (x86_64) echo "amd64" ;; (i386) e
 
 IMAGE_NAME ?= helloworld
 IMAGE_DEPLOY_NAME ?= "$(IMAGE_NAME)"
+ifeq (,$(findstring :,$(IMAGE_DEPLOY_NAME)))
+	TAGS_LIST := $(NEXTVERSION),latest
+	export TAGS_LIST
+endif
```
lemeurherve, line 21: Could we have a `else TAGS_LIST="$(IMAGE_NAME)"` here to only pass TAGS_LIST to `all_tags` in the docker bake file? That would be a bit more simple

dduportal: nope, it would make the bake function to fail (check its logic)

dduportal: I don't mind to add a comment here though on the "why"

lemeurherve: > it would make the bake function to fail (check its logic)

I must be missing something, it's unclear to me how it would fail.

AFAICT, `IMAGE_DEPLOY_NAME` is used only once in the bake file. Passing only TAG_LIST to all_tags would be the same?

resources/io/jenkins/infra/docker/jenkinsinfrabakefile.hcl
```diff
@@ -4,7 +4,7 @@ variable "REGISTRY" {
   default = "docker.io"
 }
 
-variable "NEXT_VERSION" {
+variable "TAGS_LIST" {
```
lemeurherve: nit:
```suggestion
variable "TAGS" {
```
or

```suggestion
variable "TAG_LIST" {
```

dduportal: I don't mind but why?

lemeurherve: `TAGS_LIST` sounds wrong to my ears (the plural is implied by the fact it's a list).

Feel free to ignore.

resources/io/jenkins/infra/docker/Makefile
lemeurherve: I understand now that we need REGISTRY/IMAGE_DEPLOY_NAME as first parameter of `all_tags`.

resources/io/jenkins/infra/docker/jenkinsinfrabakefile.hcl
dduportal: Renamed to `CUSTOMTAGS` as agreed in pair

lemeurherve (approved):

resources/io/jenkins/infra/docker/Makefile
```diff
@@ -13,6 +13,12 @@ export ARCH ?= $(shell case $(current_arch) in (x86_64) echo "amd64" ;; (i386) e
 
 IMAGE_NAME ?= helloworld
 IMAGE_DEPLOY_NAME ?= "$(IMAGE_NAME)"
+# If there are no ':' character in IMAGE_DEPLOY_NAME (nominal case) then we assume both $NEXTVERSION and 'latest' tags should be built/deployed
+# We use tedious Makefile because the HCL 'strcontains()' function is unavailable in Docker Bake
+ifeq (,$(findstring :,$(IMAGE_DEPLOY_NAME)))
+	CUSTOMTAGS := $(NEXTVERSION),latest
```
dduportal, line 19:
```suggestion
	CUSTOMTAGS := $(NEXT_VERSION),latest
```

Otherwise we only have the `latest` tag passed to bake (typo)

dduportal: https://github.com/jenkins-infra/pipeline-library/pull/1054

PR #1057, opened 2026-09-15
Update badge URL in README.adoc to allow public access

shenxianpeng:
### Before change

image

### After change

image


dduportal (approved):
Thanks

